Issue #168, opened 2015-01-15
Add C3 Gauge


GordonSmith:
http://c3js.org/samples/chart_gauge.html

GordonSmith:
Fixed with:  https://github.com/hpcc-systems/Visualization/pull/193

PR #173, opened 2015-01-16
GH-168 Adds c3 gauge chart

mlzummo:


GordonSmith:
As a general comment - you should add a signature to your git commit settings.

widgets/src/c3/Gauge.js
```diff
@@ -0,0 +1,43 @@
...
+        this._type = "gauge";
+        this._config.data = {
+            columns: []
+        };
+    };
+    Gauge.prototype = Object.create(Common.prototype);
+
+    Gauge.prototype.palette = function (_) {
+        if (!arguments.length) return this._palette;
+        this._config.color.pattern = _;
```
GordonSmith: If "_" is one of our Palettes - I don't think this will work. If it isn't then we should rename the method name to c3_palette or such like?

mlzummo: I agree, let me look back over the color patterns for this widget... I am
wondering if maybe we should or could standardize the palettes or provide a
helper function?

GordonSmith: I agree, while I do have some helper functions in https://github.com/hpcc-systems/Visualization/blob/master/widgets/src/common/Palette.js

I think they should probably be wrapped up into a single Palette Object and then that can be shared at the Dashboard level (maybe one of each type?), which would ensure that each individual widget is working from the same color pool.

Then at the widget level you could simply create a new instance of the Palette object specifically for it (if you wanted a different color pool)?

PR #191, opened 2015-01-27
GH-168 Adds c3 gauge chart

mlzummo:
Please Review

Signed-off-by: Mathew Zummo <Mathew.Zummo@lexisnexis.com>

### DIFF
--- a/widgets/src/c3/Gauge.js
+++ b/widgets/src/c3/Gauge.js
@@ -18,22 +18,26 @@
     
     Gauge.prototype.publish("low", 0, "number", "Gauge lower bound");
     Gauge.prototype.publish("high", 100, "number", "Gauge higher bound");
-
+    Gauge.prototype.publish("units", "%", "string", "Gauge unit data is represented in. Default: %");
+    Gauge.prototype.publish("arcWidth", 75, "number", "Gauge width of arc");
+    
     Gauge.prototype.enter = function (domNode, element) {
         this._config.gauge = {
             min: this.low(),
-            max: this.high()
+            max: this.high(),
+            units: this.units(),
+            width: this.arcWidth(),
         };
+        
+        this._config.data = { columns: [[this.columns()[0], this._data]] }; // 1D
 
         Common.prototype.enter.apply(this, arguments);     
     };
 
     Gauge.prototype.update = function (domNode, element) {
         Common.prototype.update.apply(this, arguments);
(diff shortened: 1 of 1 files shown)


GordonSmith:
Can you provide an example of your widget.data(...) call?  

This widget falls into the "1D" Data category (like a slider) and I haven't formalized that interface yet, so am curious as to what your instinct was...)

mlzummo:
var cData = [['data', 30.4]];
widget.data(cData);

Same format as the others because we are just passing it to C3 columns: { }
Something I found interesting is that any other numeric elements after index 1 just get added to the first index so that  [['data', 50,20]] actually is the equiv. of saying: [['data', 70]]

mlzummo:
We should probably standardize with an object that allows your to pass low, high, step, unit, step might not be used in the gauge but each widget can read what it needs. we wouldn't need to call .range and .step on the slider either just the data object?

GordonSmith:
widget.data(30.4) would be more intuitive?  
Also Publish("low") and Publish("high")?

mlzummo:
Let me know if I approached this correctly with the new commit.

GordonSmith:
This is looking ok now - do you have an example of usage?

mlzummo:
var c3Gauge = new C3Gauge();

c3Gauge.low(200);
c3Gauge.high(300);
c3Gauge.units("%");
c3Gauge.arcWidth("20");

c3Gauge.columns(["Label 1"]);        
c3Gauge.data(data);

c3Gauge.target(htmlId);
c3Gauge.render();

GordonSmith:
Why is arcWidth in quotes?

mlzummo:
Typo

widgets/src/c3/Gauge.js
```diff
@@ -0,0 +1,49 @@
...
+    if (typeof define === "function" && define.amd) {
+        define(["./Common"], factory);
+    } else {
+        root.Gauge = factory(root.Common);
+    }
+}(this, function (Common) {
+    function Gauge(target) {
+        Common.call(this);
+
+        this._class = "c3_gauge";
```
GordonSmith: New convention - should be "c3_Gauge" - IOW the class name should match the path c3/Gauge with an underbar instead of the "/".  I will be using this to "locate" a persisted widgets JS file so needs to match exactly...

PR #195, opened 2015-02-02
Re-factor palette to global singletons

### LINKED ISSUE #192: C3/Other Chart Colors & Palette.js
So far with C3 it looks as though there are 2 ways I can do colors:

var ColorArr = Palette.custom(["orange", "green"], 0, 100, XVAR);
this._config.color.pattern = ColorArr.range();

where XVAR is the # of bars (or # of bars in each group)

this works for .ordinal .brewer with similar options.

basically your just giving the .pattern an array of colors.

the second way allows for a bit more customization but I am not sure if its needed or how to create an elegant way of implementing it and overall implementing both methods across all charts.

EX: 
        this._config.data.colors = {
            Geography: '#ff0000',
            English: '#00ff00',
            Math: '#0000ff',
            Science: '#0000ff'
        }

        this._config.data.color = function (color, d) {
            console.log(d);
            // d will be 'id' when called for legends
            return d.id && d.id === 'English' ? d3.rgb(color).darker(d.value / 150) : color;
            
        }

this will (if i can figure out where I went wrong in my code) and based off of what i see at http://c3js.org/samples/data_color.html let you
customize the color of each bar based off of the original input color for each bar + allows you access to the data (d) for potentially doing colors based off the data value

GordonSmith:
Re-factor palette to two global singletons:
* Ordinal
* Rainbow

Put google examples into their own code section (for working offline)
Add gauge example.

Update comms mappings to support new format.

Fix IE arrows for IE <= 12

Enable mouse click for MultiChartSurface.

Fixes #192

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="chrome=1">
     <title>HPCC Systems - Dermatology Test</title>
-    <link rel="stylesheet" href="test.css">
+    <link rel="stylesheet" href="test.css" />
     <script src="../widgets/lib/requirejs/require.js"></script>
     <script src="../widgets/config.js"></script>
     <script>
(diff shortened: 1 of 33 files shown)


GordonSmith:
Not for review (yet) - just for discussion...

GordonSmith:
@mlzummo @jbrundage @dineshshetye Please Review.

mlzummo:
Code works, should be ok to merge

PR #215, opened 2015-02-15
Publish properties for text, faChar, textbox, shape, icon and Vertex

GordonSmith:

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -48,6 +48,9 @@
             <option value="src/tree/Dendrogram">Dendrogram (Tree)</option>
             <option value="src/tree/CirclePacking">CirclePacking (Tree)</option>
             <option value="src/tree/SunburstPartition">SunburstPartition (Tree)</option>
+            <option value="src/google/Pie">Google Pie</option>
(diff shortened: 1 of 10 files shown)


GordonSmith:
@jbrundage @mlzummo Please Review - also note the new "optional" parameter for Publish.

jbrundage:
This new '_enable' feature will be handy.  Alternatively we may want to rethink the idea of publish arguments.  

If '.publish()' simply expects one object argument it would give us the flexibility of adding enhancements to the publish feature.  For example... grouping publish parameters.  Thoughts?

GordonSmith:
I have done away with "_enable" in a latter pull request...  Its not actually needed...

GordonSmith:
Thats what the "ext" is at the end for adding "other arguments" I did toy with doing it the way you suggested, but I felt it made the code to hard to scan...

So now you can have:  publish("prop", 22, "integer", "My Super Prop", null, {attr: "a", attr: "b"});

mlzummo:
looks good, ill be working with these widgets a little more over the next few days so i'll let you know if these changes cause any probs

GordonSmith:
Rebased for pull.

PR #250, opened 2015-03-09
Adds support for SVGWidget draw start pos to be modified

mlzummo:

### DIFF
--- a/widgets/src/common/SVGWidget.js
+++ b/widgets/src/common/SVGWidget.js
@@ -16,6 +16,8 @@
         this.transition = new Transition(this);
 
         this._renderCount = 0;
+
+        this._drawStartPos = "center"; // overridable default svg draw start position
     };
     SVGWidget.prototype = Object.create(Widget.prototype);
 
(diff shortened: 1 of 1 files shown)


GordonSmith:
What about HTML Widgets?

mlzummo:
HTML Widgets don't seem to be be effected by this, and no where in the code for HTMLWidget do I see anything that would hint that it draws from the center out.

mlzummo:
Or would you like me to add center capability to HTMLWidget... HTML doesn't behave like SVG does in regards to this

GordonSmith:
Was wondering if you had thought about it...  I suspect they should both work the same.  This change may simplify several of the widgets (which go out of their way to offset by width / 2 and height /2

mlzummo:
can you link/show me to the widget(s) that go out of there way to offset their width, so I can take a look at the code around that?

GordonSmith:
Dendrogram.js

mlzummo:
i never did understand that with the dendrogram, the SVGwidget itself is already set to draw from center by default and the dendrogram goes and does it yet again, (I am actually running a modified version of the dendrogram right now due to alignment and fitting issues) ... are there any HTMLWidgets that are set to draw from the center?

PR #252, opened 2015-03-11
Google Maps needs to remove all elements on remove.

GordonSmith:

### DIFF
--- a/widgets/src/map/GMap.js
+++ b/widgets/src/map/GMap.js
@@ -6,7 +6,7 @@
         root.GMap = factory(root.d3, root.SVGWidget, root.Graph, root.IGMap);
     }
 }(this, function (d3, SVGWidget, Graph, IGMap) {
-    function GMap(target) {
+    function GMap() {
         Graph.call(this);
         IGMap.call(this);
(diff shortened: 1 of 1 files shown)


GordonSmith:
@jbrundage, @mlzummo Please Review

mlzummo:
works, however the "widget-wrapper" (target element) has its style property modified and that is not reverted, i am not sure if that will be/is an issue

GordonSmith:
Its the other way round - the google maps change it to a bland color and I am simply removing it...

mlzummo:
image

mlzummo:
image

GordonSmith:
The only change I see there is the removal of the background-color?  (which was added by the google chart...)

mlzummo:
div->widget-wrapper should not have any styling after widget removal because it is only applied when the widget is rendered, I believe

GordonSmith:
I shall look further - but the "width:100%; height 100vh" looks an awful lot like my HelloWorld styling!

mlzummo:
the width and height is from the hello world, that should be the only styling left on removal.

PR #259, opened 2015-03-13
Fixes gauge data update

mlzummo:
@GordonSmith @jbrundage 

### DIFF
--- a/widgets/src/c3/Gauge.js
+++ b/widgets/src/c3/Gauge.js
@@ -41,7 +41,8 @@
         this.c3Chart.internal.config.gauge_label_format = this.value_format() === "Percent" ? null : function (value, ratio) { return value; };
         this.c3Chart.internal.config.gauge_label_show = this.show_labels();
         this.c3Chart.load({
-            columns: [[this._columns, this._data]]
+            columns: [[this._columns, this._data]],
+            unload:true
         });
     };
 


widgets/src/c3/Gauge.js
```diff
@@ -41,7 +41,8 @@
         this.c3Chart.internal.config.gauge_label_format = this.value_format() === "Percent" ? null : function (value, ratio) { return value; };
         this.c3Chart.internal.config.gauge_label_show = this.show_labels();
         this.c3Chart.load({
-            columns: [[this._columns, this._data]]
+            columns: [[this._columns, this._data]],
+            unload:true
```
GordonSmith: I am not sure what this is trying to fix?  Plus it does not look correct as per the API - if anything it should be "unload: [this._columns]"?

mlzummo:
I will try your suggestion, however i'm not sure if it will work. Basically the refactor of C3 loads the data as an addition to current data so that everytime you run .render() to update something it redraws the gauge on top of itself which depending on the value doesn't look right. ... try updating values from dermatology and you will see what I mean.

mlzummo:
this PR should be closed, the real underlying issue is that if you change the column name at the same time as the data, it appends to the data an additional data set. EX of problem:

    <script>
        require(["src/c3/Gauge"], function (C3Gauge) {
            var helloWorld = new C3Gauge()
                .target("helloWorld")
                .testData()
                //.data(50)
                .render()
            ;

            setTimeout(function() { helloWorld.columns('blue').data(30).render(); },3000);
        });
    </script>

PR #262, opened 2015-03-18
SYMP_AVE_HACK

GordonSmith:

### DIFF
--- a/widgets/src/marshaller/HipieDDL.js
+++ b/widgets/src/marshaller/HipieDDL.js
@@ -60,6 +60,17 @@
                 if (val === undefined) {
                     val = item[rhsKey.toLowerCase()];
                 }
+                //  Symposium AVE Hack 
+                if (val === undefined && rhsKey.indexOf("_AVE") === rhsKey.length - 4 && item.base_count !== undefined) {
+                    var rhsSum = rhsKey.substring(0, rhsKey.length - 4) + "_SUM";
+                    val = item[rhsSum];
+                    if (val === undefined) {
+                        val = item[rhsSum.toLowerCase()];
+                    }
+                    if (val) {
+                        val /= item.base_count;
+                    }
+                }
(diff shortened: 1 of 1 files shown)


GordonSmith:
@drealeed Please Review

jchambers-ln:
Code Looks good to me -- will need to test to be sure it does the trick.

widgets/src/marshaller/HipieDDL.js
```diff
@@ -60,6 +60,17 @@
...
+                if (val === undefined && rhsKey.indexOf("_AVE") === rhsKey.length - 4 && item.base_count !== undefined) {
+                    var rhsSum = rhsKey.substring(0, rhsKey.length - 4) + "_SUM";
+                    val = item[rhsSum];
+                    if (val === undefined) {
+                        val = item[rhsSum.toLowerCase()];
+                    }
+                    if (val) {
+                        val /= item.base_count;
+                    }
+                }
```
drealeed: will this handle summing multiple avges and dividing it by the sum of multiple base_counts for aggregated wuid-based charts?

drealeed:
If the answer to the above question is yes, reviewed & signed off

jchambers-ln:
I tested this in DSP and it seems to work rendered pie, line, choropleth using it with good results.

widgets/src/marshaller/HipieDDL.js
GordonSmith: I don't think so.  

This is simply taking the the SUM and dividing by the base_count on a per row basis.

In the example I looked at the aggregate (SUM) had already been burned into the WU Result?

GordonSmith:
@jchambers-ln Where the number correct though?

jchambers-ln:
I did a few spot checks and assuming hippie is calculating Sum and count correctly the math to get the average is correct.

PR #263, opened 2015-03-20
New testData example and changes to weight .. now split into weight & data

mlzummo:
@jbrundage @GordonSmith 

Please review this change.

after discovering that the palette colors object was doing what it designed to do but the data set being fed into it, (if it had a few counties/data sets that had large numbers while the others were small in comparison) would create too gradual a change of color for the majority of the map. This PR for fix works by taking the weight field and splitting it into weight and data. "data" is the actual data (previously weight) and "weight" is now a number that can be described as a category. The data is sorted smallest num to biggest and then broken up into X (in this case 12) categories from lowest to highest. These weight/categories from 1-12 are then fed into the palette object and provide a much better representation of the data.

Gordon what are your thoughts on this or a way to get the Palette object do do this for us?


GordonSmith:
@mlzummo I think there are more modifications in this pull request other than the test data change?

I think a logarithmic rainbow palette may your described scenario?

jbrundage:
You should remove the comments, the console.log calls and the "g_drag_hack" condition (It was solely for the purposes of the demo).

mlzummo:
made approp. changes... this also adds a new weight and data field to each data field.

GordonSmith:
This type of transformation belongs closer to the data pipeline than in a specific visualization.  A good example would be if you wanted to apply this ordering transform to a multichart / pie chart and a choropleth?

GordonSmith:
Closing for further discussions.

PR #271, opened 2015-03-25
Widget Resize Tweaks

GordonSmith:
Refactor and improve widget resizing, specifically the HTML Widget.

### DIFF
--- a/widgets/src/common/HTMLWidget.js
+++ b/widgets/src/common/HTMLWidget.js
@@ -11,6 +11,53 @@
     };
     HTMLWidget.prototype = Object.create(Widget.prototype);
 
+    HTMLWidget.prototype.calcFrameWidth = function (element) {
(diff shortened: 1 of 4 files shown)


GordonSmith:
@dineshshetye @jbrundage Please Review

jbrundage:
I would think clientWidth and clientHeight should be contentWidth and contentHeight?  Otherwise, looks good

GordonSmith:
"clientWidth/Height" - I am using those words because they match:  https://developer.mozilla.org/en-US/docs/Web/API/Element/clientHeight

GordonSmith:
The overlay size should always match the SVG size and scale.  In general none of our widgets should need a scrollbar - with the current exception of Table  + Grid Layout when set to "width" mode.

The change is explicitly setting it that way - so in future if you write a widget that should allow scrollbars it will need to change the default overlay (like Table is doing).

PR #283, opened 2015-04-06
Adds Google TreeMap

mlzummo:
@GordonSmith @jbrundage @gururx 

Please Review ... should we impliment data/testData such that it is interchangeable between Dendrogram because this is a "Tree"?

### DIFF
--- a/widgets/src/google/Common.js
+++ b/widgets/src/google/Common.js
@@ -26,6 +26,7 @@
 
     Common.prototype.publish("fontSize", null, "number", "Font Size");
     Common.prototype.publish("fontName", null, "string", "Font Name");
+    Common.prototype.publish("fontColor", null, "html-color", "Font Color");
 
     Common.prototype.publish("legendShow", true, "boolean", "Show Legend");
     Common.prototype.publish("legendAlignment", "center", "set", "Legend Alignment", ["", "start", "center", "end"]);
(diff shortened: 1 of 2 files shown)


widgets/src/google/Common.js
```diff
@@ -39,6 +40,11 @@
     Common.prototype.publish("animationDuration", 0, "number", "Animation Duration");
     Common.prototype.publish("animationOnStartup", true, "boolean", "Animate On Startup");
     Common.prototype.publish("animationEasing", "linear", "set", "Animation Easing", ["", "linear", "in", "out", "inAndOut"]);
+    
+    Common.prototype.publish("title", "", "string", "Text to display above the chart"); // does not support alignment (TODO make our own title functons)
+    Common.prototype.publish("titlePosition", null, "number", "Legend Font Size");
```
GordonSmith: Description looks wrong.

widgets/src/google/TreeMap.js
```diff
@@ -0,0 +1,87 @@
...
+
+        this._chartType = "TreeMap";
+    };
+    TreeMap.prototype = Object.create(CommonND.prototype);
+    // TODO impliment ITree?
+    
+    TreeMap.prototype.publish("minColor", "", "html-color", "");
+    TreeMap.prototype.publish("midColor", "", "html-color", "");
+    TreeMap.prototype.publish("maxColor", "", "html-color", "");
+    TreeMap.prototype.publish("headerHeight", "", "number", "");
```
GordonSmith: Should be "null" for no default.
Also they need a description

GordonSmith:
If it is a tree, then yes it should be interchangeable with the ITree.  Since the Tree structures are new, we may still tweak the "default" structure if needed.

PR #293, opened 2015-04-17
Adds AmCharts

GordonSmith:


GordonSmith:
Two main comments (otherwise I am happy enough to pull):
1.  Why are the src files in amcharts/charts/xxx - rather than just in amcharts?
2.  Your UMD declaration are now wrong (and we now need to start ensuring they are consistent (I will comment on one of them so you see what I mean.

Further - we are very close to having bower package management support in place - which means all the "libs" you are about to add will not be needed...

widgets/src/amcharts/charts/Scatter.js
```diff
@@ -0,0 +1,56 @@
+"use strict";
+(function(root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3/d3", "./CommonXY", "amcharts.xy", "../../chart/INDChart", "css!./Bar"], factory);
+    } else {
+        root.Scatter = factory(root.d3, root.CommonXY, root.amcharts, root.INDChart);
```
GordonSmith: Should be:  root.amcharts_Scatter = factory(root.d3, root.amcharts_CommonXY, root.*amcharts*, root.chart_INDChart);

I am not sure what the correct root is for amcharts - you would need to check the docs/samples...

mlzummo:
New Branch & PR this PR can be closed

PR #352, opened 2015-05-05
Adds C3 Publish Params (CSS)

mlzummo:
@GordonSmith @jbrundage Please Review

... more to come ideas welcome ...


src/c3chart/Area.js
```diff
@@ -13,6 +13,34 @@
...
+    Area.prototype.update = function (domNode, element) {
+
+        if (this.isStacked()) {
+            this.c3Chart.groups([this._columns.slice(1,this._columns.length)]);
+        } else {
+            this.c3Chart.groups([]);
+        }
+
+        CommonND.prototype.update.apply(this, arguments);
+        this.updateStyles.call(this,element);
```
GordonSmith: I would suggest removing all the "updateStyle(s)" methods and (in this instance) replace with:
  element.selectAll(".c3-line")
    .style({
        "stroke-width":  this.lineWidget() + "px,
        "stroke-opacity": this.lineOpacity(),
        "stroke-dasharray": this.dashedLine().toString()
    })
  ;

GordonSmith: also this.updateStyles.call(this,element); is the same as this.updateStyles(element);

mlzummo: i'm grouping them like this because i need to make sure i call everything after c3 has rendered (otherwise the style attributes get overwritten by c3) this way i know there is a function and that it permeates each class and is the same format across everything. this also allows me to group common.js or commonND publish in there respected files but call it from the [widget].js file

src/c3chart/hpcc-c3.css
```diff
@@ -0,0 +1,158 @@
+/*-- Chart --*/
```
GordonSmith: This file should be renamed to Common.css

PR #354, opened 2015-05-06
C3 publish params 4

mlzummo:
@GordonSmith Please Review


src/c3chart/Common.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "c3", "../common/HTMLWidget", "css!c3"], factory);
+        define(["d3", "c3", "../common/HTMLWidget", "css!./common"], factory);
```
GordonSmith: Should be "css!./Common" (IOW it should match the name of the JS file)

src/c3chart/Common.js
```diff
@@ -91,10 +99,55 @@
...
+        }
+
         this.c3Chart.resize({
             width: this.width(),
             height: this.height()
         });
+
+        this.c3Chart.load(this._c3LoadDataObj);
+
+        var chartOptions = this.getChartOptions();
```
GordonSmith: the getChartOptions should be all the options that go into the "load" - the styles will get applied in the normal way within the update method..

Issue #338, opened 2015-05-01
Refactor Published Property Names


GordonSmith:

GordonSmith:
Fixed by https://github.com/hpcc-systems/Visualization/pull/394

PR #358, opened 2015-05-06
C3 Fix & more publish params

mlzummo:
@GordonSmith Please Review

### DIFF
--- a/src/c3chart/Common.js
+++ b/src/c3chart/Common.js
@@ -1,7 +1,7 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "c3", "../common/HTMLWidget"], factory);
+        define(["d3", "c3", "../common/HTMLWidget", "css!c3"], factory);
     } else {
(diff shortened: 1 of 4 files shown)


src/c3chart/c3.css
```diff
@@ -0,0 +1,200 @@
+/*-- Chart --*/
```
GordonSmith: This file should not be included.

src/c3chart/Common.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "c3", "../common/HTMLWidget"], factory);
+        define(["d3", "c3", "../common/HTMLWidget", "css!./c3"], factory);
```
GordonSmith: Should be  "css!c3" (then it will use the CSS from the bower folder).

GordonSmith:
While your fixing the above - can you reset the default values of the published params to match what they where previously - specifically the dashed line setting for the line and area chart.

GordonSmith:
Also if you add "Fixes #338" to the commit message it will automatically close that Issue when the pull request makes it to master!

PR #368, opened 2015-05-07
Changes Google Params and .draw called in common.js

mlzummo:
@GordonSmith Please Review

### DIFF
--- a/src/google/Line.js
+++ b/src/google/Line.js
@@ -15,6 +15,15 @@
     Line.prototype = Object.create(CommonND.prototype);
     Line.prototype._class += " google_Line";
 
+    /**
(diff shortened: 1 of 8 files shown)


src/google/Line.js
```diff
@@ -26,76 +35,71 @@
 
     Line.prototype.publish("smoothLines", true, "boolean", "Causes chart data lines to draw smoothly");
```
GordonSmith: Can we change this to false?  and then I will pull...

mlzummo: yes, there is more changes coming down the line though ... will just wait for the next PR tonight before the weekend

PR #379, opened 2015-05-12
JSHint Support

GordonSmith:


GordonSmith:
@anmoljagetia, @mlzummo, @jbrundage  Please Review

src/amchart/Candle.js
```diff
@@ -13,13 +13,12 @@
...
-        this._categoryField;
-        this._lowField;
-        this._highField;
-    };
-    
+        this._openField = undefined;
+        this._closeField = undefined;
+        this._categoryField = undefined;
+        this._lowField = undefined;
+        this._highField = undefined;
```
jbrundage: Is it necessary to initialize variables in this way?

GordonSmith: ...no... But the previous code does nothing and I am assuming the only reason they are there is to "self document" - IMO if they are published params, then they should be removed altogether (but that should be in a separate PR)?

(Actually now that you point it out - if they are published params, then they are now just wrong  and really _need_ to be be deleted).

src/amchart/Pie.js
```diff
@@ -93,7 +92,7 @@
         
         this.pieAlpha().forEach(function(d,i) {
             if (typeof(this._chart.chartData[i])==='undefined') { 
-                this._chart.chartData[i] = {} 
+                this._chart.chartData[i] = {};
```
jbrundage: When should/shouldn't a semi-colon be used?

GordonSmith: In this specific case you need it at the end of every assignment (once this PR is pulled the answer will be "whenever JSHint tells you"<g>).

I think the JS rule is that it is needed to separate JS statements, which is why it is not needed  at the end of code blocks (function xxx() {}, if {} and switch {}) or for the the last statement within a code block.  FWIW JSHint will insist you have them for the last statement in a code block (which is good with me!).

src/amchart/Candle.js
mlzummo: actually we want to init those are empty arrays as those are private vars (not publish params)

GordonSmith: If you want to init them - then open a diff PR - the goal of this PR is to be JSHint compliant with no logic changes...

mlzummo:
looks good to me

Issue #405, opened 2015-05-21
Add Google Scatter, Timeline and TreeMap widgets


mlzummo:

mlzummo:
Closed via #409

PR #409, opened 2015-05-21
Add Google Scatter, Timeline and TreeMap widgets #405

dtsnell4:


mlzummo:
should squash the two commits into one .. also Travis failed ... need to do gulp lint and see what needs correcting?

demos/dermatology.html
```diff
@@ -87,7 +90,6 @@
                             <option value="src/amchart/Bar">AmChart Bar</option>
                             <option value="src/amchart/Bubble">AmChart Bubble</option>
                             <option value="src/amchart/Candle">AmChart Candle</option>
-                            <option value="src/amchart/FloatingColumn">AmChart Floating Column</option>
```
mlzummo: add this line back in there

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+    TreeMap.prototype.defaultlTooltip = function(row, size, value) {
+        var data =  TreeMap.prototype.mapData;
+        return '<div style="background:#ddd; padding:10px; border-style:solid" >' +
+           'Label: ' + data.getValue(row, 0) + '<br>' +
+           'Parent: ' + data.getValue(row, 1) + '<br>' +
+           'Column 3 Label: ' + data.getColumnLabel(2) + ', Value: ' + data.getValue(row, 2) + '<br>' +
+           'Column 4 Label: ' + data.getColumnLabel(3) + ', Value: ' + data.getValue(row, 3) + '<br>' +
+           'Datatable row #: ' + row + '<br>' +
+           data.getColumnLabel(2) +' (total value of this cell and its children): ' + size + '<br>' +
+	   data.getColumnLabel(3) + ': ' + value + ' </div>';      };
```
mlzummo: indentation?

dtsnell4:
@GordonSmith @jbrundage @mlzummo Please Review.

src/google/Scatter.js
```diff
@@ -0,0 +1,215 @@
...
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "./CommonND", "../common/HTMLWidget"], factory);
+    } else {
+        root.Scatter = factory(root.d3, root.google_CommonND, root.common_HTMLWidget);
+    }
+}(this, function (d3, CommonND, HTMLWidget) {
+
+    function Scatter() {
+        CommonND.call(this);
+        this._class = "google_Scatter";
```
GordonSmith: The _class should be moved into the prototype now (this is a recent-ish change) see https://github.com/hpcc-systems/Visualization/blob/master/src/google/Column.js#L16

src/google/Scatter.js
```diff
@@ -0,0 +1,215 @@
...
+        this._tag = "div";
```
GordonSmith: This is not needed as its set in google/Common.js

src/google/Scatter.js
```diff
@@ -0,0 +1,215 @@
...
+    Scatter.prototype.publish("selectionMode", 'single', "set", "When selectionMode is 'multiple', users may select multiple data points.",['single','multiple'],{tags:['Basic']});
+
+     /**
+     * Publish Params Common To Other Libraries
+     */
+    Scatter.prototype.publish("backgroundColor", null, "html-color", "The background color for the main area of the chart. Can be either a simple HTML color string, for example: 'red' or '#00cc00', or an object with the following properties.",null,{tags:['Basic']});
+    
+    Scatter.prototype.publish("dataOpacity", 1.0, "number", "The transparency of data points, with 1.0 being completely opaque and 0.0 fully transparent. This refers to the visible data (i.e. dots).",null,{tags:['Basic']});
+    
+    Scatter.prototype.publish("tooltipIsHtml", "true", "boolean", "Set to false to use SVG-rendered (rather than HTML-rendered) tooltips.",null,{tags:['Advanced']});
```
GordonSmith: true shouldn't have quotes.

src/google/Scatter.js
```diff
@@ -0,0 +1,215 @@
...
+        
+        retVal.crosshair = {
+            color: this.crosshairColor(),
+            orientation: this.crosshairOrientation(),
+            opacity: this.crosshairOpacity(),
+            trigger: this.crosshairTrigger()
+        };
+
+        retVal.hAxis = {};
+        retVal.vAxis = {};
```
GordonSmith: Minor:  In other places you put the values straight into the structure (like textStyle below).

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+        define(["d3", "../common/HTMLWidget", "goog!visualization,1,packages:[timeline]"], factory);
+    } else {
+        root.Timeline = factory(root.d3, root.common_HTMLWidget);
+    }
+}(this, function (d3, HTMLWidget) {
+
+    function Timeline() {
+        HTMLWidget.call(this);
+        
+        this._class = "google_Timeline";
```
GordonSmith: See above comment about _class

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+        this._tag = "div";
```
GordonSmith: See above comment about _tag

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+        this._chartType = "Timeline";
+        this._data_google = [];
+
+        this.columns([]);
```
GordonSmith: Not needed.

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+        this.data([]);
```
GordonSmith: Not needed.

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+    Timeline.prototype = Object.create(HTMLWidget.prototype);
+
+    /**
+     * Publish Params Common To Other Libraries
+     */
+
+    Timeline.prototype.publish("tooltipIsHtml", "true", "boolean", "Set to false to use SVG-rendered (rather than HTML-rendered) tooltips. See Customizing Tooltip Content for more details.",null,{tags:['Advanced']});
```
GordonSmith: no quotes for "true"

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+    
+    Timeline.prototype.publish("tooltipTrigger", "focus", "set", "The user interaction that causes the tooltip to be displayed: 'focus' - The tooltip will be displayed when the user hovers over the element; 'none' - The tooltip will not be displayed.",['none', 'focus'],{tags:['Basic']});
+
```
GordonSmith: For consistency no blank line between each publish (I tend to use the blank line to group different properties as appropriate.

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+    Timeline.prototype.publish("timelineGroupByRowLabel", true, "boolean", "If set to false, creates one row for every dataTable entry. The default is to collect bars with the same row label into one row.",null,{tags:['Basic']});
+ 
+    Timeline.prototype.publish("timelineShowBarLabels", true, "boolean", "If set to false, omits bar labels. The default is to show them.", null,{tags:['Basic']});
+    
+    Timeline.prototype.publish("timelineShowRowLabels", true, "boolean", "If set to false, omits row labels. The default is to show them.",null,{tags:['Basic']});
+    
+    Timeline.prototype.publish("timelineSingleColor", null, "string", "Colors all bars the same. Specified as a hex value (e.g., '#8d8').",null,{tags:['Basic']});
+    
+    Timeline.prototype.getChartOptions = function () {
+        var retVal = [];
```
GordonSmith: Should this default to the base class chart Options?

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+        retVal.timelineShowRowLabels = this.timelineShowRowLabels();
+        retVal.timelineSingleColor = this.timelineSingleColor();
+        retVal.tooltipIsHtml = this.tooltipIsHtml();
+        retVal.tooltipTrigger = this.tooltipTrigger();
+
+        return retVal;
+    };
+
+    Timeline.prototype.enter = function (domNode, element) {
+        element.style("overflow", "hidden");
```
GordonSmith: You should be calling down to the base class here (and then this call is made there)?

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "../common/HTMLWidget", "goog!visualization,1,packages:[timeline]"], factory);
+    } else {
+        root.Timeline = factory(root.d3, root.common_HTMLWidget);
+    }
+}(this, function (d3, HTMLWidget) {
+
+    function Timeline() {
+        HTMLWidget.call(this);
```
GordonSmith: Should this be derived from google/Common.js?

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+        this._chart = new google.visualization[this._chartType](domNode);
+    };
+
+    Timeline.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+       
+        this._chart.draw(this._data_google, this.getChartOptions());
+    };
+    Timeline.prototype.data = function (_) {
+        this._data = _;
```
GordonSmith: You need to test if arguments.length === 0 and return the data.  See: https://github.com/hpcc-systems/Visualization/blob/master/src/map/ChoroplethCountries.js#L44-L64 for an example of what I mean

src/google/Timeline.js
```diff
@@ -0,0 +1,109 @@
...
+        element.style("overflow", "hidden");
+        this._chart = new google.visualization[this._chartType](domNode);
+    };
+
+    Timeline.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+       
+        this._chart.draw(this._data_google, this.getChartOptions());
+    };
+    Timeline.prototype.data = function (_) {
```
GordonSmith: For consistencey we normally have any of the setters/getters above the enter/update methods (minor)

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "../common/HTMLWidget", "goog!visualization,1,packages:[treemap]"], factory);
+    } else {
+        root.TreeMap = factory(root.d3, root.common_HTMLWidget);
+    }
+}(this, function (d3, HTMLWidget) {
```
GordonSmith: Should this be derived from google/Common.js?  If not why? (I am curious)

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+
+    function TreeMap() {
+        HTMLWidget.call(this);
+        
+        this._class = "google_TreeMap";
+        this._tag = "div";
```
GordonSmith: See above comments re class + tag

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+        this._chartType = "TreeMapChart";
+        this._data_google = [];
+        
+        this.columns([]);
+        this.data([]);
+    }
+
```
GordonSmith: No blank line (I am trying to keep all the "constructor" code together)

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+        retVal.noHighlightColor = this.noHighlightColor();
+        retVal.showScale = this.showScale();
+        retVal.showTooltips = this.showTooltips();
+        retVal.useWeightedAverageForAggregation = this.useWeightedAverageForAggregation();
+        retVal.generateTooltip = TreeMap.prototype.defaultlTooltip;
+        return retVal;
+    };
+
+    TreeMap.prototype.enter = function (domNode, element) {
+        element.style("overflow", "hidden");
```
GordonSmith: If needed  - it should be done _after_ the base class call.

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+    };
+
+    TreeMap.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+        TreeMap.prototype.mapData = this._data_google;
+        this.treemapChart.draw(this._data_google, this.getChartOptions());
+    };
+    
+    TreeMap.prototype.data = function (_) {
+        if (arguments.length) {
```
GordonSmith: See above comment(s) re data.  Typically all setters/getters will return the current value if no arguments are passed in:
So ```var data = this.data();``` should work as expected...

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+            var arr = this._columns.concat(this._data);
+            this._data_google = new google.visualization.arrayToDataTable(arr);
+        }
+    };
+
+    TreeMap.prototype.testData = function () {
+        var context = this;
+        function setColumns(_callback) {
+            context.columns([['Location', 'Parent', 'Market trade volume (size)', 'Market increase/decrease (color)']]);
+            _callback();
```
GordonSmith: No need for callbacks here as there is no async going on IOW:

```
this.columns(...)
this.data(...)
```

Should be fine...

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+
+    TreeMap.prototype.enter = function (domNode, element) {
+        element.style("overflow", "hidden");
+        HTMLWidget.prototype.enter.apply(this, arguments);
+        this.treemapChart = new google.visualization.TreeMap(element.node());
+    };
+
+    TreeMap.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+        TreeMap.prototype.mapData = this._data_google;
```
GordonSmith: This doesn't look correct...

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+        
+        setColumns( function() {
+            setData();
+        });   
+ 
+        return this;
+    };
+    
+    TreeMap.prototype.defaultlTooltip = function(row, size, value) {
+        var data =  TreeMap.prototype.mapData;
```
GordonSmith: This can be simply:  ```var data = this._data_google;``` or simply use this._data_google inside the return string.

src/google/TreeMap.js
```diff
@@ -0,0 +1,158 @@
...
+        return '<div style="background:#ddd; padding:10px; border-style:solid" >' +
+           'Label: ' + data.getValue(row, 0) + '<br>' +
```
GordonSmith: IOW this could be:  ```'Label: ' + this._data_google.getValue(row, 0) + '<br>' +```

src/google/Scatter.js
dtsnell4: I took this block of code straight from another widget.  Since it is used in several different widgets the same way, Matt and I were discussing putting it into a common file in a future commit.  Is it OK if I change it at that point?

src/google/TreeMap.js
dtsnell4: @GordonSmith because Timeline and TreeMap each require their specific google packages ("goog!visualization,1,packages:[treemap]") Matt said it would be best to derive these from HTMLWidget.  I deferred to his experience, as I do to yours....  If there is a better way to do this, please explain.

src/google/TreeMap.js
dtsnell4: @GordonSmith `this` in this context does not refer to TreeMap.  This function is called in by the "generateTooltip" option which is specific to this widget. [docs](https://developers.google.com/chart/interactive/docs/gallery/treemap#Tooltips)  Therefore `this._data_google;`  is undefined here.   Because this option automatically passes arguments (row, size, value), I have no way to pass the TreeMap context into my function to get _data_google.  As a result I came up with adding it to TreeMap.prototype.mapData, which is accessible here.  Hope that makes sense, if you you can think of a better way to get _data_google into the function please let me know.

GordonSmith: Would the following fix that?:
Replace:
```
retVal.generateTooltip = TreeMap.prototype.defaultlTooltip;
```

To:
```
var context = this;
retVal.generateTooltip = function() {
  TreeMap.prototype.defaultlTooltip.apply(this, arguements);
}
```

dtsnell4: Yes it does, Matt was just helping me do something very similar.  I am updating it now.  Thanks

mlzummo:
fails serialization test, not sure why yet

GordonSmith:
@dtsnell4 - FYI we don't get any notifications if you simply update your branch, so if you want a re-review you need to mention us in the comments...

src/google/TreeMap.js
```diff
@@ -0,0 +1,156 @@
...
+
+    TreeMap.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+        this.treemapChart.draw(this._data_google, this.getChartOptions());
+    };
+
+    TreeMap.prototype.data = function (_) {
+        var retVal = HTMLWidget.prototype.data.apply(this, arguments);
+        if (arguments.length) {
+            this._data = _;
```
GordonSmith: This will already have happened via line 96

GordonSmith:
This isn't working in https://rawgit.com/dtsnell4/Visualization/Google_Widgets/demos/dermatology.html (but could be my dial up connection here).

Otherwise it looks ok, we need to agree on a tree structure and move the test data into api/ITree.

dtsnell4:
@GordonSmith Sorry about that, I will remember to do that in the future.  

https://rawgit.com/dtsnell4/Visualization/Google_Widgets/demos/dermatology.html is working on my workstation, @mlzummo @jbrundage can you confirm?

mlzummo:
@GordonSmith @dtsnell4 I think we should just use the google tree structure and convert the rest of the widgets over. (I dont think it will take but a day or two to do that). ... dermatology is also working must be the connection

GordonSmith:
Working now - just the one comment that should be addressed (the this._data = _ one) and I will pull.

dtsnell4:
OK that is done and pushed.

PR #414, opened 2015-05-26
changes to partially sync svn

mlzummo:
@jbrundage @GordonSmith Please Review

### DIFF
--- a/src/api/INDChart.js
+++ b/src/api/INDChart.js
@@ -12,13 +12,28 @@
 
     //  Data ---
     INDChart.prototype.testData = function () {
-        this.columns(["Subject", "Year 1", "Year 2", "Year 3"]);
-        this.data([
-            ["Geography", 75, 68, 65],
-            ["English", 45, 55, 52],
-            ["Math", 98, 92, 90],
-            ["Science", 66, 60, 66]
-        ]);
+        switch (this._chartType) {
+            case 'ScatterChart':
+                this.columns(["ID", "Year 1", "Year 2", "Year 3"]);
+                this.data([
+                    [10, 75, 68, 65],
+                    [20, 45, 55, 52],
+                    [30, 98, 92, 90],
+                    [40, 66, 60, 86]
+                ]);
+                break;
+                
+            default:
+                this.columns(["Subject", "Year 1", "Year 2", "Year 3"]);
+                this.data([
+                    ["Geography", 75, 68, 65],
+                    ["English", 45, 55, 52],
+                    ["Math", 98, 92, 90],
+                    ["Science", 66, 60, 66]
+                ]);
+                break;
+        }
+       
         return this;
     };
 
--- a/src/chart/Pie.js
+++ b/src/chart/Pie.js
@@ -31,7 +31,7 @@
     Pie.prototype._class += " chart_Pie";
     Pie.prototype.implements(I2DChart.prototype);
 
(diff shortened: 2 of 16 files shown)


src/api/INDChart.js
```diff
@@ -12,13 +12,28 @@
...
     //  Data ---
     INDChart.prototype.testData = function () {
-        this.columns(["Subject", "Year 1", "Year 2", "Year 3"]);
-        this.data([
-            ["Geography", 75, 68, 65],
-            ["English", 45, 55, 52],
-            ["Math", 98, 92, 90],
-            ["Science", 66, 60, 66]
-        ]);
+        switch (this._chartType) {
```
GordonSmith: Does this work for charts with "_chartType" of undefined?

src/chart/Pie.js
```diff
@@ -31,7 +31,7 @@
     Pie.prototype._class += " chart_Pie";
     Pie.prototype.implements(I2DChart.prototype);
 
-    Pie.prototype.publish("paletteID", "default", "set", "Palette ID", Pie.prototype._palette.switch());
+    Pie.prototype.publish("paletteID", "default", "set", "Palette ID", Pie.prototype._palette.switch(),{tags:['Basic','Shared']});
```
GordonSmith: @jbrundage What does the "Shared" mean?

src/marshaller/HipieDDL.js
```diff
@@ -52,17 +52,17 @@
 
     SourceMappings.prototype.doMap = function (item) {
         var retVal = [];
-        for (var key in this.mappings) {
-            try {
+        try {
+            for (var key in this.mappings) {
```
GordonSmith: You are rolling back newer changes...

GordonSmith:
At least one change was reverting a newer commit - can you double check the others?

src/api/INDChart.js
mlzummo: yes it does, just tested

src/marshaller/HipieDDL.js
mlzummo: i wasn't sure about that one, i didn't know if we wantted the try in the loop or outside, reverting will make the try catch in the loop

GordonSmith: Yes - I switched that way so I could provide better error reporting

src/chart/Pie.js
mlzummo: I believe we are going with its shared among 1 or more widgets (not ness. across libs)

mlzummo:
@GordonSmith  i "cherry picked" all the changes from SVN -> GIT so, we should be good to go .. that one change that reverted was me cherry picking that change.

src/chart/Pie.js
GordonSmith: Can this not be calculated on the fly?  Otherwise it will for ever be accidentally wrong (what is it actually used for?)

GordonSmith:
@mlzummo Are these "fixes" for 1.0.0 or are they new features for 1.1.0?

mlzummo:
@GordonSmith Fixes for 1.0.0, i rebased this against master though, i want to make sure this doesn't break anything on 1.0.0 ... which branch can i rebase against ?

GordonSmith:
Use candidate-1.0.2 for bug fixes to 1.0.0 and use master for the "next" build

PR #428, opened 2015-05-27
Adds Auto ItemsPerPage (TABLE/PAGINATOR) GH-145

mlzummo:
@GordonSmith @jbrundage @anmoljagetia Please Review

### DIFF
--- a/src/other/Table.js
+++ b/src/other/Table.js
@@ -48,6 +48,41 @@
         this.tbody = this.table.append("tbody");
     };
 
+    Table.prototype._generateTempCell = function() {
+        var trow = this.tbody.selectAll("tr").data([[0]]);
+        trow
+            .enter()
+            .append("tr")
+        ;
+        var tcell = trow.selectAll("td").data(function (row, i) {
+            return row;
+        });
+        tcell.enter()
+            .append("td")
+            .text(function (d) {
+                return d;
+            })
+        ;
+        tcell.exit()
+            .remove()
+        ;
+        return tcell;
+    };
+
+    Table.prototype._calcRowsPerPage = function(th) {
+        if (this._paginator.numItems() === 0) { // only run on first render
+            this._paginator.numItems(1);
+            this.itemsPerPage(1);
+            this._paginator.render();
+        }
+
+        var thHeight = this.calcHeight(th);
+        var tcellHeight = this.calcHeight(this._generateTempCell());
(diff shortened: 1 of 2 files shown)


src/other/Table.js
```diff
@@ -96,22 +96,59 @@
             .remove()
         ;
 
+        var theadHeight = parseInt(th.style('height'));
+        var theadMarginPadding = parseInt(th.style('padding-top')) + parseInt(th.style('padding-bottom')) + parseInt(th.style('margin-top')) + parseInt(th.style('margin-bottom'));
```
GordonSmith: If you look in HTMLWidget there are some calcHeight functions which do this

GordonSmith:
I would suggest moving the entire "calcHeight" piece into its own method (like we do in other cases).

GordonSmith:
The other thought I had was within the calcHeight function, you could simply create a new instance of the TableWidget - render it with no rows, measure height, render with one row, measure height, destroy widget.

mlzummo:
can you show an example of where we move the calcHeight blocks of code into there own functions... i think using this.calcHeight did the trick enough?

GordonSmith:
If you add a new method called "calcRowsPerPage" or such like and them move all you calculation code into that and just call it once, it would be cleaner - XYAxis does something similar:  https://github.com/hpcc-systems/Visualization/blob/master/src/chart/XYAxis.js#L58-L80

mlzummo:
@GordonSmith made suggested changes ... Please Review

GordonSmith:
If you rollup the two commits I can comment on the individual lines.

There is one console.log that can be removed.

I suspect that if and empty data set is passed in your tmpCell will be displayed?  I think you should "remove" it once you have measured its height?

Also I would have expected all the measurement code to go into the other a diff method to keep the main update call clean.

mlzummo:
@GordonSmith ill take some suggestions around how to make the code leaner/cleaner, this is the structure i came up with but if there is a better approach can you provide some pseudo code around a better method of keeping the update call clean?

GordonSmith:
What I was trying to get at would have been (in the update method):  var maxNumRows = this.calcMaxRows(x, y, z, ...);

mlzummo:
@GordonSmith Please Review ... made suggested changes

mlzummo:
it should auto remove the tcell as well even though i don't explicitly call .remove() on it due to later updates?

GordonSmith:
Looks good.

Issue #430, opened 2015-05-27
amchart pyramid margin left and right needs to be auto calculated 


mlzummo:
amchart pyramid margin left and right needs to be auto calculated as to fit chat in smaller space and not havr to define margin for labels to show

PR #448, opened 2015-06-03
Fixes for Amcharts GH-430

mlzummo:
@GordonSmith @dtsnell4 @jbrundage 

Please Review 

### DIFF
--- a/src/amchart/CommonFunnel.js
+++ b/src/amchart/CommonFunnel.js
@@ -29,8 +29,9 @@
     CommonFunnel.prototype.publish("flip", true, "boolean", "Flip Chart",null,{tags:['Intermediate']});
     CommonFunnel.prototype.publish("reverseDataSorting", false, "boolean", "Reverse Data Sorting",null,{tags:['Intermediate']});
 
-    CommonFunnel.prototype.publish("marginLeft", null, "number", "Margin (Left)",null,{tags:['Intermediate']});
-    CommonFunnel.prototype.publish("marginRight", 150, "number", "Margin (Right)",null,{tags:['Intermediate']});
+    CommonFunnel.prototype.publish("marginLeft", 0, "number", "Margin (Left)",null,{tags:['Intermediate']});
+    CommonFunnel.prototype.publish("marginRight", 0, "number", "Margin (Right)",null,{tags:['Intermediate']});
+
     CommonFunnel.prototype.publish("marginTop", null, "number", "Margin (Top)",null,{tags:['Intermediate']});
     CommonFunnel.prototype.publish("marginBottom", null, "number", "Margin (Bottom)",null,{tags:['Intermediate']});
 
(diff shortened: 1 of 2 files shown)


mlzummo:
let me know if you would like me to reissue against 1.0.4 or if you can just down merge

src/amchart/CommonFunnel.js
```diff
@@ -141,6 +147,20 @@
         domNode.style.height = this.size().height + 'px';
 
         this._palette = this._palette.switch(this.paletteID());
+
+        switch (this.labelPosition()) {
+            case "left":
+                this._chart.marginLeft += parseInt(domNode.getBoundingClientRect().width/8);
```
GordonSmith: Why dividing by 8?

src/amchart/CommonXY.js
```diff
@@ -214,6 +214,7 @@
             theme: "none",
             type: "xy",
             automargins: false,
+            path: "//www.amcharts.com/lib/3/",
```
GordonSmith: What is the "path" for?

src/amchart/Pyramid.js
```diff
@@ -33,8 +33,8 @@
             ["Math", 98],
             ["Science", 66]
         ]);
-        this.marginLeft(15);
-        this.marginRight(150);
+        //this.marginLeft(15);
+        //this.marginRight(150);
```
GordonSmith: Remove commented code.

GordonSmith:
You will need to reissue for 1.0.4 and also can you add "Fixes GH-nnn" for each of the issues it fixes?

src/amchart/CommonXY.js
mlzummo: For some reason, and i spoke with amcharts to ... the AmCharts funnel object cant discover the correct path automatically like it should for the images so this tells it where the base path is so we dont get a 404 on the image load (scrollbar imgs etc...). AmCharts was not able to reproduce with a basic usage of requirejs and loading there lib. But I cant seem to find the root cause ... If you can shed any light i'd welcome a second opinion/option

GordonSmith: Do you have an URL I can reproduce with?

mlzummo: actually, the problem doesn't seem to be occurring in dermatology.html (i thought i had checked that) but is in playground.html and theme editor 

http://rawgit.com/mlzummo/Visualization/PLAYGROUND_RGBC/demos/playground.html

so perhaps were doing something wrong here... the error gives us an undefined in in the img url if you you load up AmChart Scatter or AmChart Bubble

src/amchart/CommonFunnel.js
mlzummo: no reason in particular it just seems to be a number that works

mlzummo:
@GordonSmith Please take a look again i removed the "path" becuase i cant reproduce the issue now

src/amchart/CommonFunnel.js
GordonSmith: Ideally each widget should draw to the max extent available and then it will be the responsibility of the container (the surface) to allocate padding consistently for all widgets?

GordonSmith:
I am not sure if ```*Fixes for GH-430``` will automatically close the issue for us - but ```Fixes GH-430``` will (I would need to check the Github docs to see what they accept).

mlzummo:
is that better?

GordonSmith:
I thought you were going to re-issue it for 1.0.4?

src/amchart/CommonFunnel.js
mlzummo: Yes, but with the labels amcharts but this difficult for the funnel widgets

mlzummo:
LGTM

mlzummo:
closing, reissuing against diff branch

PR #451, opened 2015-06-05
Adds Form widget and Input widget

jbrundage:
@GordonSmith @mlzummo @gururx Please Review


src/form/Form.css
```diff
@@ -0,0 +1,45 @@
...
+    padding:5px;
+}
+
+.form_Form input {
+    padding:0px 0px;
+}
+
+.form_Form tfoot button{
+    margin-left:5px;
+}
```
GordonSmith: Indentation is inconsistent in this file (mixed / tabs and spaces) - preference is for 4 spaces instead of tabs.

src/form/Form.js
```diff
@@ -0,0 +1,170 @@
...
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "../common/HTMLWidget", "./Input", "css!./Form"], factory);
+    } else {
+        root.form_Form = factory(root.d3, root.common_HTMLWidget,root.form_Input);
+    }
+}(this, function (d3,HTMLWidget,Input) {
+    function Form() {
+        HTMLWidget.call(this);
+        this._class = "form_Form";
```
GordonSmith: Class definition should use the "new" format - see any other widget.

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+        }
+        if(typeof(this._blur) !== 'function'){
+            this._blur = function(){};
+        }
+        if(typeof(this._click) !== 'function'){
+            this._click = function(){};
+        }
+        if(typeof(this._change) !== 'function'){
+            this._change = function(){};
+        }
```
mlzummo: why is the typeof and adding of default necessary here can this be moved below into the .blue and .click and .change functions?

mlzummo:
@GordonSmith will it be the job of the marshaller to take DDL and create the approp. input obj to pass in?

src/form/Form.js
```diff
@@ -0,0 +1,170 @@
...
+            if(!inp.isValid()){
+                msgArr.push("'"+inp.label()+"'"+' value is invalid.');
+            }
+        });
+        if(msgArr.length > 0){
+            alert(msgArr.join('\n'));
+            ret = false;
+        }
+        return ret;
+    };
```
mlzummo: look into firing an event? and being able to pass this data with it so that end user can customize the way the popup looks

src/form/Form.js
```diff
@@ -0,0 +1,170 @@
...
+        this
+            .inputs([
+                new Input()
+                    .id("textbox-test")
+                    .label("Alphanumeric")
+                    .type("textbox")
+                    .validate("^[A-Za-z0-9]+$")
+                    .value("SomeString123"),
+                new Input()
+                    .id("textbox-test")
```
GordonSmith: ID's should be unique - if you don't set the ID then it will be assigned a unique one automatically.

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "../common/HTMLWidget", "css!./Input"], factory);
+    } else {
+        root.form_Form = factory(root.common_HTMLWidget);
+    }
+}(this, function (d3,HTMLWidget) {
+    function Input() {
+        HTMLWidget.call(this);
+        this._class = "form_Input";
```
GordonSmith: Class should use new scheme (see any other widget - apart from Form)

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+
+        this._tag = "div";
+    }
+    Input.prototype = Object.create(HTMLWidget.prototype);
+    Input.prototype.publish("id", "", "string", "HTML id for the input");
```
GordonSmith: This clashes with Widget.id() and Widget._id - will need to be renamed (then my comment above will be invalid).

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+                this._inputElement = element.append("button");
+                break;
+            default:
+                this._inputElement = element.append("input").attr('type',this.type());
+                break;
+        }
+    };
+    Input.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+        this._inputElement.attr('id', this.id());
```
GordonSmith: Not needed

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+            optionHTML += '<option>selectOptions not set</option>';
+        }
+        this._inputElement.node().innerHTML = optionHTML;
+    };
+    
+    Input.prototype.checkNodeName = function (expected,elm) {
+        var node = this._inputElement.node();
+        if(node.nodeName !== expected){
+            node.remove();
+            this.enter(null,elm);
```
GordonSmith: "enter" should only ever be called once (and its called by the framework)

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+                break;
+        }
+    };
+    Input.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+        this._inputElement.attr('id', this.id());
+        
+        switch(this.type()){
+            case 'select':
+                this.checkNodeName('SELECT',element);
```
GordonSmith: I think its ok to say that "type" can only be set once and has to be set before the first render (we have other examples of this - we should add a flag to the publish property to indicate this?)

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+        return true;
+    };
+    
+    Input.prototype.blur = function (_) {
+        if (!arguments.length){
+            this._blur();
+        } else {
+            this._blur = _;
```
GordonSmith: I am not sure what your trying to achieve with the "_blur"?  In general this "Event" function would be just an empty stub, which is called when the input event files - it is up to the user of this widget to wire it up?

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+    Input.prototype.insertSelectOptions = function (optionsArr) {
+        var optionHTML = '';
+        if(optionsArr.length > 0){
+            optionsArr.forEach(function(opt){
+                optionHTML += '<option value="'+opt+'">'+opt+'</option>';
+            });
+        } else {
+            optionHTML += '<option>selectOptions not set</option>';
+        }
+        this._inputElement.node().innerHTML = optionHTML;
```
GordonSmith: These calls can be shortened to:  this._inputElement.html(optionHTML);

src/form/Input.js
```diff
@@ -0,0 +1,159 @@
...
+                this.checkNodeName('INPUT',element);
+                this._inputElement.property('checked', this.value());
+                break;
+            default:
+                this.checkNodeName('INPUT',element);
+                this._inputElement.attr('value', this.value());
+                this._inputElement.attr('type', this.type());
+                break;
+        }
+        if(typeof(this._blur) !== 'function'){
```
GordonSmith: I think here you should be doing:
```
   this._inputElement.on("blur", function(d) {context.blur(d);});
```
type of thing?s

src/form/Form.css
jbrundage: Will fix

src/form/Form.js
jbrundage: Will fix

src/form/Input.js
jbrundage: Yes, good point, I'll move this

src/form/Form.js
jbrundage: Yes, the user should have more control over the validation displayed.  Though it's my understanding that events will eventually become publish parameters?  It seems more appropriate to support customization with that feature.

GordonSmith:
@jbrundage - Can you look at my comment re "enter"?

GordonSmith:
Closing in favour of:  https://github.com/hpcc-systems/Visualization/pull/480

PR #462, opened 2015-06-08
Refactors X and Y to Data & Column in XYAxis

anmoljagetia:
@GordonSmith Please Review.
Refactors XYAxis to have dataAxis and columnAxis instead of xAxis and yAxis.

### DIFF
--- a/src/chart/XYAxis.js
+++ b/src/chart/XYAxis.js
@@ -10,22 +10,22 @@
         SVGWidget.call(this);
         this._drawStartPos = "origin";
 
-        this._xScale = "";
+        this._dataScale = "";
         this.parseDate = d3.time.format("%Y-%m-%d").parse;
     }
     XYAxis.prototype = Object.create(SVGWidget.prototype);
 
     XYAxis.prototype.publish("orientation", "horizontal", "set", "Selects orientation for the axis", ["horizontal", "vertical"]);
 
-    XYAxis.prototype.xScale = function (_) {
-        if (!arguments.length) return this._xScale;
-        this._xScale = _;
+    XYAxis.prototype.dataScale = function (_) {
+        if (!arguments.length) return this._dataScale;
+        this._dataScale = _;
         return this;
     };
 
     XYAxis.prototype.enter = function (domNode, element) {
         this.x = null;
-        switch (this._xScale) {
+        switch (this._dataScale) {
             case "DATE":
                 this.x = d3.time.scale();
                 break;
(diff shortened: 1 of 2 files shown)


src/chart/XYAxis.js
```diff
@@ -63,18 +63,18 @@
 
         var test = element.append("g");
 
-        var svgXAxis = test.append("g")
+        var svgDataAxis = test.append("g")
```
GordonSmith: This is not a "dataAxis" as it sometimes contains columns.  It _is_ an XAxis...

src/chart/XYAxis.js
```diff
@@ -63,18 +63,18 @@
...
             .attr("class", this.orientation() === "horizontal" ? "x axis" : "y axis")
             .attr("transform", "translate(0," + height + ")")
-            .call(this.orientation() === "horizontal" ? this.xAxis : this.yAxis)
+            .call(this.orientation() === "horizontal" ? this.dataAxis : this.columnAxis)
         ;
-        var svgYAxis = test.append("g")
+        var svgColumnAxis = test.append("g")
```
GordonSmith: Same for this:   This is not a "columnAxis" as it sometimes contains columns. It is an YAxis...

src/chart/XYAxis.js
```diff
@@ -49,10 +49,10 @@
 
         this.svg = element.append("g");
         this.svgData = this.svg.append("g");
-        this.svgXAxis = this.svg.append("g")
+        this.svgDataAxis = this.svg.append("g")
```
GordonSmith: This is "XAxis" as it is always horizontal.

src/chart/XYAxis.js
```diff
@@ -49,10 +49,10 @@
...
             .attr("class", this.orientation() === "horizontal" ? "x axis" : "y axis")
         ;
-        this.svgYAxis = this.svg.append("g")
+        this.svgColumnAxis = this.svg.append("g")
```
GordonSmith: This is This is "YAxis" as it is always vertical.

GordonSmith:
You need to refactor the "x" and "y" in the CSS also.

mlzummo:
I thought we were keeping X and Y axis?

GordonSmith:
@mlzummo These charts can now swap orientation (think bar / column) the styles are directed at the "data values" and "columns" rather than the x / y axis, which is why we are renaming some of the variable / class names (Note:  there are still some true x / y variables which should remane).

GordonSmith:
Looks good.

Issue #495, opened 2015-06-14
chart/Scatter


GordonSmith:
Add a new chart type "scatter"

See http://bl.ocks.org/mbostock/3887118 for an example implemented in d3

PR #504, opened 2015-06-18
GH-495 Adds the Scatter Chart

anmoljagetia:
@GordonSmith @mlzummo Please review.

Demo at : http://rawgit.com/anmoljagetia/Visualization/scatterPlot/demos/dermatology.html?src/chart/Scatter
Fixed GH-495
Signed-off-by: Anmol Jagetia <anmoljagetia@gmail.com>

### DIFF
--- a/src/chart/Scatter.css
+++ b/src/chart/Scatter.css
@@ -1,0 +1,9 @@
+.chart_Scatter .axis path,
+.chart_Scatter .axis line {
+  fill: none;
+  stroke: #000;
+  shape-rendering: crispEdges;
+}
+line.point, line.point1 {
+	stroke-width: 2;
+}
--- a/src/chart/Scatter.js
+++ b/src/chart/Scatter.js
@@ -1,0 +1,182 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "../common/SVGWidget", "./XYAxis", "../api/INDChart", "css!./Scatter"], factory);
+    } else {
+        root.chart_Column = factory(root.d3, root.common_SVGWidget, root.chart_XYAxis, root.api_INDChart);
+    }
+}(this, function (d3, SVGWidget, XYAxis, INDChart) {
+    function Scatter(target) {
+        XYAxis.call(this);
+        INDChart.call(this);
+    }
+    Scatter.prototype = Object.create(XYAxis.prototype);
+    Scatter.prototype._class += " chart_Scatter";
+    Scatter.prototype.implements(INDChart.prototype);
+
+    Scatter.prototype.publish("paletteID", "default", "set", "Palette ID", Scatter.prototype._palette.switch(),{tags:['Basic','Shared']});
+    Scatter.prototype.publish("pointShape", "cross", "set", "Shape of the data points", ["circle", "rectangle", "cross"]);
+    Scatter.prototype.publish("pointSize", 8, "number", "Point Size");
+
+    Scatter.prototype.updateChart = function (domNode, element, margin, width, height) {
+        var context = this;
+
+        this._palette = this._palette.switch(this.paletteID());
+
+        if (this._prevPointShape !== this.pointShape()) {
+            this.svgData.selectAll(".data").remove();
+            this._prevPointShape = this.pointShape();
+        }
+
+        var column = this.svgData.selectAll(".data")
+            .data(this._data, function (d) { return d; })
+        ;
+
+        column.enter().append("g")
+            .attr("class", "data")
+        ;
+
+        column
+            .each(function (d, i) {
+                var element = d3.select(this);
+                var title;
+                var title1;
+                var cx, cy, x, y, x1, y1, x2, y2;
+                var point = element.selectAll(".point").data(d.filter(function(d, i) {return i > 0;}));
+                var point1 = element.selectAll(".point1").data(d.filter(function(d, i) {return i > 0;}));
+                switch (context.pointShape()) {
+                    case "circle" :
+                        title = point
+                          .enter().append("circle")
+                            .attr("class", "point")
+                            .on("click", function (d2, idx) {
+                                context.click(context.rowToObj(d), context._columns[idx + 1]);
+                            })
+                            .append("title")
+                        ;
+                        break;
+
+                    case "rectangle" :
+                        title = point
+                          .enter().append("rect")
+                            .attr("class", "point")
+                            .on("click", function (d2, idx) {
+                                context.click(context.rowToObj(d), context._columns[idx + 1]);
+                            })
+                            .append("title")
+                        ;
+                        break;
+
+                    case "cross" :
+                        title = point
+                          .enter().append("line")
+                            .attr("class", "point")
+                            .on("click", function (d2, idx) {
+                                context.click(context.rowToObj(d), context._columns[idx + 1]);
+                            })
+                            .append("title")
+                        ;
+                        title1 = point1
+                            .enter().append("line")
+                            .attr("class", "point1")
+                            .on("click", function (d2, idx) {
+                                context.click(context.rowToObj(d), context._columns[idx + 1]);
+                            })
+                            .append("title1")
+                        ;
+                        break;
+                }
+                if (context.orientation() === "horizontal") {
+                    switch (context.pointShape()) {
+                        case "circle" :
+                            cx = "cx";
+                            cy = "cy";
+                            break;
+
+                        case "rectangle" :
+                            x = "x";
+                            y = "y";
+                            break;
+
+                        case "cross" :
+                            x1 = "x1";
+                            x2 = "x2";
+                            y1 = "y1";
+                            y2 = "y2";
+                            break;
+                    }
+                } else {
+                   switch (context.pointShape()) {
+                       case "circle" :
+                           cx = "cy";
+                           cy = "cx";
+                           break;
+
+                       case "rectangle" :
+                           x = "y";
+                           y = "x";
+                           break;
+
+                       case "cross" :
+                           x1 = "y1";
+                           x2 = "y2";
+                           y1 = "x1";
+                           y2 = "x2";
+                           break;
+                   }
+                }
+
+                switch (context.pointShape()) {
+                    case "circle" :
+                        point.transition()
+                            .attr("r", context.pointSize()/2)
+                            .attr(cx, function (d2, idx) { return context.dataScale(d[0]) + context.dataScale.rangeBand()/2 ;})
+                            .attr(cy, function (d2, idx) { return context.valueScale(d2); })
+                            .style("fill", function (d2, idx) { return context._palette(context._columns[idx + 1]); })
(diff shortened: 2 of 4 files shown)


src/chart/Scatter.css
```diff
@@ -0,0 +1,6 @@
+.axis path,
+.axis line {
```
GordonSmith: Should be prefixed with the scatter class ID

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+        define(["d3", "./XYAxis", "../api/INDChart", "css!./Scatter"], factory);
+    } else {
+        root.chart_Column = factory(root.d3, root.chart_XYAxis, root.api_INDChart);
+    }
+}(this, function (d3, XYAxis, INDChart) {
+    function Scatter(target) {
+        XYAxis.call(this);
+        INDChart.call(this);
+    }
+
```
GordonSmith: Remove blank line

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+                var point = element.selectAll(".point").data(d.filter(function(d, i) {return i > 0;}));
+                var point1 = element.selectAll(".point").data(d.filter(function(d, i) {return i > 0;}));
+                switch (context.dotRepresentation()) {
+                    case "circle" :
+                        var title = point
+                          .enter().append("circle")
+                            .attr("class", "point")
+                            .attr("r", context.pointSize()/2)
+                            .on("click", function (d2, idx) {
+                                context.click(context.rowToObj(d), context._columns[idx + 1]);
```
mlzummo: idx not defined (its i right now)

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+        XYAxis.call(this);
+        INDChart.call(this);
+    }
+
+    Scatter.prototype = Object.create(XYAxis.prototype);
+    Scatter.prototype._class += " chart_Scatter";
+    Scatter.prototype.implements(INDChart.prototype);
+
+    Scatter.prototype.publish("paletteID", "default", "set", "Palette ID", Scatter.prototype._palette.switch(),{tags:['Basic','Shared']});
+    Scatter.prototype.publish("dotRepresentation", "cross", "set", "Selects the representation for the data", ["circle", "rectangle", "cross"]);
```
GordonSmith: rename to pointShape and improve description "Point Shape" or such like.

src/chart/Scatter.js
anmoljagetia: The idx refers to the function defined on line 46. The d element has the row from the array, and then d2 is individual elements from the array (Found with the help of idx).

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+    Scatter.prototype.publish("pointSize", 8, "number", "pointSize");
```
GordonSmith: Needs a better description @mlzummo

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+
+    Scatter.prototype.updateChart = function (domNode, element, margin, width, height) {
+        var context = this;
+
+        this._palette = this._palette.switch(this.paletteID());
+
+        var column = this.svgData.selectAll(".data")
+            .data(this._data, function (d) { var k = d.filter(function(d, i) {return i > 0;}); return k + Math.random(); })
```
GordonSmith: The "key" function here is worrying, the "random" will probably force all the points to be removed and appended on _every_ update call.  Can you delete this and I can show you a better way to handle the "shape" changing issue.

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+
+        column.enter().append("g")
+            .attr("class", "data")
+        ;
+
+        column
+            .each(function (d, i) {
+                var element = d3.select(this);
+                var point = element.selectAll(".point").data(d.filter(function(d, i) {return i > 0;}));
+                var point1 = element.selectAll(".point").data(d.filter(function(d, i) {return i > 0;}));
```
GordonSmith: You will want to give this a different class name otherwise it could interfere with the other one.

src/chart/Scatter.js
GordonSmith: Its defined on line 46

src/chart/Scatter.js
GordonSmith: Another approach would have been to add a single point of type "g" and append two lines within that.

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+                switch (context.dotRepresentation()) {
+                    case "circle" :
+                        var title = point
+                          .enter().append("circle")
+                            .attr("class", "point")
+                            .attr("r", context.pointSize()/2)
```
GordonSmith: This should be set in the "update" cycle not the "enter()" cycle (and it only appears to be working now due to the random above)

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+                            .append("title")
+                        ;
+                        break;
+
+                    case "rectangle" :
+                        var title = point
+                          .enter().append("rect")
+                            .attr("class", "point")
+                            .attr("width", context.pointSize())
+                            .attr("height", context.pointSize())
```
GordonSmith: width + height should be applied at the update cycle.

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+                            })
+                            .append("title")
+                        ;
+                        break;
+
+                    case "cross" :
+                        var title = point
+                          .enter().append("line")
+                            .attr("class", "point")
+                            .attr("stroke-width", 2)
```
GordonSmith: Can this be moved into the style sheet?

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+                            .on("click", function (d2) {
+                                context.click(context.rowToObj(d), context._columns[idx + 1]);
+                            })
+                            .append("title")
+                        ;
+                        var title1 = point1
+                            .enter().append("line")
+                            .attr("class", "point")
+                            .attr("stroke-width", 2)
```
GordonSmith: Can this be moved into the style sheet?

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+                        break;
+                }
+
+                if (context.orientation() === "horizontal") {
+                    switch (context.dotRepresentation()) {
+                        case "circle" :
+                            point.transition()
+                                .attr("cx", function (d2, idx) { return context.dataScale(d[0]) + context.dataScale.rangeBand()/2 ;})
+                                .attr("cy", function (d2, idx) { return context.valueScale(d2); })
+                                .style("fill", function (d2, idx) { return context._palette(idx + 1); })
```
GordonSmith: All calls to "_palette" should use the column label and not the idx so that the colours are consistent with other charts displaying the same data...

src/chart/Scatter.js
```diff
@@ -0,0 +1,171 @@
...
+                                .attr("x2", function (d2, idx) { return context.valueScale(d2) - context.pointSize()/2; })
+                                .style("stroke", function (d2, idx) { return context._palette(idx + 1); })
+                            break;
+                    }
+                }
+
+                title
+                    .text(function (d2, idx) { return d[0] + " (" + d2 + ")" + ": " + context._columns[idx + 1]; })
+                ;
+                point.exit().remove();
```
GordonSmith: point1 will need a remove also?

anmoljagetia: Added the removal of point1 also.

src/chart/Scatter.js
anmoljagetia: If I move this to style sheet, I will have to color the stroke with every other shape, and the size estimation we have now, will also change. Also this is only relevant for the cross shape.

src/chart/Scatter.js
mlzummo: ah i missed that, odd it was giving me an error about idx ...

src/chart/Scatter.js
```diff
@@ -0,0 +1,184 @@
...
+
+        this._palette = this._palette.switch(this.paletteID());
+
+        if (this._prevPointShape !== this.pointSize()) {
+            this.svgData.selectAll(".data").remove();
+            this._prevPointShape = this.pointSize();
+        }
+
+        var column = this.svgData.selectAll(".data")
+            .data(this._data, function (d) { var k = d.filter(function(d, i) {return i > 0;}); return k; })
```
GordonSmith: The whole key function is not needed:  ```, function (d) { var k = d.filter(function(d, i) {return i > 0;}); return k; }```

src/chart/Scatter.js
```diff
@@ -0,0 +1,184 @@
...
+    Scatter.prototype.publish("paletteID", "default", "set", "Palette ID", Scatter.prototype._palette.switch(),{tags:['Basic','Shared']});
+    Scatter.prototype.publish("pointShape", "cross", "set", "Shape of the data points", ["circle", "rectangle", "cross"]);
+    Scatter.prototype.publish("pointSize", 8, "number", "Point Size");
+
+    Scatter.prototype.updateChart = function (domNode, element, margin, width, height) {
+        var context = this;
+
+        this._palette = this._palette.switch(this.paletteID());
+
+        if (this._prevPointShape !== this.pointSize()) {
```
GordonSmith: Should be ```pointShape()```

src/chart/Scatter.js
```diff
@@ -0,0 +1,184 @@
...
+            this.svgData.selectAll(".data").remove();
+            this._prevPointShape = this.pointSize();
```
GordonSmith: Should be ```pointShape()```

src/chart/Scatter.js
```diff
@@ -0,0 +1,184 @@
...
+                            })
+                            .append("title")
+                        ;
+                        break;
+
+                    case "cross" :
+                        title = point
+                          .enter().append("line")
+                            .attr("class", "point")
+                            .attr("stroke-width", 2)
```
GordonSmith: Can this be a style?

src/chart/Scatter.js
```diff
@@ -0,0 +1,184 @@
...
+                            .on("click", function (d2, idx) {
+                                context.click(context.rowToObj(d), context._columns[idx + 1]);
+                            })
+                            .append("title")
+                        ;
+                        title1 = point1
+                            .enter().append("line")
+                            .attr("class", "point1")
+                            .attr("stroke-width", 2)
```
GordonSmith: Can this be a style?

anmoljagetia: Adding this to style will add a stroke on everything else as well, i.e. circles and also the rectangles. In that case, we will have to resize everything, and also handle the strokes on circle and rectangles separately.

GordonSmith: Not if you added it as "line.point, line.point1 {...}"

PR #506, opened 2015-06-18
Graph Annotations

GordonSmith:

### DIFF
--- a/src/marshaller/HTML.js
+++ b/src/marshaller/HTML.js
@@ -23,7 +23,7 @@
     };
 
     HTML.prototype.testData2 = function () {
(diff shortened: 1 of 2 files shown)


GordonSmith:
@mlzummo Please Review

mlzummo:
should we have put/pass a version string for the button Annotation code? J/Dan implemented the form input widget to render in a surface which looked nice for a demo, not 100% sure how it would work in the long run but would allow even more customization from the looks of it...

GordonSmith:
Not sure I follow?  These annotations are on the GraphVertex?

mlzummo:
your right i was looking at this end of day, and was seeing words similar to what I had in the surface button annotations so i jumped the gun a bit...

mlzummo:
looks good, is there a way to test this at the moment?

GordonSmith:
Rename testData2 to testData and open src/marshaller/HTML in dermatology

PR #511, opened 2015-06-22
GH-508  XAxis Scales

GordonSmith:
Add support for linear scale.
Add support for time scale.
Publish "sample" data property.

Fixes GH-508

### DIFF
--- a/src/chart/Column.js
+++ b/src/chart/Column.js
@@ -9,22 +9,14 @@
     function Column(target) {
         XYAxis.call(this);
         I2DChart.call(this);
+
+        this._linearGap = 25;
     }
     Column.prototype = Object.create(XYAxis.prototype);
     Column.prototype._class += " chart_Column";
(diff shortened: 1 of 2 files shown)


GordonSmith:
@mlzummo @anmoljagetia Please Review

mlzummo:
looks good, however with the separation of testData from the widgets comming, are you sure you want to start the trend of having a new sample data publish param now, doesn't seem appropriate? (though it does make it easy for testing/demoing)

mlzummo:
... unless this is going to tie into the new testData format later on down the road?

anmoljagetia:
Looks good. The sampleData has a blank category before the three options, should we keep that ?

GordonSmith:
@mlzummo I had the same thought - but since it will get replaced by something new, its a step in the right direction?

GordonSmith:
@anmoljagetia - Yes as the default should be no sample data.

PR #533, opened 2015-06-26
Adds src/layout/Border (proof of concept)

jbrundage:
@GordonSmith Please review...

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -180,6 +180,7 @@
                     sourceWidget = sourceWidget || currWidget;
                     Persist.clone(sourceWidget, function (widget) {
                         serializationSurface
+                            .surfacePadding(0)
                             .widget(widget)
                             .render()
                         ;
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -91,6 +91,7 @@
                             <option value="src/layout/Surface">Surface Layout</option>
                             <option value="src/layout/Cell">Cell Layout</option>
                             <option value="src/layout/Grid">Grid Layout</option>
+                            <option value="src/layout/Border">Border Layout</option>
                             <option value="src/amchart/Area">AmChart Area</option>
                             <option value="src/amchart/Bar">AmChart Bar</option>
                             <option value="src/amchart/Bubble">AmChart Bubble</option>
--- a/src/form/Form.js
+++ b/src/form/Form.js
@@ -16,7 +16,7 @@
 
     Form.prototype.publish("validate", true, "boolean", "Enable/Disable input validation");
     Form.prototype.publish("inputs", [], "widgetArray", "Array of input widgets");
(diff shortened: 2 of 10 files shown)


GordonSmith:
@dtsnell4 @jbrundage Visually this layout looks correct.  From the framework architecture point of view my thoughts are:
1 We are about to have a PopupSurface - what if we also had an "AbsoluteSurface" this would be one which would be positioned at specific x, y, w, h (maybe its the same as the popup surface, except the popup surface has a "modal" true option?
2 This new Surface would then replace your Border Cell
3 The Border layout should then contain _all_ the logic for the positioning of these new surfaces.
4 We would then re-work the Grid Layout to also use the Absoulute Surface instead of its Cell (again moving the cell row, col, rowSpan, colSpan "knowledge" into the Grid Layout
5 We will probably want the Border Layout to also support "splitters" so the user can visually re-size the border cells - design mode again.

4 + 5 would be done as separate pull requests.

Thoughts?

jbrundage:
1&2 How soon will the PopupSurface/AbsoluteSurface be ready for use? (With a fresh, Monday morning, set of eyes... I realize that I'm only using BorderCell.js to store the x,y,w,h)
3 This is nearly true already, I'll update this pull request with this change.
4 This makes sense.
5 Splitters? Should the user be able to split a Surface into two Surfaces? Should the user be able to drag a Surface onto another Surface to swap their locations/sizes?

What do you think about my logic in Border.prototype.borderLayoutObject (line 66)? I gave each Cell's placement variables two attributes, px and % ("%" being a percentage of either the parent container's height or width). It seems like something that could be reused.
For example:
If a parent element has "width:1000px" and a child cell has width = {"px":150,"%":50}. Then the cell would have "width:650px".

dtsnell4:
As for the pop surface, I showed Gordon what I did with the Popup button widget I made Friday.  It’s very basic and simply uses the existing Layout Surface.  Any modal/nonmodal/positioning etc. can be done from the button widget so a specific Popup Surface widget technically is not necessary… unless there is some specific functionality it needs that I am not aware of, in which case just let me know and I’ll get on it.  Otherwise I’m on hold with the popup widget for now, playing with ideas for the Tabbed widget.

GordonSmith:
@dtsnell4 We still need a separate "PopupSurface" (which may be the same thing as an AbsoluteSurface, but with a modal option) as it will be used without a button in some situations.

dtsnell4:
OK, so should I be working on that or am I waiting on something (e.g. absolute surface)...?

GordonSmith:
@dtsnell4 - Yes I think if you carry on with developing the PopupSurface (for now) the above are really just "thoughts" rather than "todos" but with the goal of getting us all thinking in the same direction.

dtsnell4:
Will do.

Issue #516, opened 2015-06-24
Add "popup dialog" support


GordonSmith:
I would imagine this will need (but am allowed to be wrong):
* layout/PopupSurface
* layout/DropButton (which implements the IInput API and opens an instance of the PopupSurface, positioned below it)

PR #544, opened 2015-07-02
Popup widget #516

dtsnell4:
@GordonSmith @mlzummo @jbrundage 

So I have refactored the Popup widget from button-controlled widget to a layout. 

Dermatology: http://rawgit.com/dtsnell4/Visualization/PopupWidget_516/demos/dermatology.html?src/layout/Popup

Basically redesigned so that you simply call yourwidget.updateState(true/false) to display and hide the widget. The test data creates a test button to do it, but with this implementation you can do it from anywhere.

Some points to consider... I can think of more advanced options, but these are the basics for me:
1.	Should we not load the widget until it is displayed the first time? If the end user never displays the popup, should we include that overhead on page load?
2.	What options should be included for closing the popup? A close button (X) in the top right? Clicking anywhere outside the popup (i.e. a modal dialog)? Include the future Titlebar widget? In my mind this could get complicated. Currently the button toggles the popup, but that’s not practical UX.
3.	Should I include a modal state option, i.e. no interaction with the page until the modal is dismissed (optional background mask)?
4.	As for positioning, should we include a default "centered" option and do the calculations automatically for the dev?

Please let me know your thoughts/comments.

### DIFF
--- a/src/layout/Popup.js
+++ b/src/layout/Popup.js
@@ -1,0 +1,138 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "../common/HTMLWidget", "../layout/Surface", "../form/Input", "../c3chart/Column", "css!./Popup", "css!font-awesome",], factory);
+    } else {
+        root.layout_Popup = factory(root.d3, root.common_HTMLWidget, root.layout_Surface, root.form_Input, root.c3chart_Column);
+    }
+}(this, function (d3, HTMLWidget, Surface, Input, Column) {
+    function Popup() {
+        HTMLWidget.call(this);
+
+        this._tag = "div";
+
+        this._surfaceButtons = [];
+    }
+    Popup.prototype = Object.create(HTMLWidget.prototype);
+    Popup.prototype._class += " layout_Popup";
+    
+    Popup.prototype.publish("popupState", false, "boolean", "State of the popup, visible (true) or hidden (false)",null,{});
+    Popup.prototype.publish("top", null, "number", "Top position property of popup",null,{});
+    Popup.prototype.publish("bottom", null, "number", "Bottom position property of popup",null,{});
+    Popup.prototype.publish("left", null, "number", "Left position property of popup",null,{});
+    Popup.prototype.publish("right", null, "number", "Right position property of popup",null,{});
+    Popup.prototype.publish("position", "relative", "set", "Value of the 'position' property",["absolute", "relative", "fixed", "static", "initial", "inherit" ],{tags:['Private']});
+    Popup.prototype.publish("showTestButton", false, "boolean", "Test Button in target",null,{});
+    Popup.prototype.publish("buttonName", null, "string", "Button name property",null,{});
+    Popup.prototype.publish("buttonLabel", "", "string", "Button label text",null,{});
+    Popup.prototype.publish("widget", null, "widget", "Widget",null,{tags:['Private']});
+
+
+    Popup.prototype.testData = function () {
+        this.showTestButton(true);
+        this.position("absolute");
+        this.top(30);
+        
+        var context = this;
+        this._testButton = new Input()
+            .type("button")
+            .value("Open Popup")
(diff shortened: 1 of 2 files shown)


src/layout/Popup.js
```diff
@@ -0,0 +1,157 @@
...
+        this._tag = "div";
+
+        this._surfaceButtons = [];
+    }
+    Popup.prototype = Object.create(HTMLWidget.prototype);
+    Popup.prototype._class += " layout_Popup";
+    
+    Popup.prototype.publish("showTestButton", false, "boolean", "Test Button in target",null,{});
+    Popup.prototype.publish("popupState", false, "boolean", "State of the popup, visible (true) or hidden (false)",null,{});
+    Popup.prototype.publish("width", null, "number", "Popup width",null,{});
```
GordonSmith: width and height are already used in the base class Widget - overriding them will probably cause pain.

src/layout/Popup.js
```diff
@@ -0,0 +1,157 @@
...
+    Popup.prototype.update = function (domNode, element) {
+        HTMLWidget.prototype.update.apply(this, arguments);
+        var context = this;        
+
+        element
+            .style("position",this.position())
+            .style("left",this.left() + "px")
+            .style("right",this.right() + "px")
+            .style("top",this.top() + "px")
+            .style("bottom",this.bottom() + "px")
```
GordonSmith: If these styles were published for the existing surface, would it "work" as a the popup surface?

dtsnell4: I suppose it would

dtsnell4: I am curious though, why does this widget need anything other than the div.popupWidget that is created as a container?  If the developer wants a surface or titlebar or buttons, etc. they have all those widgets at their disposal.  I feel like I might be missing something...

src/layout/Popup.js
```diff
@@ -0,0 +1,145 @@
...
+        ;
+        
+        if (this.popupState()) {
+            element.style("display", "block");
+        } else {
+            element.style("display", "none");
+        }
+//
```
mlzummo: remove comment

src/layout/Popup.js
```diff
@@ -0,0 +1,145 @@
...
+    Popup.prototype.publish("right", null, "number", "Right position property of popup",null,{});
+    Popup.prototype.publish("position", "relative", "set", "Value of the 'position' property",["absolute", "relative", "fixed", "static", "initial", "inherit" ],{tags:['Private']});
+    
+    Popup.prototype.publish("showTestButton", false, "boolean", "Test Button in target",null,{});
+    Popup.prototype.publish("buttonName", null, "string", "Button name property",null,{});
+    Popup.prototype.publish("buttonLabel", "", "string", "Button label text",null,{});
+
+    Popup.prototype.publish("widget", null, "widget", "Widget",null,{tags:['Private']});
+
+
```
mlzummo: remove extra spaces

src/layout/Popup.js
```diff
@@ -0,0 +1,145 @@
...
+                    callback(widget);
+                }
+            }
+        });
+        
+        return this;
+    };
+
+    Popup.prototype.click = function(obj) {
+        console.log('Clicked: ' + obj.id);
```
mlzummo: use double quotes everywhere possible (even in publish param tag)

src/layout/Popup.js
```diff
@@ -0,0 +1,145 @@
...
+            this.popupState(visible).update(this._domNode, this._element);
+    };
+    
+    Popup.prototype.enter = function (domNode, element) {
+        HTMLWidget.prototype.enter.apply(this, arguments);
+
+        if (this._testButton) {
+            this._testButton.target(domNode.parentNode).render();
+        }
+        this._domNode = domNode;
```
mlzummo: prolly a better way to get this information than restoring it in a private var?

dtsnell4: Probably but I could not think of one.  Suggestions are welcomed.

PR #545, opened 2015-07-02
Tabbed content issue 518

dtsnell4:
@GordonSmith @jbrundage @mlzummo 

Similar to the Popup widget, here is a basic implementation of the Tabbed widget.  I used an array to pass the data similar to how we did the Surface Buttons.  

Demo here: http://rawgit.com/dtsnell4/Visualization/Tabbed_Content_518/demos/dermatology.html?src/layout/TabbedContent

The problem I have with that approach is passing the widget in the array.  If you inspect the console when running the demo, you will see an error "Uncaught TypeError: Converting circular structure to JSON" which is caused by the widget.  However everything runs and is displayed fine and works as expected, but I'm obviously uncomfortable with that.  If I pass the widget as a publish parameter that error goes away obviously, but that is not practical for this implementation with an unknown number of tabs/widgets.  If anyone has any suggestions on another way to do this please let me know, as well as other comments.

### DIFF
--- a/src/layout/TabbedContent.js
+++ b/src/layout/TabbedContent.js
@@ -1,0 +1,122 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3", "src/common/HTMLWidget", "src/api/IInput", "src/layout/Surface", "src/chart/Pie", "src/chart/MultiChart", "src/google/Line", "css!./TabbedContent"], factory);
+    } else {
+        root.form_Form = factory(root.d3, root.common_HTMLWidget, root.layout_IInput, root.layout_Surface, root.chart_Pie, root.chart_MultiChart, root.chart_Line);
+    }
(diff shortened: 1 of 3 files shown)


src/layout/TabbedContent.js
```diff
@@ -0,0 +1,135 @@
...
+        HTMLWidget.call(this);
+
+        this._tag = "div";
+        this._tabButtons = [];
+    }
+    TabbedContent.prototype = Object.create(HTMLWidget.prototype);
+    TabbedContent.prototype._class += " layout_TabbedContent";
+
+    TabbedContent.prototype.publish("width", null, "number", "Popup content width",null,{});
+    TabbedContent.prototype.publish("height", null, "number", "Popup content height",null,{});
```
GordonSmith: No need to publish width and height (and they are used in the base class) - this widget should always expand to fit the div it is being placed into (like all the other widgets).

dtsnell4: Yes, I had those in there for my own testing, I removed them.

PR #551, opened 2015-07-04
Allow inputs to be placed side by side.

GordonSmith:
Add option to drop empty fields from values() call.

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -246,25 +246,19 @@
                     })
                 ;
             }
-            resize = function () {
-                if (mainDiv) {
-                    mainDiv.style.width = window.innerWidth - 16 + "px";
-                    mainDiv.style.height = window.innerHeight - 16 - 40 + "px";
-                }
-
-                frame
+            frame.doResize = Surface.prototype.debounce(function () {
+                mainDiv.style.width = window.innerWidth - 16 + "px";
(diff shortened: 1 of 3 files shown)


GordonSmith:
@jbrundage Please Review

jbrundage:
Looks good, but the "Button Test:" label cell is unnaturally wide in this example. Maybe we should force the label cells to consume the least amount of space possible?
https://rawgit.com/GordonSmith/Visualization/FORM_LAYOUT/demos/dermatology.html?src/form/Form

GordonSmith:
@jbrundage Sounds good - how?

GordonSmith:
Note:  Serialization test is not working...

GordonSmith:
@jbrundage Please re-review (I added a new common/WidgetArray.js to fix the serialization issue).
Also I think we expanding the prompts + inputs correctly (inpputs class have a width of 100%), its just in this example none of the three inputs have any sensible way to expand...

jbrundage:
The Form table isn't resizing itself when its container is resized.

jbrundage:
The 'Button Test:' cell width is so wide that it's awkwardly far from the input it's describing.  Maybe we could wrap these labels in a <span> and use .getBoundingClientRect().width on all of the label spans to determine how wide the label cells should be (setting each label cell width to the width of the widest label cell span for each column).

jbrundage:
Looks good

PR #557, opened 2015-07-08
Adds Border Layout

jbrundage:
@GordonSmith please review


src/layout/BorderCell.js
```diff
@@ -0,0 +1,25 @@
...
+        Surface.prototype.enter.apply(this, arguments);
+        element.classed("layout_Surface", true);
+    };
+
+    BorderCell.prototype.update = function (domNode, element) {
+        Surface.prototype.update.apply(this, arguments);
+    };
+
+    return BorderCell;
+}));
```
GordonSmith: This isn't needed any more?  It doesn't add anything to regular Surface?

GordonSmith:
If you set the top border to 10 (or less) something weird happens...

jbrundage:
The weirdness is a result of Surface.js not being able to comprehend height and/or width of 0 or less.  Should I fix this as part of the Border layout pull request or handle it independently?

Issue #518, opened 2015-06-24
Add "Tabbed Surface"


GordonSmith:
Similar to MultiChart (but do not use that as a reference) allows several charts to be displayed in on location, selectable via a "Tabbing" paradigm.

Should use:  https://github.com/hpcc-systems/Visualization/issues/519

PR #560, opened 2015-07-09
Issue #518: Tabbed Content, basic implementation.

dtsnell4:
Basic layout of tabs and content panels

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -126,6 +126,9 @@
     <div id="main" style="width:100%; height:100vh;margin-top:40px;padding:8px;overflow:hidden">
     </div>
     <script>
+        var g_widget;
+        var g_persist;
+        var g_serial;
         var def = window.location.search.split("?")[1];
         if (def) {
             document.getElementById("widgetSelect").value = def;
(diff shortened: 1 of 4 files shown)


dtsnell4:
@GordonSmith I have no idea why this keeps failing, its complaining it can't find a file because it has a double 'src/src' in the path which is understandable.  But no where in my code do I have that directory doubled like that.  I'm at a loss.  It passes on my local environment. Same with PR #544

GordonSmith:
I think you should restructure it so that the user can add/remove tabs at runtime - as it stands you must add all your tabs before the first call to render.

Also I don't think the serialization test in dermatology is going to work - for that you would probably need to change your annotations array to be a WidgetArray.

dtsnell4:
@GordonSmith Is there an example of something using a Widget Array?

GordonSmith:
The Grid uses one:      
```
Grid.prototype.publish("content", [], "widgetArray", "widgets",null,{tags:['Private']});
```

dtsnell4:

I changed it to use the widget array as you suggested (Todo: create a removeTab() method).  It works fine in Dermatology, however it still errors on the serialization test.  I think I know why, but I don't quite know what to do about it.  I need to look into how Persist and the Property editor work.  In the meantime if anyone sees what I am doing wrong or has any suggestions, please let me know.

Issue #541, opened 2015-07-01
TitleWidget


GordonSmith:
A hybrid Title/Toolbar to replace the existing HTML Surface Title.

```
[(i)          My Title              <w><w><w>]
```

Type of thing.

mlzummo:
@GordonSmith I am thinking of just closing this thoughts? ... J has a pretty cool toolbar that uses the grid i had a version that worked ok but nothing that can replace the title ....

GordonSmith:
I am fine with that

PR #564, opened 2015-07-10
GH-541 Adds Toolbar Widget

mlzummo:

### DIFF
--- a/src/layout/Toolbar.js
+++ b/src/layout/Toolbar.js
@@ -1,11 +1,11 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "../common/HTMLWidget", "../form/Input", "./Cell", "../common/Text", "../common/Icon", "../chart/Pie", "./Surface", "css!./Toolbar"], factory);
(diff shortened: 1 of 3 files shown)


mlzummo:
@GordonSmith @jbrundage Please Review ... preliminary

mlzummo:
prolly going to change obj.type to just read the widget class

mlzummo:
@dtsnell4

mlzummo:
will widgetarray publish param work with an array ofobjects that contain a property that contains a widget among other properties?

src/layout/Toolbar.js
```diff
@@ -0,0 +1,103 @@
...
+                    obj.widget.target(this);
+                }
+                if (obj.type === "button") {
+                    obj.widget.target(this).render(function(widget) {
+                        widget._inputElement.style("display","inline-block");
+                        widget._element.style("width",obj.width+"px");
+                        widget._inputElement.style("height",obj.height+"px");
+                    });
+                }
+                d3.select(this).style("padding-left",(idx > 0 ? (context.gutter()/2)+"px" : null));
```
jbrundage: These publish parameters should be applied outside of the context of .enter() so that they'll be applied whenever these publish param values change (instead of solely applying on the initial rendering of each data node).

PR #591, opened 2015-07-21
Fixes #590 - Adds support for form default value in DDL

### LINKED ISSUE #590: DDL Add Default property data propogation to form Inputs

mlzummo:
@GordonSmith @jchambers-ln @drealeed @jbrundage @dtsnell4 

Please Review

### DIFF
--- a/src/marshaller/HipieDDL.js
+++ b/src/marshaller/HipieDDL.js
@@ -59,7 +59,7 @@
                 if (val === undefined) {
                     val = item[rhsKey.toLowerCase()];
                 }
-                //  Symposium AVE Hack 
+                //  Symposium AVE Hack
                 if (val === undefined && rhsKey.indexOf("_AVE") === rhsKey.length - 4 && item.base_count !== undefined) {
                     var rhsSum = rhsKey.substring(0, rhsKey.length - 4) + "_SUM";
                     val = item[rhsSum];
@@ -480,6 +480,7 @@
         this.dashboard = dashboard;
         this.id = visualization.id;
         this.label = visualization.label;
+        this.default = visualization.default || "";
         this.title = visualization.title || visualization.id;
         this.type = visualization.type;
         this.icon = visualization.icon || {};
@@ -566,6 +567,7 @@
                                 .name(field.id)
                                 .label((field.properties ? field.properties.label : null) || field.label)
                                 .type("textbox")
+                                .value(field.properties.default)
                             ;
                         }))
                     ;


mlzummo:
http://rawgit.com/mlzummo/Visualization/GH-590-FORM_DDL_DEFAULT/demos/marshaller/index.html?Hostname=10.173.147.1&Port=8010&Protocol=http&Wuid=W20150720-160015&ResultName=leeddx_issue_983_formvalueref_Comp_Ins026_DDL

drealeed:
Looks good to me--tested with other DDLs and it's working

mlzummo:
need a way to do a render callback for future date so that form can be submited on default value entered on init.

mlzummo:
or around line: https://github.com/hpcc-systems/Visualization/blob/master/src/marshaller/HipieDDL.js#L646

create a callback function that checks what a widget is (surface, grid, etc) and checks if its a form and if its a form calls submit

mlzummo:
a hack that could be used (but if it takes longer than 250ms to load the dash it wont work) 
setTimeout(function() { widget.submit(); }, 250)

drealeed:
now that gordon's pointed out the render feature, do you want to rework this change to take advantage of that?

mlzummo:
@GordonSmith @drealeed @jbrundageln

I already knew about this, I needed it specifically the way i implemented
it as i can specify a callback function without having to render on the
spot (without a target too) ... there are several other advantages as well
and Jaman agrees, However that's not to say there isn't a better way using
Gordon's notification methods to get the form to submit on default value on
initial render, I just do not know of any other method currently.

GordonSmith:
@drealeed @mlzummo The setting up of default values is fine - but an "auto" submit feature does not seem to be correct (as a concept) at this implementation level:
*  As it stands there will be an additional call to the server (after the initial call with refresh = true).
*  What happens if there are two forms?

mlzummo:
if there is two forms its submits them in the order they are handled by the marshaller (in theory) ... because the calls done by the marshaller something like this would have to be handled there? there has to be a check that the data is indeed loaded before the submit is executed

mlzummo:
there are some other issues though with doing this that i opened up an issue for

mlzummo:
@GordonSmith Please Re-Review

src/marshaller/HipieDDL.js
```diff
@@ -480,6 +480,7 @@
         this.dashboard = dashboard;
         this.id = visualization.id;
         this.label = visualization.label;
+        this.default = visualization.default;
```
GordonSmith: Change to:  visualization.default || "";

src/marshaller/HipieDDL.js
```diff
@@ -566,6 +567,7 @@
                                 .name(field.id)
                                 .label((field.properties ? field.properties.label : null) || field.label)
                                 .type("textbox")
+                                .value(field.properties.default ? field.properties.default : "")
```
GordonSmith: Change to ```.value(field.properties.default)```

src/form/Input.js
```diff
@@ -84,8 +84,7 @@
             w.blur(w);
         });
         var context = this;
-        this._inputElement.on("change", function (w) {
-
+        this._inputElement.on("keyup", function (w) {
```
GordonSmith: I am nervous that this won't work for some of the esoteric inputs (like html-color)

src/form/Input.js
```diff
@@ -98,6 +97,19 @@
...
+            case "checkbox":
+                this.value(this._inputElement.property("checked"));
+                break;
+            default:
+                this.value(this._inputElement.property("value"));
+                break;
+        }
+        this.change(this);
+        return this.value();
+    };
```
GordonSmith: Is this needed - it looks like it is doing more than just getValue?

src/form/Form.js
```diff
@@ -240,6 +242,19 @@
         ;
     };
 
+    Form.prototype.inputsRendered = function() {
```
GordonSmith: Wouldn't ```this._renderCount > 0``` give you the same thing?

GordonSmith:
Is this for master or candidate-1.4.2?  (as it stands I will not pull into 1.4.2 as there are too many new features)

mlzummo:
@GordonSmith Removed all new features except for base default code ... Please Re-Review

GordonSmith:
Looks good - can you reissue against 1.4.2 so?

mlzummo:
closing and reissuing against 1.4.2

Issue #584, opened 2015-07-17
Create a Tooltip Widget for the D3 Charts 


dtsnell4:
Assigned to Dan Snell

GordonSmith:
@dtsnell4 I am still planning on tweaking your PR for this and pulling into 1.6.0

PR #592, opened 2015-07-21
D3 Charts Tooltip Widget (Issue #584)

dtsnell4:
@GordonSmith @mlzummo @jbrundage  Please review

### DIFF
--- a/src/chart/Bubble.js
+++ b/src/chart/Bubble.js
@@ -1,11 +1,11 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "../common/SVGWidget", "../api/I2DChart", "../common/Text", "../common/FAChar", "css!./Bubble"], factory);
+        define(["d3", "../common/SVGWidget", "../api/I2DChart", "../common/Text", "../common/FAChar", "../layout/Tooltip", "css!./Bubble"], factory);
     } else {
-        root.chart_Bubble = factory(root.d3, root.common_SVGWidget, root.api_I2DChart, root.common_Text, root.common_FAChar);
+        root.chart_Bubble = factory(root.d3, root.common_SVGWidget, root.api_I2DChart, root.common_Text, root.common_FAChar, root.layout_Tooltip);
     }
-}(this, function (d3, SVGWidget, I2DChart, Text, FAChar) {
+}(this, function (d3, SVGWidget, I2DChart, Text, FAChar, Tooltip) {
     function Bubble(target) {
         SVGWidget.call(this);
         I2DChart.call(this);
@@ -24,6 +24,7 @@
     Bubble.prototype.implements(I2DChart.prototype);
 
     Bubble.prototype.publish("paletteID", "default", "set", "Palette ID", Bubble.prototype._palette.switch(),{tags:['Basic','Shared']});
+    Bubble.prototype.publish("useTooltip", true, "boolean", "Use the Tooltip Widget for the chart", null,{tags:[]});
 
     Bubble.prototype.size = function (_) {
         var retVal = SVGWidget.prototype.size.apply(this, arguments);
@@ -34,10 +35,24 @@
         }
         return retVal;
     };
-
+    
     Bubble.prototype.update = function (domNode, element) {
-        var context = this;
+        if (this.useTooltip()) {
+            if (!this._parentOverlay && this._parentWidget._parentOverlay) {
+                this._parentOverlay = this.locateOverlayNode();  
+            }
+            this._parentOverlay.selectAll(".layout_Tooltip").each(function(d) {
+                d.node().parentNode.remove();
+            });
+            
+            this._tooltip = new Tooltip()
+                .registerWidget(this)
+                .target(this._parentOverlay.node())
+                .render()
+            ;
+        }
 
+        var context = this;
         this._palette = this._palette.switch(this.paletteID());
         var node = element.selectAll(".node")
             .data(this._data.length ? this.d3Pack.nodes({ children: this.cloneData() }).filter(function (d) { return !d.children; }) : [], function (d) { return d[0]; })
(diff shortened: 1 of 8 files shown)


dtsnell4:
Notes:

I used the Table widget and Popup widget to make the Tooltip.  Obviously the Popup widget widget hasn't been merged yet, so it is included in this PR exactly as it is in its own PR.

I understand events are still not finalized, so for the sake of this widget, I simply handled the mouse events the same way as the click events are already being handled.

src/chart/Bubble.js
```diff
@@ -35,6 +36,23 @@
...
+            });
+            
+            this._tooltip = new Tooltip()
+                .registerWidget(this)
+                .target(this._parentOverlay.node())
+                .render()
+            ;
+        }
+    };
+
```
dtsnell4: This code is obviously to instantiate the Tooltip widget.  Because it is the same in every chart, perhaps we should create a chart/Common.js file?

src/chart/Pie.js
```diff
@@ -75,14 +92,25 @@
         arc.enter().append("g")
             .attr("class", "arc")
             .attr("opacity", 0)
+            //Events
             .on("click", function (d) {
-                context.click(context.rowToObj(d.data), context._columns[1]);
+                console.log(d);
```
dtsnell4: will be removed

demos/dermatology.html
```diff
@@ -94,6 +94,8 @@
                             <option value="src/layout/Cell">Cell Layout</option>
                             <option value="src/layout/Grid">Grid Layout</option>
                             <option value="src/layout/Border">Border Layout</option>
+                            <option value="src/layout/Popup">Popup Layout</option>
+                            <option value="src/layout/Tooltip">Tooltip Layout</option>
```
dtsnell4: This will be removed

src/chart/Bubble.js
mlzummo: Agreed, and on that note i think a lot of the charts that inherit scatter should actually have that scatter code placed in a "common" file instead of inheriting directly from scatter from an organizational POV.

src/chart/Bubble.js
```diff
@@ -48,13 +66,22 @@
             .attr("class", "node")
             .attr("opacity", 0)
             .on("click", function (d) {
-                context.click(context.rowToObj(d), context._columns[1]);
+                    console.log("data", context._data);
```
mlzummo: @dtsnell4 @GordonSmith 
indent, + should this inherit IXCHart such that you can do .on("click", this.click); ?

src/layout/Popup.js
```diff
@@ -0,0 +1,138 @@
...
+
+    Popup.prototype.widgetSize = function (widgetDiv) {
+        var width = this.clientWidth() - this.calcFrameWidth(widgetDiv);
+        var height = this.clientHeight() - this.calcFrameHeight(widgetDiv);
+ 
+        return { width: width, height: height };
+    };
+    
+    Popup.prototype.updateState = function (visible) {
+            this.popupState(visible).render();
```
mlzummo: fix indent

src/layout/Popup.js
```diff
@@ -0,0 +1,138 @@
...
+    };
+    
+    Popup.prototype.enter = function (domNode, element) {
+        HTMLWidget.prototype.enter.apply(this, arguments);
+
+        if (this._testButton) {
```
mlzummo: can this not be handled within testData or structured such that you have a hardcoding for test related data in .enter?

src/layout/Popup.js
```diff
@@ -0,0 +1,138 @@
...
+
+    Popup.prototype.exit = function (domNode, element) {
+        if (this.widget()) {
+            this.widget(null);
+            this.render();
+        }
+        HTMLWidget.prototype.exit.apply(this, arguments);
+    };
+
+    Popup.prototype.render = function (callback) {
```
mlzummo: is there a diff way to handle the callback, as i have a PR for a new kind of callback method and this would interfere with it potentially https://github.com/hpcc-systems/Visualization/pull/602  ... its still very much debatable if it will get used though

src/layout/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+    Tooltip.prototype.hideTooltip = function(data) {
+        this.tooltipState(false).render();
+    };
+    
+    Tooltip.prototype.registerWidget = function(widget) {
+        var context = this;
+        widget
+            .on("mouseover", function(d, i) {
+                var data = [],
+                    columns = [];
```
mlzummo: indent

src/chart/Bubble.js
```diff
@@ -35,6 +36,23 @@
         return retVal;
     };
 
+    Bubble.prototype.enter = function () {
+        if (this.useTooltip()) {
+            if (!this._parentOverlay && this._parentWidget._parentOverlay) {
+                this._parentOverlay = this._parentWidget._parentOverlay;  //This is because charts do not get an overlay div when they are inside the MultiChart
+            }
```
GordonSmith: This will fail if the multichart is inside a SVG widget (like Surface) - the "correct" way to get the overlayNode is:
```
var overlayNode = this.locateOverlayNode();
```

src/chart/Bubble.js
```diff
@@ -35,6 +36,23 @@
         return retVal;
     };
 
+    Bubble.prototype.enter = function () {
+        if (this.useTooltip()) {
```
GordonSmith: I don't think we want this test here - otherwise you can't disable/enable it after it has rendered once.

src/chart/Bubble.js
GordonSmith: Agree 100% - I am wondering if it might even be pushed all the way into "widget"?  We could then (re)use it on other charts also (like c3)?  It is my plan that we will use the new Legend with all the charts and the same could be true for this tooltip (as long as the 3rd party ones fire the needed events)?

src/chart/Bubble.js
GordonSmith: Or is it a "ITooltip" - then any chart which want to implement an ITooltip would be forced to honor the contract of "mouseover/out/move"?

src/layout/Tooltip.js
```diff
@@ -0,0 +1,141 @@
...
+            .on("mousemove", function() {
+                var table = d3.select(context._table.node()).select("table"),
+                    height = table.node().offsetHeight,
+                    width = table.node().offsetWidth,
+                    rect;
+                
+                if (widget._parentRelativeDiv) { //This is because charts do not get assigned this when they are inside the MultiChart
+                    rect = widget._parentRelativeDiv.node().getBoundingClientRect();
+                } else {
+                    rect = widget._parentWidget._parentRelativeDiv.node().getBoundingClientRect();
```
GordonSmith: Going up 1 level will not always work as you can have SVG widgets inside SVG widgets etc.  I think you are looking for the rect of the widget in the same coordinate space as the tooltip? (so you prevent it from being clipped on the right/bottom)?

dtsnell4: @GordonSmith yes that is correct.  What would be the appropriate method for doing that?

src/chart/Bubble.js
dtsnell4: @GordonSmith @mlzummo @jbrundage So what is the consensus, to move this into Widget.js?  And if so, where would be the appropriate place to put it?

GordonSmith:
Closing in favor of:  https://github.com/hpcc-systems/Visualization/pull/726

Issue #607, opened 2015-07-28
layout/Grid Design Mode grid-drop-target only displays in nested Grids


jbrundage:
The CSS selectors for grid-drop-target are incorrect

GordonSmith:
Fixed by GH-640

PR #608, opened 2015-07-28
GH-607 Grid layout Design Mode now displays grid-drop-target

jbrundage:
Fixes GH-607

@GordonSmith Please Review

### DIFF
--- a/apps/SampleSite/grid.html
+++ b/apps/SampleSite/grid.html
@@ -1,0 +1,54 @@
+﻿<!doctype html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <title>HPCC Systems - Dermatology Test</title>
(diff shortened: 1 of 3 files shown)


GordonSmith:
@jbrundage I don't see how removing .layout_Grid changes anything as a cell drop target is always within a .layout_Grid div?

jbrundage:
The .grid-drop-target is a temporary fixed div that exists outside of the .layout_Grid wrapper. It's created when dragging starts and its removed when dragging ends.

It seemed to be working just fine on dermatology.html, but there are nested Grids on that page.

jbrundage:
Actually... maybe the better fix would be to make the .grid-drop-target a child element of .layout_Grid.  There's no reason it shouldn't be...

PR #618, opened 2015-08-03
Pal opts

mlzummo:
@GordonSmith @jbrundage @dtsnell4 

Draft of new palette options

### DIFF
--- a/src/common/Palette.js
+++ b/src/common/Palette.js
@@ -88,6 +88,12 @@
             return ordinalCache[newID];
 
         };
+        ordinal.cloneNotExists = function (newID) {
+            if (ordinalCache[newID]) {
+                return ordinalCache[newID];
+            }
+            return this.clone(newID);
+        };
         ordinal.switch = function (_id, _colors) {
             if (id === _id) {
                 return this;
@@ -180,6 +186,12 @@
         rainbow.clone = function (newID) {
             rainbowCache[newID] = palette_rainbow(newID, this.color());
             return rainbowCache[newID];
+        };
+        rainbow.cloneNotExists = function (newID) {
+            if (rainbowCache[newID]) {
+                return rainbowCache[newID];
+            }
+            return this.clone(newID);
         };
         rainbow.switch = function (_id, _colors) {
             if (id === _id) {
--- a/src/tree/Dendrogram.js
+++ b/src/tree/Dendrogram.js
@@ -16,11 +16,12 @@
     Dendrogram.prototype = Object.create(SVGWidget.prototype);
     Dendrogram.prototype._class += " tree_Dendrogram";
     Dendrogram.prototype.implements(ITree.prototype);
-    
+
     Dendrogram.prototype.publish("paletteID", "default", "set", "Palette ID", Dendrogram.prototype._palette.switch(),{tags:['Basic','Shared']});
     Dendrogram.prototype.publish("textOffset", 8, "number", "Text offset from circle",null,{tags:['Private']});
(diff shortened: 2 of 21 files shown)


src/tree/Dendrogram.js
```diff
@@ -16,11 +16,13 @@
...
-    
+
     Dendrogram.prototype.publish("paletteID", "default", "set", "Palette ID", Dendrogram.prototype._palette.switch(),{tags:['Basic','Shared']});
     Dendrogram.prototype.publish("textOffset", 8, "number", "Text offset from circle",null,{tags:['Private']});
-    Dendrogram.prototype.publish("orientation", "horizontal", "set", "Orientation", ["horizontal","vertical"],{tags:['Private']}); 
-     
+    Dendrogram.prototype.publish("orientation", "horizontal", "set", "Orientation", ["horizontal","vertical"],{tags:['Private']});
+    Dendrogram.prototype.publish("paletteUseIdx", false, "boolean", "Change color option to use data index",null,{tags:['Basic','Shared']});
+
+
```
mlzummo: indent

mlzummo:
@GordonSmith Please Review ... can we continue with this approach?

mlzummo:
@GordonSmith Can you review this ... we would like to make it simpler and more intuitive to deal with the palette behavior

GordonSmith:
As per the phone conversation here is my issue with this approach:
* User gets all data back from server and displays it (say 10 bars on a bar chart).
* User now applies a filter and displays it (say 3 of the 10 bars)
* In the second case the bars will have changed color in relation to there x-axis label.

The solution discussed was to ensure that (optionally) each widget instance got its own "cloned" version of the palette (which I also suspect could have a cleaner solution).

mlzummo:
what would be the behavior if it did exist? ... is there anywhere we can do the palette cloning once instead of now having to do it in everysingle widget file?

also we all would like to see the option of being able to pass the index into color generation (default behavior is unchanged) would like to keep this!

GordonSmith:
If it did exist it would just return the current one matching.

```this._palette = this._palette.switch(this.paletteID());```  Already exists in every "common" file plus a few more., your just adding the "if" beside it.

**Comment by GordonSmith** [comment]
Final thoughts - if this is for the marshallers then I would suggest this:
* for each datasource create a cloned instance of the palette from the default.
* for each widget that is created, make it use the paletteID of its datasource

Then the solution is 100% within the base marshaller and is logical in so far as you probably want to share palettes between the visualizations that share the same datasource!

**Comment by mlzummo** [comment]
use of cloned palette does not fix the issue still fixed by passing an index val

**Comment by mlzummo** [comment]
not sure why as d3_scale_ordinal does return starting at 0

**Comment by GordonSmith** [comment]
Using a cloned palette will work exactly the same as using the idx - until the columns change, then it will maintain the colors in sync with the column names (specifically for the Multi Series Charts).

**Comment by mlzummo** [comment]
nvm all is working

File: src/common/Palette.js [hunk 2]
```diff
@@ -65,6 +65,9 @@
                 }
             }
             if (!scale) {
+                console.log('here');
+                console.log(d3.scale.category20());
+
```
GordonSmith: Remove logging

GordonSmith:
You could add a unit-test which checks that every widget which has a paletteID property also has a useClonedPalette property (nice test for future widgets...)

mlzummo:
fixes GH-676

GordonSmith:
FYI The Fixes only works in the commit...

PR #627, opened 2015-08-04
Adds Layered layout

jbrundage:
Gordon, is this what you had in mind for the Layered layout?

It works very similarly to src/layout/Grid.js except the **setContent** method takes **zIndex** as an argument:
Layered.prototype.setContent = function (row, col, **zIndex**, widget, title, rowSpan, colSpan)

@GordonSmith @dtsnell4 @mlzummo Please review

https://rawgit.com/jbrundage/Visualization/Adds_Layered_Layout/demos/dermatology.html?src/layout/Layered

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -94,6 +94,7 @@
                             <option value="src/layout/Cell">Cell Layout</option>
                             <option value="src/layout/Grid">Grid Layout</option>
                             <option value="src/layout/Border">Border Layout</option>
+                            <option value="src/layout/Layered">Layered Layout</option>
                             <option value="src/amchart/Area">AmChart Area</option>
                             <option value="src/amchart/Bar">AmChart Bar</option>
                             <option value="src/amchart/Bubble">AmChart Bubble</option>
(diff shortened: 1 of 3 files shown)


dtsnell4:
J, I have a question...  based on Gordon's description of the use case he was given, I can imagine the user-dev might want to place the widgets in very precise pixel locations, rather than in an ordered grid.  Using the standard grid system of placement, is this possible?

jbrundage:
You bring up a good point. The Grid's cell density could be set to 200 or 300 improving the precision of the Surface placement, but high cell densities are currently inefficient because the "designMode" creates a div to visually represent each Grid cell.

If we remove the drop-cell DIVs in layout/Layered.js designMode then we should be able to set the grid row count and column count as high as we'd like.

GordonSmith:
Yes you could have a layered layout based off of the grid layout (why didn't you derive from it, rather than duplicate it?), but I suspect that would be a overly complicated solution.

jbrundage:
I duplicated it because I wasn't initially sure which pieces might need to change. 

I could refactor this Layered version of Grid to derive from Grid rather than duplicate it... Deriving from Grid for this layout would allow us to re-purpose the "designMode" code.

GordonSmith:
Closing in favor of:  https://github.com/hpcc-systems/Visualization/pull/647

PR #639, opened 2015-08-05
Adds debug option to require config for global obj/array of widgets

mlzummo:
@GordonSmith @dtsnell4 @jbrundage 

Please Review

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -134,13 +134,17 @@
         if (def) {
             def.split("&").forEach(function (param, idx) {
                 var paramParts = param.split("=");
-                if (widgetPath === "" && paramParts.length === 1) {
+                if (paramParts[0] === "hpcc_debug") {
+                    window.__hpcc_debug = paramParts[1];
+                } else if (widgetPath === "" && paramParts.length === 1) {
                     widgetPath = decodeURIComponent(paramParts[0]);
                 } else {
                     params[decodeURIComponent(paramParts[0])] = decodeURIComponent(paramParts[1]);
                 }
             });
-            document.getElementById("widgetSelect").value = widgetPath;
+            if (widgetPath) {
+                document.getElementById("widgetSelect").value = widgetPath;
+            }
         }
 
         var body = document.getElementById("body");
--- a/src/common/Widget.js
+++ b/src/common/Widget.js
@@ -43,6 +43,13 @@
         this._watchArr = [];
 
         this._renderCount = 0;
+
+        if (window.__hpcc_debug) { // jshint ignore:line
+            if (window.g_all === undefined) {
+                window.g_all = {};
+            }
+            window.g_all[this._id] = this;
+        }
     }
     Widget.prototype._class = " common_Widget";
 


src/config.js
```diff
@@ -1,3 +1,4 @@
+require.debug = true;
```
GordonSmith: Move it inside the config and prefix with __hpcc_

dtsnell4:
Fine with me

GordonSmith:
Looks fine - but should default to false.

jbrundage:
In the future when we're debugging something in a production application, how would we change this to true via the console?

mlzummo:
> requirejs.s.contexts._.config.__hpcc_debug = true;

(requirejs) should be registered in the global name space (but if we are not using requirejs) we should prolly have an alternative method

mlzummo:
@GordonSmith Please Re-Review

demos/dermatology.html
```diff
@@ -134,13 +134,16 @@
         if (def) {
             def.split("&").forEach(function (param, idx) {
                 var paramParts = param.split("=");
+                if (paramParts[0] === "hpcc_debug") { return; }
```
GordonSmith: Are you forcing this to be the first param?  I would have though it could be a param at any position and easiest to have it as the last (from a typing it in point of view...

src/common/Widget.js
```diff
@@ -43,6 +43,25 @@
         this._watchArr = [];
 
         this._renderCount = 0;
+
+        this._hpcc_debug = false;
+        var context = this;
+        var def = window.location.search.split("?")[1];
```
GordonSmith: Why do you need to do this for _every_ widget creation?  And its too heavy IMO.

src/common/Widget.js
```diff
@@ -43,6 +43,25 @@
...
+        if (def) {
+            def.split("&").forEach(function (param, idx) {
+                var paramParts = param.split("=");
+                if (decodeURIComponent(paramParts[0]) === 'hpcc_debug' && (decodeURIComponent(paramParts[1]) === "true" || decodeURIComponent(paramParts[1]) === "1")) {
+                    context._hpcc_debug = true;
+                }
+            });
+        }
+
+        if (requirejs.s.contexts._.config.__hpcc_debug || this._hpcc_debug || window._hpcc_debug) { // jshint ignore:line
```
GordonSmith: Just make this ```if (window.__hpcc_debug)``` then the developer has many different ways to sat it if needs be.

demos/dermatology.html
GordonSmith: Also it was my understanding that this wasn't really useful in the dermatology, but more for your app?

GordonSmith:
Having thought about this overnight, from the "src" folder point of view you should only ever have:
```
if (window.__hpcc_debug) {...}
```
Then the developer can simply add ```<script>window.__hpcc_debug=true;</script>``` in the HTML as they need it (as long as they don't commit it).  You can even include that in one of the commented section in dermatology now?

If you want to make dermatology (or any test page) parse the URL and set this global value is would be simply a luxury.

mlzummo:
@GordonSmith the way i have it now allows for easier access to setting debug from production environments where at first we might not have access to changing the underlying code.

src/common/Widget.js
mlzummo: we can push it off into a global and check for it ... but splitting a small string shouldn't be heavy,

GordonSmith:
Ok sure - but my above comment still stands, the only test you should need in the "src" folder is this: 
```
if (window.__hpcc_debug) {...}
```
If you want to set it to true in the config.js file that is fine with me, but you won't be parsing the URL in the Widget constructor for obvious reasons.

jbrundage:
@mlzummo window.__hpcc_debug is a better approach in my opinion. Adding a breakpoint before the check and enabling debug via console is easy enough.

GordonSmith:
I also assume that the production servers will be using the minimized code?

jbrundage:
Yes, I'm just interested in being able to have console access to the instances of the widgets (to help isolate a bug if it isn't easily reproduced)

mlzummo:
@GordonSmith Please Re-Review

GordonSmith:
"src" looks good.  Dermatology, the following is not working:

mlzummo:
@GordonSmith thats becuase your linking master and this hasent been pulled yet.

try this: http://rawgit.com/mlzummo/Visualization/WIDGET_CONSOLE_DEBUG/demos/dermatology.html?src/chart/Pie&outerText=true&hpcc_debug=1

mlzummo:
still not working actually

mlzummo:
@GordonSmith Fixed

demos/dermatology.html
```diff
@@ -134,13 +134,17 @@
         if (def) {
             def.split("&").forEach(function (param, idx) {
                 var paramParts = param.split("=");
+                console.log(paramParts[0]);
```
GordonSmith: remove console.log

demos/dermatology.html
```diff
@@ -134,13 +134,17 @@
         if (def) {
             def.split("&").forEach(function (param, idx) {
                 var paramParts = param.split("=");
+                console.log(paramParts[0]);
+                if (paramParts[0] === "hpcc_debug") { window.__hpcc_debug = true; return; }
                 if (widgetPath === "" && paramParts.length === 1) {
                     widgetPath = decodeURIComponent(paramParts[0]);
                 } else {
                     params[decodeURIComponent(paramParts[0])] = decodeURIComponent(paramParts[1]);
                 }
```
GordonSmith: Change to:
```
if (paramParts[0] === "hpcc_debug") { 
    window.__hpcc_debug = paramParts[1];  // Then you can also have options like "FULL", "LITE" if you wanted to
} else if (widgetPath === "" && paramParts.length === 1) {
    widgetPath = decodeURIComponent(paramParts[0]);
} else {
    params[decodeURIComponent(paramParts[0])] = decodeURIComponent(paramParts[1]);
}
```

mlzummo:
@GordonSmith Fixed per suggestions

GordonSmith:
Looks good.

PR #651, opened 2015-08-06
Adds Utility Obj

mlzummo:
@GordonSmith @jbrundage @dtsnell4 

Please Review

### DIFF
--- a/src/common/Utility.js
+++ b/src/common/Utility.js
@@ -1,0 +1,58 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define([], factory);
+    } else {
+        root.common_Utility = factory();
+    }
+}(this, function () {
+
+    function _naturalSort(a, b, order, idx, sortCaseSensitive) {
+        var re = /(^([+\-]?(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)?)?$|^0x[0-9a-f]+$|\d+)/gi,
+            sre = /(^[ ]*|[ ]*$)/g,
+            dre = /(^([\w ]+,?[\w ]+)?[\w ]+,?[\w ]+\d+:\d+(:\d+)?[\w ]?|^\d{1,4}[\/\-]\d{1,4}[\/\-]\d{1,4}|^\w+, \w+ \d+, \d{4})/,
+            hre = /^0x[0-9a-f]+$/i,
+            ore = /^0/,
+            i = function(s) { return !sortCaseSensitive && ("" + s).toLowerCase() || "" + s; },
+            // convert all to strings strip whitespace
+            x = i(idx ? a[idx] : a).replace(sre, "") || "",
+            y = i(idx ? b[idx] : b).replace(sre, "") || "",
+            // chunk/tokenize
+            xN = x.replace(re, "\0$1\0").replace(/\0$/,"").replace(/^\0/,"").split("\0"),
+            yN = y.replace(re, "\0$1\0").replace(/\0$/,"").replace(/^\0/,"").split("\0"),
+            // numeric, hex or date detection
+            xD = parseInt(x.match(hre), 16) || (xN.length !== 1 && x.match(dre) && Date.parse(x)),
+            yD = parseInt(y.match(hre), 16) || xD && y.match(dre) && Date.parse(y) || null,
+            oFxNcL, oFyNcL;
+        // first try and sort Hex codes or Dates
+        if (yD) {
+            if ( xD < yD ) { return order === "ascending" ? -1 : 1; }
+            else if ( xD > yD ) { return order === "ascending" ? 1 : -1; }
+        }
+        // natural sorting through split numeric strings and default strings
+        for(var cLoc=0, numS=Math.max(xN.length, yN.length); cLoc < numS; cLoc++) {
+            // find floats not starting with "0", string or 0 if not defined (Clint Priest)
+            oFxNcL = !(xN[cLoc] || "").match(ore) && parseFloat(xN[cLoc]) || xN[cLoc] || 0;
+            oFyNcL = !(yN[cLoc] || "").match(ore) && parseFloat(yN[cLoc]) || yN[cLoc] || 0;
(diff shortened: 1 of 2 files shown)


GordonSmith:
Would common be a better location for this?

jbrundage:
I also would have assumed it should go in "other"

GordonSmith:
Currently "other" depends on "common" - If we ever use a Util function in common, then that dependency will be circular (which is technically ok - but flags to me that Util may belong in common).

mlzummo:
@GordonSmith Moved to common/

src/common/Utility.js
```diff
@@ -0,0 +1,69 @@
...
+                oFxNcL += '';
+                oFyNcL += '';
+            }
+            if (oFxNcL < oFyNcL) { return order === "ascending" ? -1 : 1; }
+            if (oFxNcL > oFyNcL) { return order === "ascending" ? 1 : -1; }
+        }
+        return 0;
+    };
+
+    return Utility;
```
GordonSmith: There should be no "Utility" function and then your content will simply be:
```
function _naturalSort(...) {
  //  This function is not accessable outside of this file.
}
return {
    naturalSort: function(...) {
        ...
    }
};
```

src/common/Utility.js
```diff
@@ -0,0 +1,69 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define([], factory);
+    } else {
+        root.common_Utility = factory();
+    }
+}(this, function () {
+    function Utility() {
```
GordonSmith: Remove

src/common/Utility.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    /**
+     * USAGE: this._data = Utility.sort(this._data, "naturalsort", context._currentSortOrder === 1 ? "ascending" : "descending", idx, true);
+     */
+    Utility.sort = function (data, type, order, idx, sortCaseSensitive, options) {
```
GordonSmith: rename to naturalSort

src/common/Utility.js
```diff
@@ -0,0 +1,69 @@
...
+        var context = this;
+        if (!type) { type = "natualsort"; }
+        return data.sort(function(a,b) {
+            switch (type) {
```
GordonSmith: remove switch statement that doesn't do anything

GordonSmith:
Looking at the code I would suggest refactoring it and expose it as a "comparitor" like d3.ascending - then you can re-use it to sort and Array of Strings as well as an Array of Objects,

mlzummo:
how will I be able to handle case sensitivity and later adding the option of how to sort numbers (beginning or end)

GordonSmith:
Same as you are now?  I am not suggesting deleting the code, just refactoring.  As it stands I can't use it to sort the following:  ```["A100", "A10", "A1"]```

src/common/Utility.js
```diff
@@ -0,0 +1,65 @@
...
+    if (typeof define === "function" && define.amd) {
+        define([], factory);
+    } else {
+        root.common_Utility = factory();
+    }
+}(this, function () {
+
+    /**
+     *
+     */
```
GordonSmith: Remove empty comments

src/common/Utility.js
```diff
@@ -0,0 +1,65 @@
...
+            }
+            if (oFxNcL < oFyNcL) { return order === "ascending" ? -1 : 1; }
+            if (oFxNcL > oFyNcL) { return order === "ascending" ? 1 : -1; }
+        }
+        return 0;
+    }
+
+    return {
+        /*
+         * USAGE: this._data = Utility.sort(this._data, context._currentSortOrder === 1 ? "ascending" : "descending", idx, true);
```
GordonSmith: I am not sure how I feel about this comment - it does have some useful information (ascending and descending are two options for "order") but the user still needs to dig into the code to workout how to use the function (which is fine IMO).

src/common/Utility.js
```diff
@@ -0,0 +1,65 @@
...
+         */
+        naturalSort: function(data, order, idx, sortCaseSensitive, options) {
```
GordonSmith: options not used.

GordonSmith:
Some thoughts on inline JSDoc comments:
* They need to add value.
* They need to be 100% accurate and 100% complete  - the developer needs to be confident in this otherwise they will still just dig into the sources to see what it does - at which point the comment becomes noise.
* In my experience comments over time get out of sync with the sources.
* A wrong comment can be just as damaging as a bug (more so in some respects) and are impossible to test.

GordonSmith:
Can you add some unit tests for this (the framework for this just got added over the weekend, so ping me I can show you where/how).

Further:  A set of unitTests also doubles as "documentation" as it will clearly show working examples and is guaranteed to be in sync with the sources...

test/runner.html
```diff
@@ -91,6 +91,9 @@
         require(["./widgets"], function () {
             mocha.run();
         });
+        require(["./utility"], function () {
+            mocha.run();
+        });
     </script>
```
GordonSmith: You should combine those to:  
```
require(["./widgets", "./utility"], function () {
           mocha.run();
        });
```

test/utility.js
```diff
@@ -0,0 +1,28 @@
...
+
+            var sortedListD = Utility.naturalSort(list,"descending",null,true);
+            var expectedListD = [456, "222", 123, "111", "c", "b", "a", "CD456", "AB123", "AB", "AA"];
+
+            assert.deepEqual(sortedListA,expectedListA);
+            assert.deepEqual(sortedListB,expectedListB);
+            assert.deepEqual(sortedListC,expectedListC);
+            assert.deepEqual(sortedListD,expectedListD);
+
+            done();
```
GordonSmith: You haven't tested the 3rd param?

test/runner.html
mlzummo: is there a quick way we can turn on an off a group then? the way it is now i can quickly comment out the moch.run() for any particular one

mlzummo: http://localhost:8080/test/runner.html?grep=utility    ?

GordonSmith:
@mlzummo Can you rebase.

PR #668, opened 2015-08-09
Adds the Gantt Chart

anmoljagetia:


GordonSmith:
Closing as the Gantt charts is now the same as a Column with the following settings:
* orientation: vertical
* stacked: true
* data with ranges.

PR #672, opened 2015-08-09
Loads widgets without calling testData method

anmoljagetia:
This unit test loads the widgets into a div element on the test page, without calling the testData method on it.


GordonSmith:
Closing as a variant has already been added to master.

PR #675, opened 2015-08-10
Adds code validation to the Framework

anmoljagetia:
Presently, there are no rules in the .jscsrc file, but we can add more, and ensure coding guidelines. In a seperate PR, I'll be working to fix the quotemark rule.


gulpfile.js
```diff
@@ -86,6 +87,15 @@ gulp.task('build-css', css.bind(null, false));
 
 gulp.task('optimize-css', css.bind(null, true));
 
+gulp.task('jscs', function() {
+    gutil.log("JSCS the files...." + '\n');
+    var lintFilter = filter(["**", "!config.js", "!map/us-counties.js", "!map/us-states.js", "!map/countries.js"]);
```
GordonSmith: Move lintFilter out of task so it can be reused in lint and jscs

GordonSmith:
jscs should also be called from "test"

gulpfile.js
anmoljagetia: I am assuming there will be different files in them at a later point. (I am unsure of it, right now, so for now, should I move them out?)

GordonSmith: I am 99% sure that the same list will be used for both.

GordonSmith:
Can you roll up to one commit

Issue #582, opened 2015-07-17
HPCC Table rendering problem when changing columns data


dunguyen-kms:
The hpcc table has problem with headers when I change columns data

- Columns is [column1, column2, column3]. You will see
![hpcc-table1](https://cloud.githubusercontent.com/assets/12228166/8743097/466de968-2c96-11e5-9a3c-f57ecd4381bb.jpg)

- Columns is [cloumn1, column4, column3, column5]. You expect
![hpcc-table2-expected](https://cloud.githubusercontent.com/assets/12228166/8743122/9cfaf104-2c96-11e5-87a7-41c38c3be7c3.jpg)
But you will see
![hpcc-table2](https://cloud.githubusercontent.com/assets/12228166/8743271/75217dc8-2c97-11e5-975d-d35861e6e0ef.jpg)

The column order has problem when you re-render table again and new columns have some old existing column.

mlzummo:
tableWidget
    .columns([])
    .data([])
    .render()
    .columns(newCols)
    .data(newData)
    .render()
;

is the current work around for that

dunguyen-kms:
@mlzummo - b/c of performance, I delay the render() function by using setTimeOut method when the data or columns change. So this workaround does NOT work in my case

mlzummo:
@dunguyen-kms can you provide me a link to where i can run/view the code your using?

dunguyen-kms:
@mlzummo I actually want to show you those lines of code. But this is an internal project so I can't.
I wrap the HPCC table to an Angular directive. And I also use setTimeOut to delay rendering whenever there's something change in columns and data. That's the reason why I can't use the workaround in this case

PR #683, opened 2015-08-11
GH-582 Table columns render in wrong sequence

GordonSmith:
Adding/removing or editing the columns will cause them to lose their correct order.

Fixes GH-582

### DIFF
--- a/src/other/Table.js
+++ b/src/other/Table.js
@@ -99,7 +99,7 @@
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
 
-        var th = this.thead.selectAll("th").data(this._columns, function (d) { return d;});
+        var th = this.thead.selectAll("th").data(this._columns);
         th
             .enter()
             .append("th")


GordonSmith:
@buunguyen Please revie:  http://rawgit.com/GordonSmith/Visualization/GH-582/demos/dermatology.html?src/other/Table
(You can turn columns on and then edit it).

buunguyen:
LGTM.

@dunguyen-kms this should fix the problem you're having.

mlzummo:
im curious as to "why" this fixes the issue?

buunguyen:
My review is probably lame (I'm not familiar with d3). Just run a [quick test](http://jsfiddle.net/y17p8dbn/), it appears that this fix, while solving the reordering, has an issue: existing data won't get replaced (i.e. I expect to see 0, 1, 3 in my test but it still shows 0, 1, 2). Hope @GordonSmith can shed some light.

mlzummo:
that sounds above right, i recall it specially being changed to what it is now becuase of that

GordonSmith:
@buunguyen Your quickTest is not using the "3 stages" of the d3 selectAll.data pattern.  See https://jsfiddle.net/y17p8dbn/1/ for the working version.

GordonSmith:
@mlzummo Can you re-issue that missing ";" fix against candidate-1.4.4?

buunguyen:
@GordonSmith your example shows 0, 1, 2 instead of 0, 1, 5, 6. Is it the expected behavior?

GordonSmith:
remove the "s" from the http and try again<g>.

GordonSmith:
Final review still needed....

buunguyen:
To be honest, I'm not clear why it works the way it does. I remove the `exit()` from your example (i.e. break the 3 stage pattern) and it appears to still work http://jsfiddle.net/u9m0fsrn/. That said, this fix seems to achieve the effect and that's probably good enough.

Guess I should pick up some more d3 knowledge :).

GordonSmith:
In that example exit is never called beause the number of "p" elements only got longer.
This example would have an exit:  http://jsfiddle.net/u9m0fsrn/1/

Issue #685, opened 2015-08-11
Theme serials need a seamless integration


jbrundage:
Applying a theme serialization needs to be simple for the developer. They should be able to paste their theme string in one location regardless of what widgets will be used.

PR #686, opened 2015-08-11
Adds global theme string support

jbrundage:
Using a global variable, window.__hpcc_theme, to declare the theme string to be applied to any/all widgets.

Fixes GH-685

@GordonSmith Please review

Is this a preferable structure for the theme serial? If not, what do you suggest?


jbrundage:
@GordonSmith This is ready to be inspected again.

I've added the *Theme dropdown* along with a few themes to dermatology.html.

I've also added the Theme Text checkbox.

https://rawgit.com/jbrundage/Visualization/Theme_Global/demos/dermatology.html?src/layout/Grid

src/other/Persist.js
```diff
@@ -31,7 +31,144 @@
...
+                                    var paramIsIncluded = false;
+                                    var classParts = widget._class.trim().split(" ");
+                                    classParts.forEach(function(c) { 
+                                        if(!paramIsIncluded){
+                                            if(propObj[c] === undefined){
+                                                propObj[c] = {};
+                                            } 
+                                            if (propObj[c][item.id] === undefined) { 
+                                                propObj[c][item.id] = widget[item.id](); 
+                                                paramIsIncluded = true;
```
GordonSmith: You can break out of the loop at this point?

src/other/Persist.js
```diff
@@ -31,7 +31,144 @@
...
+                                            } 
+                                            else if (propObj[c][item.id] === widget[item.id]()) {
+                                                paramIsIncluded = true;
```
GordonSmith: You can break out of the loop at this point?

src/other/Persist.js
```diff
@@ -31,7 +31,144 @@
...
+                                }
+                                break;
+                        }
+                    }
+                });
+                function _isFilterMatch(str,arr){
+                    var ret = false;
+                    arr.forEach(function(substr){
+                        if(str.indexOf(substr) !== -1){
+                            ret = true;
```
GordonSmith: You should break out of the forEach at this point.

src/other/Persist.js
jbrundage: As it turns out... there's no built-in break for Array.forEach. I could turn it into a **for(var i in classParts)** though.

GordonSmith: You should use Array.some I think (or for)

jbrundage: I've switched them to **for** just now. Array.some sounds interesting, I havent used it before.

GordonSmith:
@jbrundage With those "break" changes your theme strings should be more concise, can you update the ones you use in the dermatology page?

jbrundage:
@GordonSmith They should produce the same strings. This boolean **paramIsIncluded** was being used to break without actually breaking. I should remove it now that it's actually using break.

jbrundage:
@GordonSmith I believe this is ready to be pulled

GordonSmith:
Looks good - can you remove the dermotology_1.html file?

jbrundage:
@GordonSmith It's removed

PR #699, opened 2015-08-12
Fixes oversight in NON-AMD version with amcharts img directory resolution

mlzummo:
@GordonSmith Please Review 

it seems that the dist version didn't like it when the property was present

### DIFF
--- a/src/amchart/CommonFunnel.js
+++ b/src/amchart/CommonFunnel.js
@@ -133,9 +133,11 @@
             type: "funnel",
             autoResize: true,
             autoMargins: true,
-            chartScrollbar: {},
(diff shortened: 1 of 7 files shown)


mlzummo:
the previous way seems to "sporatically" have problems (for example i got it to not work once but its now working)

mlzummo:
@GordonSmith I cant get amchart Pie to pass phantomjs mocha, even on preivous branches now that have already passed in the past (locally as well)

GordonSmith:
@mlzummo @buunguyen The tests passed fine - but in the console output you are seeing "console.log / console.eror" messages from the browser itself which is confusing I know, but that is why we have been removing them.  If you open the test page in a browser and look at the console log you will see them there also.  We should just continue to remove them.

Issue #542, opened 2015-07-01
Table "fixed" column titles and row titles.


GordonSmith:
Similar to this:  ![image](https://cloud.githubusercontent.com/assets/966863/8457203/f08c51f0-2007-11e5-85f1-a903f3403a72.png)

PR #702, opened 2015-08-12
GH-542 Fixed table header and first column

dtsnell4:
Adds functionality for a fixed table header and fixed first column for the table widget.

Fixes GH-542

Signed-off-by: Dan Snell <dan.snell@lexisnexis.com>

### DIFF
--- a/src/other/Table.js
+++ b/src/other/Table.js
@@ -17,34 +17,68 @@
         this._selectionPrevClick = null;
     }
     Table.prototype = Object.create(HTMLWidget.prototype);
-    Table.prototype.constructor = Table;
     Table.prototype._class += " other_Table";
+    Table.prototype.constructor = Table;
 
     Table.prototype.testData = function () {
         this
-            .columns(["Lat", "Long", "Pin"])
+            .columns(["Lat", "Long", "Pin", "Forth Column", "Fifth Column", "sixth Column", "Seventh Column", "eighth Column", "Nineth Column", "Tenth Column"])
             .data([
-                [37.665074, -122.384375, "green-dot.png"],
-                [32.690680, -117.178540],
-                [39.709455, -104.969859],
-                [41.244123, -95.961610],
-                [32.688980, -117.192040],
-                [45.786490, -108.526600],
-                [45.796180, -108.535652],
-                [45.774320, -108.494370],
-                [45.777062, -108.549835, "red-dot.png"]
+                [37.665074, -122.384375, "green-dot.png", 4525, 4243545, 24354, 54, 2552345, 2455, 245435],
+                [32.690680, -117.178540, "", 4525, 423545, 24354, 354, 2552345, 2455, 245435],
+                [39.709455, -104.969859, "", 4525, 423545, 24354, 524, 2552345, 2455, 245435],
+                [41.244123, -95.961610, "", 4525, 423545, 24354, 564, 2552345, 2455, 245435],
+                [32.688980, -117.192040, "", 4525, 423545, 24354, 454, 2552345, 2455, 245435],
+                [45.786490, -108.526600, "", 4525, 423545, 24354, 534, 2552345, 2455, 245435],
+                [45.796180, -108.535652, "", 4525, 423545, 24354, 254, 2552345, 2455, 245435],
+                [45.774320, -108.494370, "", 4525, 423545, 24354, 51, 2552345, 2455, 245435],
+                [37.665074, -122.384375, "green-dot.png", 4525, 423545, 24354, 504, 2552345, 2455, 245435],
+                [32.690680, -117.178540, "", 4525, 423545, 24304, 54, 2552345, 2455, 245435],
+                [39.709455, -104.969859, "", 4525, 423545, 249354, 54, 2552345, 2455, 245435],
+                [41.244123, -95.961610, "", 4525, 423545, 247354, 54, 2552345, 2455, 245435],
+                [32.688980, -117.192040, "", 4525, 423545, 243654, 54, 2552345, 2455, 245435],
+                [45.786490, -108.526600, "", 4525, 423545, 245354, 54, 2552345, 2455, 245435],
+                [45.796180, -108.535652, "", 4525, 423545, 243354, 54, 2552345, 2455, 245435],
+                [45.774320, -108.494370, "", 4525, 423545, 243454, 54, 2552345, 2455, 245435],
+                [37.665074, -122.384375, "green-dot.png", 4525, 423545, 243254, 54, 2552345, 2455, 245435],
+                [32.690680, -117.178540, "", 4525, 4243545, 24354, 54, 2552345, 2455, 245435],
+                [39.709455, -104.969859, "", 4525, 4233545, 24354, 54, 2552345, 2455, 245435],
+                [41.244123, -95.961610, "", 4525, 4235145, 24354, 54, 2552345, 2455, 245435],
+                [32.688980, -117.192040, "", 4525, 4523545, 24354, 54, 2552345, 2455, 245435],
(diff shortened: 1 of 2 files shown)


dtsnell4:
@GordonSmith @mlzummo @jbrundage 

Here is rawgit for demo: http://rawgit.com/dtsnell4/Visualization/Table_Widget_Fixed_Header_542/demos/dermatology.html?src/other/Table

I obviously multiplied the testData to be able to see the scrollbars.

Lint is failing due to not finding file "gulp-mocha-phantom.js".  Please advise.

dtsnell4:
One other thing popped up this afternoon as I was putting together this PR.  @mlzummo received an email (and I believe @GordonSmith  did as well) from someone asking for this feature, but just for the header, not the first column.  I implemented it as either both or none.  Should I go back and separate the two, so you can have either-or?  Or would that be a future release?

mlzummo:
yes i think you should

src/other/Table.js
```diff
@@ -17,36 +17,69 @@
...
+                [37.665074, -122.384375, "green-dot.png", 4525, 423545, 24354, 54, 23552345, 2455, 245435],
+                [32.690680, -117.178540, "", 4525, 423545, 24354, 54, 2552345, 2455, 2405435],
+                [39.709455, -104.969859, "", 4525, 423545, 24354, 54, 2552345, 2455, 2495435],
+                [41.244123, -95.961610, "", 4525, 423545, 24354, 54, 2552345, 2455, 2454835],
+                [32.688980, -117.192040, "", 4525, 423545, 24354, 54, 2552345, 2455, 2475435],
+                [45.786490, -108.526600, "", 4525, 423545, 24354, 54, 2552345, 2455, 2456435],
+                [45.796180, -108.535652, "", 4525, 423545, 24354, 54, 2552345, 2455, 2455435],
+                [45.774320, -108.494370, "", 4525, 423545, 24354, 54, 2552345, 2455, 2445435],
+                [45.777062, -108.549835, "red-dot.png", 4525, 423545, 24354, 54, 25523345, 2455, 245435]
             ])
```
dtsnell4: I was planning on removing the extra data once this code was finalized.  However, do we want to use a larger data set than before so developers can see the new features?

src/other/Table.js
```diff
@@ -17,36 +17,69 @@
...
+            .fixedHeader(true)
         ;
         return this;
     };
 
-    Table.prototype.publish("pagination", false, "boolean", "enable or disable pagination",null,{tags:["Private"]});
+    Table.prototype.publish("pagination", false, "boolean", "Enable or disable pagination",null,{tags:['Private']});
+//    should the following default to true?
+    Table.prototype.publish("fixedHeader", false, "boolean", "Enable or disable fixed table header and first column",null,{tags:['Private']});
```
dtsnell4: Do we want the table to default to fixed headers or normal mode?

src/other/Table.js
```diff
@@ -17,36 +17,69 @@
...
     Table.prototype.publishProxy("itemsPerPage", "_paginator");
     Table.prototype.publishProxy("pageNumber", "_paginator", "pageNumber",1);
 
     Table.prototype.enter = function (domNode, element) {
         HTMLWidget.prototype.enter.apply(this, arguments);
-        this._parentElement.style("overflow", "auto");
 
-        this.table = element.append("table");
+        this.table = element.append("table").attr("id", "table" + this._id);
```
dtsnell4: I need to remove the attribute "id" addition, my early implementation used it but no longer.

src/other/Table.js
```diff
@@ -77,7 +110,9 @@
         return {
             _id: d,
             element: function () {
-                return context.tbody.selectAll("tr").filter(function (d2) { return d2 === d; });
+                return context.tbody.selectAll("tr").filter(function (d2) { 
+                    return d2 === d; 
+                });
             }
```
dtsnell4: I can return this to its original state.  I was originally putting additional code here but removed it.

src/other/Table.js
```diff
@@ -99,7 +134,11 @@
     Table.prototype.update = function (domNode, element) {
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
-
+        this._parentElement
+            .style("position", "relative")
+            .style("overflow", "hidden")
+        ;
+
```
dtsnell4: This addition is no longer necessary and will be removed.

src/other/Table.js
```diff
@@ -219,23 +272,190 @@
...
+            divrow.selectAll(".labels-wrapper").data([0])
+                .enter()
+                .append("table")
+                .classed("labels-wrapper", true)
+            ;
+            divrow.exit().remove();
+
+            var rowsWrapper = context._parentElement.select(".rows-wrapper");
+            var colsWrapper = context._parentElement.select(".cols-wrapper");
+            var colLabelsWrapper = colsWrapper.select('.labels-wrapper');
```
dtsnell4: It turns out divcol = colsWrapper and divrow = rowsWrapper.  (it did not start that way) I will combine these.

src/other/Table.js
```diff
@@ -219,23 +272,190 @@
...
+            }
+            function _copyLabelContents(){
+                colLabelsWrapper.html(theadSelection.html());
+                colLabelsWrapper
+                    .style("width", context.table.style("width"))
+                ;
+                
+                var origThead = element.selectAll("th");
+                var newThead = context._parentElement.selectAll(".cols-wrapper th");
+                origThead[0].forEach(function(el, i){
```
dtsnell4: This should just use d3's .each() function.  Not sure why I did this.  Similar for following line

src/other/Table.js
```diff
@@ -219,23 +272,190 @@
...
+                var borderWidth = parseInt(context.table.select("td").style("border-width"));
+                var rowSelection = context.table.selectAll('tbody > tr > td:first-child');
+                var rowWrapperWidth = rowSelection.node().getBoundingClientRect().width;
+                var theadWidth = parseInt(rowWrapperWidth) + parseInt(2 * borderWidth);
+                var rowContents = "<thead><tr>";
+                rowContents += "<th style='width:" + theadWidth + "px'>";
+                rowContents += origThead[0][0].innerHTML + "</th></tr></thead>";
+                
+                rowSelection[0].forEach(function(row){
+                    rowContents += '<tr><td class="row-label">'+ row.innerHTML +'</td></tr>';
```
dtsnell4: Same as above

src/other/Table.js
GordonSmith: Its fine - we have another ticket which is a more general "move testData out of widgets" comment.  And one we can now do cleanly with the unit test framework in place.

src/other/Table.js
```diff
@@ -125,11 +164,11 @@
             })
         ;
         th.select(".thIcon")
-            .text(function (column) {
+            .text(function (column, d) {
```
GordonSmith: This might be confusing for some as we tend to use "d, idx" or "d, i" - this should be "column, idx"?

dtsnell4: You are absolutely correct. I will change it.

dtsnell4:
@GordonSmith Please re-review.  Also, do you have an opinion on separating the fixed header and the fixed first column into two distinct features?

GordonSmith:
I think it should be two publish properties (but one pull request) - I look forward to playing with this.

dtsnell4:
Ok, no problem.

GordonSmith:
This is broken on:  http://rawgit.com/dtsnell4/Visualization/Table_Widget_Fixed_Header_542/demos/test.html?src/other/Table

GordonSmith:
Closing in favor of:  https://github.com/hpcc-systems/Visualization/pull/756

Issue #703, opened 2015-08-12
2d chart mode shouldn't include nd chart types


buunguyen:
If chart mode of MultiChartSurface is set to 2D, it will list both 2D and ND chart types.

GordonSmith:
Which is correct as 2D data will display fine in NDCharts (the opposite is not true).

buunguyen:
Doesn't that mean ND mode should include 2D charts but not the other way around (as the current code does)?

GordonSmith:
No 2D data can be displayed in the ND Charts, but ND Data cannot be displayed in 2D Charts, example:
IOW How would you display this info in http://rawgit.com/hpcc-systems/Visualization/master/demos/dermatology.html?src/chart/Column in a Pie chart?

buunguyen:
I understand that but thought it means the code should do the opposite, i.e. ND mode will include charts of 2D and ND while 2D mode only include charts of 2D. I guess I don't know the context in which `mode` is used. If it's correct then it's fine, I'll close this pr.

GordonSmith:
It didn't help that it was wrong for a good while!  I read it as "show me the charts that are compatible with the given data shape."

PR #704, opened 2015-08-12
[GH-703] Remove ND chart types from 2D mode of MultiChartSurface

buunguyen:


buunguyen:
@mlzummo please help review

GordonSmith:
It should not be removed.  Those "categories" of of charts are Data shapes and it is perfectly valid to display 2D data inside a chart which supports ND Data shapes (iow if you can display it in a pie chart, you can also display it in a line chart) - however the opposite is not true.

Issue #713, opened 2015-08-13
Slider doesn't fire event during play sequence


buunguyen:
We're trying to build a similar page like [the Rugby demo](http://rawgit.com/GordonSmith/Visualization/COMMS_DEMO/demos/tmp.html) using hpcc-viz 1.4.2. However, upon playing, the click event isn't fired. After comparing the source of 1.4.2 and the source of the slider used in the Rugby demo, I noticed that in 1.4.2, `context._click()` is not invoked in the `setInterval` loop while the Rugby demo does. That seems to be the root cause. I'm not sure which version the Rugby demo uses, but can we expect the same behavior (i.e. click event triggered while the slider is played) in 1.4.4?

PR #714, opened 2015-08-13
GH-713 Fixes click event on Slider play

anmoljagetia:
Fixes GH-713 (#713)
Signed-off-by: Anmol Jagetia <anmoljagetia@gmail.com>

### DIFF
--- a/src/form/Slider.js
+++ b/src/form/Slider.js
@@ -138,6 +138,7 @@
                         .data(tick)
                         .render()
                     ;
+                    context._click();
                 } else {
                     context.pause();
                 }
@@ -146,6 +147,7 @@
                     .data(tick)
                     .render()
                 ;
+                context._click();
             }
         }, context.playInterval());
     };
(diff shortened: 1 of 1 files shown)


anmoljagetia:
@GordonSmith @buunguyen Please review!

buunguyen:
Question: is there a difference between `context.click()` and `context._click()`? Also not sure where is the latter defined?

GordonSmith:
Yes one works and the doesn't

buunguyen:
:-). Just checked the [source of the slider](http://rawgit.com/GordonSmith/Visualization/COMMS_DEMO/src/form/Slider.js) used in Rugby demo again, there's a separate `_click()` method that calls to `click()`. I guess @anmoljagetia can use the same mechanism.

anmoljagetia:
I just saw the error, and realised that the new function is itself missing. Should I also refactor [these lines](https://github.com/anmoljagetia/Visualization/blob/sliderClickFix/src/form/Slider.js#L334-L341) to use the new function ?

buunguyen:
@anmoljagetia I would say yes unless Gordon thinks otherwise. The Rugby slider does that.

GordonSmith:
Resolved by: https://github.com/hpcc-systems/Visualization/pull/716

Issue #712, opened 2015-08-13
Should remove RequireJS dependencies from bower.json?


buunguyen:
I'm not sure if this has been discussed: since hpcc-viz can work in non-AMD and even non-RequireJS environments, its bower.json shouldn't define dependencies to RequireJS, RequireJS-CSS, RequireJS-Plugins. Otherwise, in non-AMD projects, these dependencies are unnecessarily pulled to the project after every `bower install`. In addition, we have to manually exclude them to be wired to the page as we use automatic script injection mechanism.

GordonSmith:
...maybe - my goal is to make it as easy as possible for both the AMD and the non AMD folks.  Technically the dist-amd folder doesn't use the bower_dependencies at all, since its included all dependencies into the bundles (including requireJS) - moving it to package.json will probably satisfy me from a developers point of view.  (I am worried I am missing something obvious)

buunguyen:
In non-AMD: we certainly don't want `bower install` to pull over RequireJS etc.
In AMD: from what you said, we don't want to do that either because RequireJS is already bundled. (Even if it's not bundled, I don't think we should include because hpcc-viz widgets are UMD.)

So my vote to remove RequireJS from bower.json :).

I'm not sure about moving to package.json though. Currently package.json only declares dep dependencies, e.g. package required to build/test. Do we need RequireJS deps to be pulled to node_modules to build/test? Do the users when they install to their project need that?

GordonSmith:
Actually its already there and it is very much a dev dependency (I will need to add the plugins) - you were already using it for your build when I added my stuff!
Also developing the widgets without require would be too slow, especially for things like the dermatology test page.

buunguyen:
You're right. I don't remember why they weren't added to package.json as opposed to bower.json given they're dev, not lib, dependencies. So might be moving to package.json is the right way to go.

mlzummo:
@GordonSmith I think this is resolved

GordonSmith:
yup

PR #721, opened 2015-08-14
[GH-712] Move RequireJS deps to package.json and fix build

buunguyen:
This avoids RequireJS packages to be pulled to user apps. Fixes config for build to work. Also adds hpcc-viz-form.js. 

@GordonSmith please review.


package.json
```diff
@@ -3,27 +3,29 @@
...
-    "csso": "^1.3.11",
-    "gulp-git": "^1.2.3",
-    "gulp-bump": "^0.3.0",
-    "gulp-filter": "^2.0.2",
-    "gulp-tag-version": "^1.2.1",
-    "yargs": "^3.8.0",
-    "jshint-stylish": "~1.0.2",
-    "gulp-jshint": "~1.10.0",
-    "gulp-jscs": "~1.6.0"
+    "requirejs-plugins": "^1.0.2",
```
GordonSmith: This needs to use my branch, otherwise the visualizations will fail to load in DOJO AMD environments.

GordonSmith:
Can you rebase this for master (which will be 1.6.0) shortly - mainly because there are a bunch of additional tests which make catch some "post build" issues.

buunguyen:
@GordonSmith will do. Is there any chance we can make this fix land on 1.4.4? It will help us get rid of a bower exclusion workaround we have to add in all projects. In fact, with all the fixes so far for 1.4.4, I hope we'll get rid of all workarounds.

GordonSmith:
I suspect 1.4.4 and 1.6.0 will be released around the same time.  If you want it in 1.4.4 I would also suggest not sorting the requires (as its contributing to the upmerge clash)

buunguyen:
@GordonSmith npm does the sorting itself, probably a new feature. But I can certainly rearrange it manually.

I didn't know 1.6.0 and 1.4.4 can be released on the same day. What's the difference between them, as I need to use one of them. As long as the non-AMD fix + other bugs/prs I submitted go into one of these version, I'll use it. So far I've been thinking they go to 1.4.4, not sure if these commits will be rebased to 1.6.0?

GordonSmith:
1.4.4 should be 100% bug fixes to 1.4.2, 1.6.0 will include new features (which I consider this to be) (a 2.x.x. release would signify breaking changes)

buunguyen:
@GordonSmith got you. Can I assume all fixes in 1.4.4 will also go to 1.6.0?

GordonSmith:
@buunguyen Yes

Issue #738, opened 2015-08-19
Pagination:  Position is wrong when horizontal scroll-bar is visible


GordonSmith:

mlzummo:
@GordonSmith do we have an example and does the changes to table still have this issue?

PR #756, opened 2015-08-19
GH-542 Fixed table header and first column

GordonSmith:
Adds functionality for a fixed table header and fixed first column for the table widget.

Fixes GH-542

Signed-off-by: Dan Snell <dan.snell@lexisnexis.com>

### DIFF
--- a/src/other/Table.css
+++ b/src/other/Table.css
@@ -5,7 +5,18 @@
     border-collapse: collapse;
 }
 
-.other_Table th {
+.other_Table .tableDiv {
+    position: absolute;
+}
+
+/*Please take note of text selection on the table, especially in regards to the fixed first column, as well as shift-clicking.  The following code would disable text selection completely.  Which would be the lesser evil? (uncomment in browser to see the difference)*/
+.other_Table, .labels-wrapper {
+/*    -webkit-user-select: none;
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jbrundage, @mlzummo this Dans table header code with some fixes (still not perfect) can you have a play and let me know if we should pull it?  
http://rawgit.com/GordonSmith/Visualization/Table_Widget_Fixed_Header_542/demos/dermatology.html?src/other/Table

jbrundage:
Ideally it should pull the styles for the fixed header/first-column from the table rather than from CSS.

Nothing seems to be broken per se... though enhancements could be added later. 

I suggest pulling it.

mlzummo:
i second pulling, however does this issue still exist with this version?

https://github.com/hpcc-systems/Visualization/issues/738

GordonSmith:
@jbrundage I am kind of at the stage with it where it works as was written - my main change was moving the different parts inside a root div, so it worked ok in SVG Surface and some minor tidy ups to help me understand the code. If I was to carry on I would probably end up re-structuring it to be more D3 like. 
@mlzummo Its better, but not perfect (it scrolls with the HScroll bar).

PR #780, opened 2015-08-25
Fixes Table Widget

dtsnell4:
Fixes Table Widget fixed header and column in SVG surfaces as well as issues with row selection and sorting.


dtsnell4:
@GordonSmith @mlzummo @jbrundage I fixed the table so it displays properly in the SVG surfaces.  There were also some issues with row selection that I fixed.  

@GordonSmith did you see my question about separating the header and column into two publish parameters so they can be turned on independently?  I started on that before I left, do you want me to continue or move on?

GordonSmith:
@dtsnell4 If its a quick addition it might be worth while.

Issue #771, opened 2015-08-21
Table pagination bugs in 1.4.6


buunguyen:
There are two separate bugs although could be related:

* Items-per-page isn't honored
* After changing page, the row count is automatically changed

Demo page: https://github.com/buunguyen/Visualization/commit/9f0eedeab62b32c81c95c284e69e0ec36ad0c1bc.

GordonSmith:
Resolved by:  https://github.com/hpcc-systems/Visualization/pull/804

PR #792, opened 2015-08-26
GH-771 TABLE (AUTO CALC ITEMS PER PAGE) AND PAGINATOR UPDATE

mlzummo:
@buunguyen @GordonSmith 

Please Review

http://rawgit.com/mlzummo/Visualization/GH-771_TABLE_PAGINATOR_UPDATE/demos/dermatology.html?src/other/Table

Adds new publish parameter autoCalcIPP
Changes Paginator so that its fixed height
Adds new features to Paginator (jump to page)

Fixes GH-771

### DIFF
--- a/src/other/Paginator.js
+++ b/src/other/Paginator.js
@@ -16,15 +16,44 @@
         this._numList = []; //pn
     }
     Paginator.prototype = Object.create(HTMLWidget.prototype);
+    Paginator.prototype.constructor = Paginator;
     Paginator.prototype._class += " other_Paginator";
 
-    Paginator.prototype.publish("itemsPerPage", 2, "number", "Pagination items per page",null,{tags:['Private']});
-    Paginator.prototype.publish("numItems", 10, "number", "Pagination total number of items",null,{tags:['Private']});
-    Paginator.prototype.publish("pageNumber", 1, "number", "Pagination set or get the page number",null,{tags:['Private']});
+    Paginator.prototype.publish("itemsPerPage", 2, "number", "Pagination items per page",null,{tags:["Private"]});
+    Paginator.prototype.publish("numItems", 10, "number", "Pagination total number of items",null,{tags:["Private"]});
+    Paginator.prototype.publish("pageNumber", 1, "number", "Pagination set or get the page number",null,{tags:["Private"]});
+
+    Paginator.prototype.testData = function() {
+        return this;
+    };
 
     Paginator.prototype.enter = function (domNode, element) {
         HTMLWidget.prototype.enter.apply(this, arguments);
+        var context = this;
+
         this.paginator = element.append("ul").attr("class","paginator pagination pagination-sm");
+        this.side = element.append("div").attr("class","paginator pagination side");
+
+        this.side.append("span")
+            .classed("side", true)
+            .text("Page ")
+        ;
+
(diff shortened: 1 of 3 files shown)


jbrundage:
**currentPageNumber** text input doesn't change the page when its value is changed

GordonSmith:
Can you add some "float" magic to prevent this:
image

GordonSmith:
Plus you have some lint issue

GordonSmith:
And the auto calc has issues:  
image

mlzummo:
@GordonSmith for the first screenshot i cant reproduce this also the autocalc the issue isnt apparent can you explain whats wrong?

GordonSmith:
1.  Make your window smaller so you get a horizontal scroll bar and then scroll it as per the pic, you will see the pagination numbers get clipped on the LHS so you can't even see them.
2.  The auto calc has got the number of rows wrong as you can't see the pagination buttons.  V-Overflow=hidden might fix it?

GordonSmith:
Your calculation is wrong somewhere:  
image

mlzummo:
@GordonSmith Please Re-Review - Fixed calc issue (please see where i used clientHeight with a d3 first element selection) would like to know if that looks ok there

src/other/Paginator.js
```diff
@@ -23,9 +23,38 @@
...
+    };
+
     Paginator.prototype.enter = function (domNode, element) {
         HTMLWidget.prototype.enter.apply(this, arguments);
-        this.paginator = element.append("ul").attr("class","paginator pagination pagination-sm");
+        var context = this;
+
+        this.container = element.append("div");
+        this.paginator = this.container.append("ul").attr("class","paginator pagination pagination-sm");
+        this.side = this.container.append("div").attr("class","paginator pagination side");
```
mlzummo: if you can think of a better name than .side please let me know

jbrundage: Dan's fixedHeader and fixedColumn use **cols-wrapper** and **rows-wrapper**. In your case, **pagination-wrapper** seems appropriate.

src/other/Paginator.js
```diff
@@ -23,9 +23,38 @@
...
+
+        this.side.append("span")
+            .classed("side", true)
+            .text("Page ")
+        ;
+
+        this.side.append("input")
+            .attr("type","text")
```
jbrundage: If you change this to "number" it would allow the user to use the up/down arrow keys to switch between pages (when the input has focus).

GordonSmith:
Ok this is looking good (and I am tempted to default it on - some comments:
1 I think I would prefer if it was right justified (see third point for related issue).
2 Clicking on the first, 1,2,3, last buttons should update the value in the edit field to match the current page.
3 When there are only a few rows on the last page it would be nice if the paginator stayed at the bottom of the visible area, at the moment its always below the last row drawn - I am thinking absolute positioning will solve this and 1?

src/other/Paginator.js
mlzummo: @jbrundage .paginator is fine, i meant the "this.side" variable ... unless i do leftContainer/rightContainer for paginiator and .side ?

mlzummo:
closing reissuing against 1.4.8

Issue #695, opened 2015-08-12
Axis tick formatting.


GordonSmith:
See:  https://github.com/mbostock/d3/wiki/Formatting#d3_format

Issue #816, opened 2015-08-31
Allow functions to be used as publish properties


GordonSmith:
Allow things like:

```
mySHape = new SHape()
  .colorFill(function() { return "red";})}
;
```

GordonSmith:
Closed "timed-out", please reopen if needed.

PR #819, opened 2015-08-31
GH-816 Allow properties to be functions

GordonSmith:
var myShape = new Shape()
  .colorFill(function() { return "red";})}
;

Fixes GH-695

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -176,7 +176,7 @@
             }
         }
         var themeSelectObj = {
-            'Default':{},
+            'Default': null,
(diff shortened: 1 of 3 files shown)


GordonSmith:
@jbrundage Please Review - I refactored how the themes were working (I think I also fixed an issue with them:  Open Shap Widget, change theme, switch to Surface Widget, select default theme - would not reset),

jbrundage:
Very good, the *origDefaultValue* implementation makes more sense than *trueDefaultValue*

I like the capability of "functions" as param values... how will this work with serialization? it wont attempt to turn the function into a string, will it?

mlzummo:
looks good, looks like functions will be serialized as all props are "casted" into functions with d3.functor?

jbrundage:
I'm assuming Persist will grab the *value* of a param to be serialized (the returned value from the function).

drealeed:
Gordon, will this be used to support databomb aggregations, by defining functions in the ddl to perform sum, avg, max, etc?

GordonSmith:
@drea nope - I was planning on using it for the axis text formatting (but turns out I had that issue solved already).
@jbrundage The value is what is serialized (and thus deserialized - which would replace the inline function with a value) - maybe the functor support is only for non published properties.
Hmmm I need to think about this a little bit more.

GordonSmith:
Closing in favor of https://github.com/hpcc-systems/Visualization/pull/845 (which only has the theme fixes)

PR #851, opened 2015-09-02
GH-847 Google map zoom to fit issue

GordonSmith:
Zoom to fit still expected old format.

Fixes GH-847

### DIFF
--- a/src/map/GMap.js
+++ b/src/map/GMap.js
@@ -357,5 +357,27 @@
         });
     };
 
+    GMap.prototype.zoomTo = function (selection) {
+        var foundCount = 0;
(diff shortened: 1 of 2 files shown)


mlzummo:
was the issue it was just not working because it was in gmapgraph? only diff i see is that we pass this.data() to zoomtofit now: https://github.com/GordonSmith/Visualization/blob/GH-847/src/map/GMap.js#L379

GordonSmith:
The internal data shape changed from an array of json to a regular array

GordonSmith:
@mlzumm Please Review

mlzummo:
looks good

PR #853, opened 2015-09-02
GH-827 Adds Amchart Balloon

mlzummo:
Adds AMCHART BALLOON with same output formatting as HPCC Charts

Fixes GH-827


GordonSmith:
Looks fine.

Issue #895, opened 2015-09-08
Pagination Regression


GordonSmith:
@jbrundage Latest theme changes removed some new pagination features.

jbrundage:
Which pagination features were removed?

Does the problem occur here?
https://rawgit.com/hpcc-systems/Visualization/candidate-1.8.0/demos/dermatology.html?src/other/Table

GordonSmith:
Its placement and yes - size page so you get a HScroll and you will see it...

jbrundage:
@GordonSmith I can't replicate the issue you're talking about.

mlzummo:
i can't either

GordonSmith:
It only happened on the 1.8.0 branch:  http://rawgit.com/hpcc-systems/Visualization/candidate-1.8.0/demos/dermatology.html?src/other/Table&pagination=1

Resize your page until the HScroll appears - then drag it around - @mlzummo had fixed this already and then @jbrundage your theme fix for tables removed it again:
![image](https://cloud.githubusercontent.com/assets/966863/9756674/efb6078c-56d3-11e5-96f6-5d1f8e7356f6.png)

jbrundage:
I'm still unable to experience this issue. The pagination controls don't move with the scroll-bar in my browser.
![image](https://cloud.githubusercontent.com/assets/1891935/9766285/1aa563cc-56e6-11e5-8893-16ff68672ca8.png)

GordonSmith:
What URL Did you use for that?

jbrundage:
The one you provided:
http://rawgit.com/hpcc-systems/Visualization/candidate-1.8.0/demos/dermatology.html?src/other/Table&pagination=1

GordonSmith:
That because the fix got pulled:  https://github.com/hpcc-systems/Visualization/pull/896

PR #896, opened 2015-09-08
GH-895 Table Pagination Regression

GordonSmith:
Re-implemented pagination calculations and div location.
Table:  Reset sort when new data is loaded.
Pagination:  Optionally set number of adjacent pages to show.

Fixes GH-895

### DIFF
--- a/src/other/Table.js
+++ b/src/other/Table.js
@@ -15,6 +15,7 @@
         this._paginator = new Paginator();
         this._selectionBag = new Bag.Selection();
         this._selectionPrevClick = null;
+        this._paginatorTableSpacing = 4;
     }
     Table.prototype = Object.create(HTMLWidget.prototype);
     Table.prototype.constructor = Table;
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jbrunade, @mlzummo Please Review

src/other/Table.js
```diff
@@ -621,7 +623,14 @@
             }
             return context._currentSortOrder * -1;
         });
-        this.render();
+        return this;
+    };
+
+    Table.prototype.headerClick = function (column, idx) {
```
mlzummo: column not being used?

GordonSmith: Correct - but it is being passed in and a user may override and use it.

mlzummo: should change the order just to make it more clean since idx is used by default, but that might be just pulling hairs ... so you dont need to do headerClick(null, idx);

mlzummo:
looks good

Issue #897, opened 2015-09-08
Table:  Resetting table to empty causes errors


GordonSmith:
myTable.columns([]).data([])

It may be dependent on fixed header/column and or pagination etc.

GordonSmith:
@dtsnell4 - FYI you left out the "fixes GH-XXX" in your commit msg.

PR #902, opened 2015-09-08
GH-897 Reset Table Data

dtsnell4:
Resetting the table data (.columns([]) and .data([]) ) would cause the widget to error.

Fixes GH-897

### DIFF
--- a/src/other/Table.js
+++ b/src/other/Table.js
@@ -321,11 +321,17 @@
         cells.exit()
             .remove()
         ;
-
-        if (this.data().length) {
-            this.fixedLabelsUpdate(domNode, element);
+        if (!this.data().length) {
+            d3.select(".cols-wrapper").remove();        
+        }
+        if (!this.columns().length) {
+            d3.select(".rows-wrapper").remove();
         }
 
+        if (this.data().length || this.columns().length) {
+            this.fixedLabelsUpdate(domNode, element);
+        }
+        
         this._paginator
             .right((this.hasVScroll(this.tableDiv) ? this.getScrollbarWidth() : 0 ) + this._paginatorTableSpacing)
             .bottom((this.hasHScroll(this.tableDiv) ? this.getScrollbarWidth() : 0) + this._paginatorTableSpacing)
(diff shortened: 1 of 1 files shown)


dtsnell4:
@GordonSmith @jbrundage @mlzummo Please review

dtsnell4:
http://rawgit.com/dtsnell4/Visualization/GH-897_Table_Reset/demos/dermatology.html?src/other/Table

src/other/Table.js
```diff
@@ -322,9 +322,11 @@
             .remove()
         ;
 
-        if (this.data().length) {
-            this.fixedLabelsUpdate(domNode, element);
+        if (!this.data().length || !this.columns().length) {
+            this.fixedHeader(false);
+            this.fixedColumn(false);
```
GordonSmith: If I clear out the data and then load more in - I would expect it to still have fixed headers/columns.

dtsnell4:
@GordonSmith @jbrundage @mlzummo Please re-review.  New strategy.

src/other/Table.js
```diff
@@ -322,10 +322,12 @@
             .remove()
         ;
 
-        if (this.data().length) {
+        d3.select(".rows-wrapper").remove();
+        d3.select(".cols-wrapper").remove();
```
GordonSmith: Does this mean the header and column is re-rendered every update cycle?

dtsnell4: Good point, let me refactor that.

src/other/Table.js
```diff
@@ -321,11 +321,17 @@
         cells.exit()
             .remove()
         ;
+        if (!this.data().length) {
+            d3.select(".cols-wrapper").remove();        
+        }
+        if (!this.columns().length) {
+            d3.select(".rows-wrapper").remove();
+        }
```
dtsnell4: @GordonSmith I made these conditional so they only execute when the appropriate data is resetting.  I may even be over thinking this.  When the data is emptied and then new data is added, the header and footer re-render fine.  These statements are ONLY necessary when the data is emptied and then the table is rendered.  The fixed header and footer code never runs so they are left in the dom.  Is this even a realistic scenario or am I getting carried away?

PR #907, opened 2015-09-09
GH-904 Choropleth Tooltips displaying "undefined"

GordonSmith:
Fixes GH-904


GordonSmith:
@mlzummo Please Review

src/map/ChoroplethCounties.js
```diff
@@ -22,7 +22,7 @@
 
         this.columns(["County", "Weight", "Label"]);
```
mlzummo: columns and data index count doesnt match?

mlzummo: columns and data index count doesnt match?

GordonSmith: Fixed (but it was only in testData anyway)

mlzummo:
looks good

Issue #909, opened 2015-09-09
Marshaller:  First click on table does not have changed === true.


GordonSmith:
http://viz.hpccsystems.com/v1.2.2/apps/legacyDash.html?Hostname=10.241.100.159&Port=8010&Protocol=http&Wuid=W20150908-144721&ResultName=leeddx_stocktest_Comp_Ins141_DDL

PR #913, opened 2015-09-10
GH-909 HipieDDL _changed flag is set incorrectly

GordonSmith:
_changed was being set based on the previous request.  It needs also take into
consideration which widget triggers the request (both might send in the same
value, but the updates will be different).

Fixes GH-909


dtsnell4:
Looks good

PR #929, opened 2015-09-15
testing if build process is broken

mlzummo:


mlzummo:
@GordonSmith I think it is safe to say the build process is broken somehow for cand 1.8

GordonSmith:
Looks ok to me (now that I restarted it)!  Something weird is happening at travis I think...

Issue #747, opened 2015-08-19
Ability show what widget will update another and filter fields


mlzummo:

jbrundage:
What is this?

From the title it seems like the goal is to somehow visually indicate which widgets will update other widgets? Any thoughts on how this should be accomplished?

GordonSmith:
Its only of interest in the HTML Marshaller and I can expose a method which will tell you - but I suspect when a user moves over one widget which updates others, the others should get highlighted (not dissimilar to some of the drag and drop highlighting?

GordonSmith:
Further - it might end up being related to how we may want to highlight items in the legend and the chart as well?

jbrundage:
So I'll add an onHover method to layout/Cell.js like this:
```javascript
Cell.prototype.onHoverEnter = function (widgetArr){
    for(var i in widgetArr){
        this.createOverlayUpdateIndicator(widgetArr[i]);
    }
}
Cell.prototype.onHoverExit = function (widgetArr){
    for(var i in widgetArr){
        this.removeOverlayUpdateIndicator(widgetArr[i]);
    }
}
```
And I'll call **onHover(...with the response from your exposed method...)**. Does this make sense or am I misunderstanding?

jbrundage:
@GordonSmith Do you have any input regarding this enhancement?

GordonSmith:
Its only going to be used in marshaller/HTML and I will need to add a method that will tell you the list to highlight...

jbrundage:
In #947 I added an 'array' publish param that will highlight dom elements by ID. 

To be used like this:
Grid.content()[0].indicateTheseIds(*["_w10","_w15","_w27"]*)

Can your method simply return the dom IDs of the widgets to highlight?

GordonSmith:
Yes it can.

PR #947, opened 2015-09-21
GH-747 Adds Hover Indicators to layout/Cell.js

jbrundage:
Fixes GH-747

@GordonSmith @dtsnell4 @mlzummo I've added **indicateTheseIds** publish param to layout/Cell.js

It holds an array of DOM IDs and displays a div over each corresponding html element when the use hovers over the Cell.

https://rawgit.com/jbrundage/Visualization/GH-747-HoverIndicators/demos/dermatology.html?src/layout/Grid.hoverIndicator

### DIFF
--- a/src/layout/Cell.js
+++ b/src/layout/Cell.js
@@ -1,11 +1,11 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["./Surface"], factory);
+        define(["d3", "./Surface"], factory);
     } else {
-        root.layout_Cell = factory(root.layout_Surface);
+        root.layout_Cell = factory(root.d3, root.layout_Surface);
     }
-}(this, function (Surface) {
+}(this, function (d3, Surface) {
     function Cell() {
         Surface.call(this);
         this._dragHandles = ["nw", "n", "ne", "e", "se", "s", "sw", "w"];
@@ -19,6 +19,12 @@
     Cell.prototype.publish("gridRowSpan", 1, "number", "Grid Row Span",null,{tags:["Private"]});
     Cell.prototype.publish("gridColSpan", 1, "number", "Grid Column Span",null,{tags:["Private"]});
     Cell.prototype.publish("handleSize", 6, "number", "Grid Row Position",null,{tags:["Private"]});
+    
+    Cell.prototype.publish("indicateTheseIds", [], "array", "Array of DOM Ids to display update-indicators over.",null,{tags:["Private"]});
+    
+    Cell.prototype.publish("indicatorGlowColor", "#EEEE11", "html-color", "Glow color of update-indicator",null,{tags:["Basic"]});
+    Cell.prototype.publish("indicatorBorderColor", "#F48A00", "html-color", "Border color of update-indicator",null,{tags:["Basic"]});
+    Cell.prototype.publish("indicatorOpacity", 0.8, "number", "Opacity of update-indicator",null,{tags:["Basic"]});
 
     Cell.prototype.enter = function (domNode, element) {
         Surface.prototype.enter.apply(this, arguments);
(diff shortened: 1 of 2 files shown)


GordonSmith:
That is not really what I am expecting (or might be a bug) - all should look normal, but when I hover over one cell, it should subtly change color(say very very pale yellow) and then the "indicateTheseIDs" list should be highlighted in a slightly less subtle shade of yellow (one of the standard "highlighter" colors).

src/layout/Cell.js
```diff
@@ -19,16 +19,29 @@
...
+    Cell.prototype.publish("indicatorColor", "#FF00FF", "html-color", "Background color of update-indicator",null,{tags:["Basic"]});
+    Cell.prototype.publish("indicatorOpacity", 0.2, "number", "Opacity of update-indicator",null,{tags:["Basic"]});
 
     Cell.prototype.enter = function (domNode, element) {
         Surface.prototype.enter.apply(this, arguments);
         element.classed("layout_Surface", true);
+        
+        //Example values...
+        this.indicateTheseIds(["_w43","_w59"]);
+
```
GordonSmith: Can you move this into its own Test in test/layoutFactory - I suspect the issue I think I am seeing is becuase this is being set for every cell...

jbrundage: @GordonSmith Done

GordonSmith:
Ok that is working better, I think the highlighting should be done at the cell/surface border level rather than the content - thoughts?  Here is what I am seeing - I would expect the titlebar and border to be used to indicate and leave the content alone:
image

jbrundage:
@GordonSmith The coloring is just a fixed div placed overtop of whichever ID it's given. If it should be applied at the Cell/Surface level the method that specifies which widgets need to be highlighted could just return the IDs of the Cell/Surface.

IMO if we modify the Surface styles directly it's a bit messier of an implementation (and puts limitations on how this feature can be used going forward).

dtsnell4:
What if you just removed the background color of the div, expanded it to the size of the Surface container and changed it to a box-shadow to give it an "outer glow" effect?

jbrundage:
@dtsnell4 That would work well. I'll update this PR with that change...

src/layout/Cell.js
```diff
@@ -139,6 +148,38 @@
...
+                div.className = "update-indicator";
+                div.style.position = "fixed";
+                div.style.top = rect.top + "px";
+                div.style.left = rect.left + "px";
+                div.style.width = rect.width + "px";
+                div.style.height = rect.height + "px";
+                div.style.opacity = opacity;
+                div.style.padding = "0px";
+                div.style.zIndex = 1000;
+                div.style.backgroundColor = bgColor;
```
GordonSmith: I would rather you used ```d3.append().style({xxx:yyy});``` paradigm:
* d3 fixes a few cross browser inconsistencies 
* We could then use transitions if we wanted to
* general code consistency

src/layout/Cell.js
```diff
@@ -139,6 +148,38 @@
...
+            }
+        };
+    };
+    Cell.prototype.onMouseLeave = function (){
+        var element = this._element;
```
GordonSmith: ```this.element()``` is preferable.

GordonSmith:
@jbrundage bump.

jbrundage:
@GordonSmith @dtsnell4 I've made the suggested changes. Please review

https://rawgit.com/jbrundage/Visualization/GH-747-HoverIndicators/demos/dermatology.html?src/layout/Grid.hoverIndicator

dtsnell4:
That looks good to me.

mlzummo:
@jbrundage it doesnt always turn off the indication when you mouse out

GordonSmith:
@jbrundage Code looks ok - I tried to tweak the colours in the dermatology page, but the new values didn't seem to change?

jbrundage:
@GordonSmith I looked into it just now and this is a separate bug. The PropertyEditor is reversing the order of the "Cell" widgets. You would have to change the indicatorColor params in the bottom two "Cell" sections to see the color change.

GordonSmith:
@jbrundage Can you open a new issue for that?

Issue #950, opened 2015-09-22
New Property Editor + IE


GordonSmith:
https://github.com/hpcc-systems/Visualization/blame/candidate-1.8.0/src/other/PropertyEditor.js#L95

matches is not available on IE...

jbrundage:
@GordonSmith @mlzummo #955 fixes this issue, but it's failing the Travis CI build for unrelated reasons.

PR #955, opened 2015-09-23
GH-950 Replaces matches with filter in PropEdit

jbrundage:
Fixes GH-950

@GordonSmith @dtsnell4 Please review

### DIFF
--- a/src/other/PropertyEditor.js
+++ b/src/other/PropertyEditor.js
@@ -92,7 +92,7 @@
     PropertyEditor.prototype.enter = function (domNode, element) {
         Accordion.prototype.enter.apply(this, arguments);
         
-        if(!this._parentElement.node().matches(".other_PropertyEditor div")){
+        if(this._parentElement.filter(".other_PropertyEditor div").empty()){
             this._parentElement.style("overflow", "auto");
         }
         


src/other/PropertyEditor.js
```diff
@@ -92,7 +92,7 @@
     PropertyEditor.prototype.enter = function (domNode, element) {
         Accordion.prototype.enter.apply(this, arguments);
         
-        if(!this._parentElement.node().matches(".other_PropertyEditor div")){
+        if(this._parentElement.filter(".other_PropertyEditor div")[0].length === 0){
```
GordonSmith: I am not 100% sure what your testing for here, but would the following work:
```
if(this._parentElement.filter(".other_PropertyEditor div").empty()){
```

jbrundage: Yes, it would. It's a conditional to test whether or not "this" is the outer-most instance of PropertyEditor.

GordonSmith: I prefer the empty() method to the ```[0].length === 0``` as the in the latter you are breaking the object encapsulation (your assuming knowledge of the inner workings which may change in the future).

jbrundage: Agreed, ill make the changes now

PR #956, opened 2015-09-25
Adds XYAxis Label Rotation

mlzummo:
@GordonSmith @dtsnell4 @jbrundage - Please Review

Adds XYAxis Label Rotation

Fixes GH-597

Signed-off-by: Mathew Zummo mzummo@gmail.com

### DIFF
--- a/src/chart/XYAxis.js
+++ b/src/chart/XYAxis.js
@@ -26,7 +26,8 @@
     XYAxis.prototype.publish("xAxisTypeTimePattern", "%Y-%m-%d", "string", "Time Series Pattern");
     XYAxis.prototype.publish("xAxisDomainLow", "", "string", "X-Axis Low");
     XYAxis.prototype.publish("xAxisDomainHigh", "", "string", "X-Axis High");
-    XYAxis.prototype.publish("xAxisOverlapMode", "stagger", "set", "X-Axis Label Overlap Mode", ["none", "stagger", "hide"]);
+    XYAxis.prototype.publish("xAxisOverlapMode", "rotate", "set", "X-Axis Label Overlap Mode", ["none", "stagger", "hide", "rotate"]);
+    XYAxis.prototype.publish("xAxisLabelRotation", null, "number", "X-Axis Label Rotation");
 
     XYAxis.prototype.publish("yAxisTitle", "", "string", "Y-Axis Title");
     XYAxis.prototype.publish("yAxisTickCount", null, "number", "Y-Axis Tick Count", null, { optional: true });
@@ -203,7 +204,7 @@
             return d.map(function (item) {
                 return this.formatValue(item);
             }, this);
-        } 
+        }
         switch (this.yAxisType()) {
             case "time":
                 return this._dateParserValue(d);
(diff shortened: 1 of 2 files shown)


mlzummo:
http://rawgit.com/mlzummo/Visualization/XYAXIS_LABEL_ROTATION_2/demos/dermatology.html?src/chart/Column

src/chart/XYAxis.js
```diff
@@ -368,6 +369,27 @@
...
+                        var e_bbox = elm.node().getBoundingClientRect();
+                        if (xRotation > 0 && xRotation <= 90) {
+                            elm.style("text-anchor", "start");
+                            elm.attr("transform", "rotate(" + (xRotation ? xRotation : 0) + ")");
+                            elm.attr("dy", "-.5em");
+                            elm.attr("dx", "8px");
+                        } 
+                        else if (xRotation > 90 && xRotation < 180) {
+                            elm.style("text-anchor", "start");
+                            elm.attr("transform", "rotate(" + (xRotation ? xRotation : 0) + ")scale(-1,-1)translate(" + (-e_bbox.width) + "," + (-e_bbox.height) + ")");
```
GordonSmith: When rotate is disabled there is no code branch which sets this back to 0.

src/chart/XYAxis.js
```diff
@@ -368,6 +369,27 @@
...
                 break;
+            case "rotate" :
+                var xRotation = this.xAxisLabelRotation();
+                xAxis.selectAll(".tick > text")
+                    .each(function() {
+                        var elm = d3.select(this);
+                        var e_bbox = elm.node().getBoundingClientRect();
+                        if (xRotation > 0 && xRotation <= 90) {
+                            elm.style("text-anchor", "start");
+                            elm.attr("transform", "rotate(" + (xRotation ? xRotation : 0) + ")");
```
GordonSmith: When rotate is disabled there is no code branch which sets this back to 0.

src/chart/XYAxis.js
```diff
@@ -368,6 +369,27 @@
                     .attr("visibility", function (d, i) { return i % margin.overlapModulus ? "hidden" : null; })
```
GordonSmith: When hide is disabled there is no code branch which sets this back to visible.

GordonSmith:
1. The overlap rotation should only be applied when the labels overlap (margin.overlapModulus > 1)
2. For angles 1->90 (the common case) it should use text anchor-end as per our IM conversation
3. When changing from rotate->stagger there is no code path which resets the rotation back to 0
4. Angles of -45 (or 135) don't render as expected

mlzummo:
@GordonSmith 
angles should range from 0-360 degrees ... its currently set to operate in clock-wise fashion ... but if we want to default the text reading left to right with beginning at the bottom (which will then allow us to use text anchor end for 1->90 cases) angles will then be 0-360 in reverse clockwise fashion ... 135 atleast to me right now seems to be rendering as i would expect it (then again i coded it to work exactly that way it is right now ...) ... ill make the suggested changes in a new branch

mlzummo:
@GordonSmith 

There also looks to be more complexity to get the functionality you want ... getting d3 to rotate backwards for example which i can do by converting what angle we give it to the number d3 actually needs to rotate it correctly ... but i will look at this more as well.
 
http://grab.by/KIeW

Issue #964, opened 2015-09-28
WU Filtering uses case sensitive field names


GordonSmith:
Should fallback to lowercase.

PR #966, opened 2015-09-28
GH-964 WU Filtering Field IDs

GordonSmith:
Should fallback to case insensitive field matching.

Fixes GH-964


GordonSmith:
@drealeed Please Review

Something external must have changed, as the existing code has been there since Sept 2014: https://github.com/hpcc-systems/Visualization/blame/master/src/other/Comms.js#L779

drealeed:
I've tested my visualizations against this branch and the problem's fixed. Reviewed & signed off

PR #976, opened 2015-10-01
GH-957 Adds c3chart/Combo.js

jbrundage:
Fixes GH-957

@mlzummo @dtsnell4 Please review

I added a more complex sample to show capabilities of the various combinations of chart types (We should probably do the same with the other libraries:
http://rawgit.com/jbrundage/Visualization/C3-ComboChart/demos/dermatology.html?src/c3chart/Combo

EDIT: Replaced sample URL with one that works

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -300,7 +300,13 @@
                     currWidget = widget;
                     if (params) {
                         for (var key in params) {
-                            currWidget[key](params[key]);
+                            if(currWidget["__meta_"+key] !== undefined){
+                                if(currWidget["__meta_"+key].type === "array"){
+                                    currWidget[key](params[key].split(","));
+                                } else {
+                                    currWidget[key](params[key]);
+                                }
+                            }
                         }
                     }
                     main
--- a/src/c3chart/Combo.js
+++ b/src/c3chart/Combo.js
@@ -1,0 +1,71 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["./CommonND"], factory);
+    } else {
+        root.c3chart_Combo = factory(root.c3chart_CommonND);
+    }
+}(this, function (CommonND) {
+    function Combo(target) {
+        CommonND.call(this);
+
+        this._type = "bar";
+        this._previousTypes = undefined;
+    }
+    Combo.prototype = Object.create(CommonND.prototype);
+    Combo.prototype.constructor = Combo;
+    Combo.prototype._class += " c3chart_Column";
+
+    Combo.prototype.publish("stacked", false, "boolean", "Stack Chart",null,{tags:["Basic"]});
+    Combo.prototype.publish("defaultType", "bar", "set", "Default chart type", ["bar","line","spline","area","area-spline","step","area-step","scatter"],{tags:["Basic"]});
+    Combo.prototype.publish("types", [], "array", "Array of chart types (ex:bar|line|spline|area|area-spline|step|area-step|scatter)",null,{tags:["Basic"]});
+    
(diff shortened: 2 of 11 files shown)


demos/dermatology.html
```diff
@@ -296,7 +296,9 @@
                     currWidget = widget;
                     if (params) {
                         for (var key in params) {
-                            currWidget[key](params[key]);
+                            if(typeof(currWidget[key]) === "function"){
```
mlzummo: i suspect there are other areas of of code that can use this same fix in the library?

jbrundage: Potentially

demos/dermatology.html
```diff
@@ -309,6 +311,7 @@
                             displaySerialization(currWidget);
                             displaySerializationText(currWidget);
                             displayThemeText(currWidget);
+                            frame.doResize();
```
GordonSmith: Why is this needed?

src/c3chart/Combo.js
```diff
@@ -0,0 +1,71 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["./CommonND"], factory);
+    } else {
+        root.c3chart_Column = factory(root.c3chart_CommonND);
```
GordonSmith: Wrong root name

GordonSmith:
Lint is failing

demos/dermatology.html
jbrundage: I'm not sure how that ended up there... I'll investigate later today

jbrundage: At one point this was needed otherwise the labels were cut off until after the first update. This issue doesn't seem to be happening anymore. I'll remove this.

GordonSmith:
@jbrundage Sample page is failing to load in Chrome.
@mlzummo @dtsnell4 Please Review

jbrundage:
@GordonSmith I've added a working link to the top comment in this thread

src/c3chart/Combo.js
```diff
@@ -0,0 +1,71 @@
...
+        CommonND.call(this);
+
+        this._type = "bar";
+        this._previousTypes = undefined;
+    }
+    Combo.prototype = Object.create(CommonND.prototype);
+    Combo.prototype.constructor = Combo;
+    Combo.prototype._class += " c3chart_Column";
+
+    Combo.prototype.publish("isStacked", false, "boolean", "Stack Chart",null,{tags:["Basic"]});
```
GordonSmith: I don't think "isStacked" does what I expect it to do.  Also in chart/Column we have a property called "stacked" is it the same thing?

src/c3chart/Combo.js
```diff
@@ -0,0 +1,71 @@
...
+    Combo.prototype.publish("baseType", "bar", "set", "Base chart type",["bar","line","spline","area","area-spline","step","area-step","scatter"],{tags:["Basic"]});
```
GordonSmith: What is baseType?  looks like the default type?

jbrundage: Yes, it's the default chart type (I'll change it to defaultType, that makes more sense)

src/c3chart/Combo.js
jbrundage: Upon further inspection... this Combo chart's "isStacked" mode stacks the values of each individual chart type in the same way that AMCharts stacks their values while in stacking mode.

jbrundage: @GordonSmith 

Notice that the **bar** values are stacked additively while the **line** values aren't effected by stack (because only one data set is using **line**):
http://rawgit.com/jbrundage/Visualization/C3-ComboChart/demos/dermatology.html?src/c3chart/Combo&types=bar,bar,line&isStacked=1

GordonSmith: I suspect my use of "stacked" in hpcc charts is the odd one out.  Can you rename isStacked to stacked?

PR #979, opened 2015-10-02
Allow URL to specify which test to display.

GordonSmith:

### DIFF
--- a/demos/dermatology.html
+++ b/demos/dermatology.html
@@ -140,6 +140,7 @@
     <script>
         var def = window.location.search.split("?")[1];
         var widgetPath = "";
+        var widgetTest = "";
         var params = {};
         if (def) {
             def.split("&").forEach(function (param, idx) {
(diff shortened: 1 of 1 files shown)


mlzummo:
injecting Publish Params would be nice to have again too, can you add that to this PR... i think we can assume that a dataset must be displayed when doing so

demos/dermatology.html
```diff
@@ -375,7 +379,7 @@
                 ;
                 if (first) {
```
mlzummo: when would this ever be false btw?

GordonSmith: This is inside the resize method - and the last line of code calls ``` frame.doResize();``` so on the "first call" we want to process the url, but never again.

GordonSmith:
The publish param injection never went away:
http://rawgit.com/GordonSmith/Visualization/DERM_TEST_SEL/demos/dermatology.html?src/chart/Line&orientation=vertical

mlzummo:
ahh i see, GH minimized the area of code where that's being done and i thought it was missing

GordonSmith:
@mlzummo Good to pull?

mlzummo:
yes

Issue #991, opened 2015-10-05
create overwritable timeout callback function in comms.js


mlzummo:

mlzummo:
not sure where this should be ... and perhaps we should make it so render does = function(widget,error); so that its called anyways and person can catch an error after setting the timout period?

mlzummo:
@GordonSmith

mlzummo:
var timeout = true;
        var timeoutCounter = 0; 
        var mTimeoutHander = setInterval(function() {
            console.log('running...')
            if (timeoutCounter >= 5) {
                clearInterval(mTimeoutHander);
                if (timeout) {
                    console.log('timeout occurred');
                    if (callback) {
                        var error = {
                            error: "timeout"
                        };
                        callback(null,error);
                    }
                }
            }
            timeoutCounter++;
        }, 1000);

        //  Parse DDL  ---
        var context = this;
        if (this.ddlUrl()[0] === "[" || this.ddlUrl()[0] === "{") {
            this.marshaller.parse(this.ddlUrl(), function () {
                populateContent();
            });
        } else {
            this.marshaller.url(this.ddlUrl(), function () {
                populateContent();
            });
        }

mlzummo:
if we do it here as well we can pass in a timeouttime Publish Param to HTML Marshaller and read it very easly

mlzummo:
.render(function (m, error) {
									if (m === null && error) {
										if (error.desc === "timeout") {
											alert("Timeout Occurred");
											jq('#'+target+' .z-apply-loading-indicator').remove();	
											return;
										}
									}

PR #992, opened 2015-10-05
GH-991 Marshaller Timeout Callback

mlzummo:
@GordonSmith Proof of concept please let me know your thoughts about how this should be handled

Adds timeout handling to HTML Marshaller

Fixes GH-991

### DIFF
--- a/src/marshaller/HTML.js
+++ b/src/marshaller/HTML.js
@@ -16,6 +16,12 @@
     HTML.prototype.publish("ddlUrl", "", "string", "DDL URL",null,{tags:["Private"]});
     HTML.prototype.publish("databomb", "", "string", "Data Bomb",null,{tags:["Private"]});
     HTML.prototype.publish("proxyMappings", {}, "object", "Proxy Mappings",null,{tags:["Private"]});
+    HTML.prototype.publish("timeout", 5, "number", "Timeout (Seconds)",null,{tags:["Private"]});
+
+    HTML.prototype.timeoutCallback = function (callback) {
+        this._timeoutCallback = callback;
+        return this;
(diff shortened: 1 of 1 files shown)


mlzummo:
timeout # could be handled by timeoutCallback as a second param if we went that route instead of handling it in render() ... there are a few ways to do this ... and prolly even diff areas this could handled...

mlzummo:
.render(function (m, error) {
									if (m === null && error) {
										if (error.desc === "timeout") {
											alert("Timeout Occurred");
											jq('#'+target+' .z-apply-loading-indicator').remove();	
											return;
										}
									}

GordonSmith:
I would get this working at the comms layer first (since this is where the timeout is actually detected and (not) handled):  https://github.com/hpcc-systems/Visualization/blob/master/src/other/Comms.js#L223-L229

The common pattern I have seen in other libraries is to handle the timeout as a second (optional) function:
So it would be ```xxx.jsonp(url, request, function(response) {...}, function(error) {...});```

This would then be propagated from the comms to the widget and onto the render call?



https://github.com/hpcc-systems/Visualization/blob/master/src/other/Comms.js#L223-L229

mlzummo:
the only problem with this approach is i will need to proogation a callback function from HTML.js to HipieDDL.js all the way down to Comms.js (to set timeout time and the callback function)

mlzummo:
also we are going to have to make many many modifications for the different types of data sources comms/hipie support

mlzummo:
neeed to figure out how to handle subsequent timeouts when new Marshaller is created in dataSource object in HippieDDL.js

Issue #1012, opened 2015-10-08
HPCC Chart (long labels get cut off)


mzummo:

mlzummo:
![image](https://cloud.githubusercontent.com/assets/1057527/10375126/c040685a-6dc4-11e5-9d2c-a64a1be1bbe5.png)

mlzummo:
also looks like stagger is making "science" move too

GordonSmith:
Closed "timed-out", please reopen if needed.

PR #1013, opened 2015-10-09
Adds XYAxis Label Rotation

mlzummo:
Fixes GH-597

### DIFF
--- a/src/chart/XYAxis.css
+++ b/src/chart/XYAxis.css
@@ -3,6 +3,10 @@
   font: 10px sans-serif;
 }
 
+.chart_XYAxis .tick > text {
+  font: 10px sans-serif;
+}
+
 .chart_XYAxis .axis path,
 .chart_XYAxis .axis line {
   fill: none;
--- a/src/chart/XYAxis.js
+++ b/src/chart/XYAxis.js
@@ -26,7 +26,8 @@
     XYAxis.prototype.publish("xAxisTypeTimePattern", "%Y-%m-%d", "string", "Time Series Pattern");
     XYAxis.prototype.publish("xAxisDomainLow", "", "string", "X-Axis Low");
     XYAxis.prototype.publish("xAxisDomainHigh", "", "string", "X-Axis High");
-    XYAxis.prototype.publish("xAxisOverlapMode", "stagger", "set", "X-Axis Label Overlap Mode", ["none", "stagger", "hide"]);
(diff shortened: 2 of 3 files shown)


mlzummo:
http://rawgit.com/mlzummo/Visualization/XYAXIS_LABEL_ROTATION_3_INVERSE/demos/dermatology.html?src/chart/Column.long-label

test/chartFactory.js
```diff
@@ -16,6 +16,17 @@
                     );
                 });
             },
+            "long-label": function (callback) {
+                require(["test/DataFactory", "src/chart/Column"], function (DataFactory, Column) {
+                    callback(new Column()
+                        .columns(DataFactory.ND.subjects.columns)
+                        .data(DataFactory.ND.subjects["data-long-label"])
```
GordonSmith: FYI - You can inline the data here - you only want it in the factory file if its shared between two samples.

src/chart/XYAxis.css
```diff
@@ -3,6 +3,10 @@
   font: 10px sans-serif;
 }
 
+.chart_XYAxis .tick > text {
+  font: 10px sans-serif;
+}
```
GordonSmith: Is this needed?

src/chart/XYAxis.js
```diff
@@ -359,19 +360,53 @@
...
                     .attr("visibility", function (d, i) { return i % margin.overlapModulus ? "hidden" : null; })
+                    .attr("transform", "rotate(0)")
+                ;
+                break;
+            case "rotate" :
+                var deg = -(this.xAxisLabelRotation()) || 0;
+                xAxis.selectAll(".tick > text")
+                    .each(function() {
+                        var elm = d3.select(this);
+                        if (deg !== 0 && margin.overlapModulus > 1) {
```
GordonSmith: This test can be done outside the "each" which will then let you conditionally fall through to the default case when modulus === 0 or deg === 0.

GordonSmith:
Test page seems to work well (there is an issue with the margin calculation, but its not really related to your change)

src/chart/XYAxis.css
mlzummo: I can remove it ... but i put it there to fix a bug i noticed i can't recall what the bug was 100% ... i believe  it was due to some css overwriting it cause the font size wasnt specific enough to the tick text

mlzummo: it allows you to set the label font size as well ... so keep it?

mlzummo:
@GordonSmith i noticed this yesterday with the margin:

https://github.com/hpcc-systems/Visualization/issues/1012

Issue #1028, opened 2015-10-13
Add MegaChart as a compilation of existing widgets


jbrundage:
Surface+Border+Text+Legend+Chart
  * Surface
    * Border
      * Left container = TextWidget = Y-Axis Title
      * Bottom container = TextWidget = X-Axis Title
      * Center container = Any chart widget
      * Right container = Legend widget attached to chart within Center container

GordonSmith:
Looks about correct - I would also like to see a button on the surface for show/hide the legend - but that could be a v2 feature?

GordonSmith:
I would also make the Surface the base class of the MegaChart (currently its the BorderLayout)

Issue #1027, opened 2015-10-13
Border layout widget height/width calculations are off


jbrundage:
Border.prototype.borderLayoutObject needs to rely on **this.width** and **this.height** rather than using *getBoundingClientRect*

GordonSmith:
Commit message didn't reference this Issue.

PR #1029, opened 2015-10-13
GH-1028 Adds MegaChart widget

jbrundage:
Fixes GH-1028
Fixes GH-1027

http://rawgit.com/jbrundage/Visualization/MEGA_CHART_CLUSTER/demos/dermatology.html?src/cluster/MegaChart

### DIFF
--- a/src/common/Text.js
+++ b/src/common/Text.js
@@ -18,6 +18,8 @@
     Text.prototype.publish("fontSize", null, "number", "Font Size (px)", null, { tags: ["Intermediate"] });
     Text.prototype.publish("anchor", "middle", "set", "Anchor Position", ["start", "middle", "end"], { tags: ["Intermediate"] });
     Text.prototype.publish("colorFill", null, "html-color", "Fill Color", null, { tags: ["Basic"] });
+    
+    Text.prototype.publish("rotation", 0, "number", "Degrees of rotation", null, { tags: ["Basic"] });
 
     Text.prototype.enter = function (domNode, element) {
         SVGWidget.prototype.enter.apply(this, arguments);
@@ -26,6 +28,7 @@
 
     Text.prototype.update = function (domNode, element) {
         SVGWidget.prototype.update.apply(this, arguments);
+        var context = this;
         this._textElement
             .attr("font-family", this.fontFamily())
             .attr("font-size", this.fontSize())
@@ -51,6 +54,7 @@
         } catch (e) {
         }
         var xOffset = 0;
+        var yOffset = 0;
         switch(this.anchor()) {
             case "start":
                 xOffset = -bbox.width / 2;
@@ -59,11 +63,18 @@
                 xOffset = bbox.width / 2;
                 break;
         }
-        var yOffset = -(bbox.y + bbox.height / 2);
-
+        var yShift = -(bbox.y + bbox.height / 2);
+        var xShift = xOffset;
+        
+        var r = xShift + yShift;
+        var theta = -this.rotation() * Math.PI/180;
+        
+        xOffset = Math.round(r * Math.sin(theta));
+        yOffset = Math.round(r * Math.cos(theta));
+        
         this._textElement
             .style("text-anchor", this.anchor())
-            .attr("transform", function (d) { return "translate(" + xOffset + "," + yOffset + ")"; })
+            .attr("transform", function (d) { return "translate(" + xOffset + "," + yOffset + ")rotate(" + context.rotation() + ")"; })
         ;
     };
 
--- a/src/composite/MegaChart.js
+++ b/src/composite/MegaChart.js
@@ -1,0 +1,96 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["../layout/Surface", "../layout/Border", "../chart/MultiChart", "../common/Text", "../other/Legend"], factory);
+    } else {
+        root.composite_MegaChart = factory(
+                root.layout_Surface,
+                root.layout_Border,
+                root.chart_MultiChart,
+                root.common_Text,
+                root.other_Legend
+            )
+        ;
+    }
+}(this, function (Surface, Border, MultiChart, Text, Legend) {
+    var _chartTypes = new MultiChart()._NDChartTypes;
+    function MegaChart() {
+        Surface.call(this);
+
+        this._tag = "div";
+        
+        this._layout = new Border();
+        
+        this._chart = new MultiChart();
+        this._valueTitle = new Text();
+        this._domainTitle = new Text();
+    }
+    MegaChart.prototype = Object.create(Surface.prototype);
+    MegaChart.prototype.constructor = MegaChart;
+    MegaChart.prototype._class += " composite_MegaChart";
+
+    MegaChart.prototype.publishProxy("valueAxisTitle","_valueTitle","text");
+    MegaChart.prototype.publishProxy("domainAxisTitle","_domainTitle","text");
+    
+    MegaChart.prototype.publish("showLegend",false,"boolean","Show/Hide Legend", null, {tags:["Basic"]});
+    
+    MegaChart.prototype.publish("gutter", 8, "number", "Gap Between Cells",null,{tags:["Basic"]});
+    
+    MegaChart.prototype.publish("chartType", "LINE", "set", "Chart Type", _chartTypes.map(function (item) { return item.id; }),{tags:["Basic"]});
+    
+    MegaChart.prototype.enter = function (domNode, element) {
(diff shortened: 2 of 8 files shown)


GordonSmith:
1. The X/Y labels should have the same font options that the Surface Title has (bold, size etc...)
2. Rather than X/Y Axis it should be "Domain" and "Value" Axis (the PP label) - I will be changing the XYAxisWidget at some point - mainly as it is very confusing when the orientation is changed?
3.  The legend shrink wrap won't work due to the Table auto resizing to consume all the space (open a new issue to add "fixedSize" PP to the Table).
4.  Rename Cluster folder to "Composite"
5.  The Composite folder will need to be added to the build and to the non AMD demo/test pages.

GordonSmith:
6.  The mega chart should probably not show any of the borders if a 1D or 2D chart is selected (or maybe they are not options for the Mega Chart?

src/other/Legend.js
```diff
@@ -95,7 +95,9 @@
         
         var table = element.select(".tableDiv > table");
         var tableRect = table.node().getBoundingClientRect();
-        var elementRect = element.node().getBoundingClientRect();
+        var elementRect = this._parentElement.node().getBoundingClientRect();
+        console.log("elementRect.height:");
+        console.log(elementRect.height);
```
GordonSmith: remove logging

src/common/Text.js
```diff
@@ -60,10 +62,15 @@
                 break;
         }
         var yOffset = -(bbox.y + bbox.height / 2);
-
+        
+        if (Math.abs(context._rotation) === 90) {
+            xOffset = -(bbox.y + bbox.height / 2);
+            yOffset = 0;
+        }
```
GordonSmith: Why is this needed for +/-90 and not for say +/-89?

src/other/Table.js
```diff
@@ -134,7 +136,13 @@
     Table.prototype.update = function (domNode, element) {
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
-       
+        if(this.fixedSize()){
+            var box = d3.select(".tableDiv > table").node().getBoundingClientRect();
+            element.style({width: box.width + "px",height: box.height + "px"});
```
GordonSmith: Would simply removing the width + height styles here be simpler?

src/composite/MegaChart.js
```diff
@@ -0,0 +1,66 @@
...
+    MegaChart.prototype = Object.create(Surface.prototype);
+    MegaChart.prototype.constructor = MegaChart;
+    MegaChart.prototype._class += " composite_MegaChart";
+
+    MegaChart.prototype.publish("xAxisTitle","","string","Title to display for X-Axis", null, {tags:["Basic"]});
+    MegaChart.prototype.publish("yAxisTitle","","string","Title to display for Y-Axis", null, {tags:["Basic"]});
+    
+    MegaChart.prototype.publish("showLegend",false,"boolean","Show/Hide Legend", null, {tags:["Basic"]});
+    
+    MegaChart.prototype.publish("gutter", 8, "number", "Gap Between Widgets",null,{tags:["Basic"]});
```
GordonSmith: "Gap Between Cells"

GordonSmith:
Looks good - some small comments/questions above.

src/composite/MegaChart.js
```diff
@@ -0,0 +1,66 @@
...
+    if (typeof define === "function" && define.amd) {
+        define(["../layout/Surface", "../layout/Border", "../common/Text", "../other/Legend"], factory);
+    } else {
+        root.composite_MegaChart = factory(
+                root.layout_Surface,
+                root.layout_Border,
+                root.common_Text,
+                root.other_Legend
+            )
+        ;
```
GordonSmith: Should be one line for consistency

src/composite/MegaChart.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+}(this, function (Surface, Border, Text, Legend) {
+    function MegaChart() {
+        Surface.call(this);
+
+        this._tag = "div";
```
GordonSmith: Not needed

src/composite/MegaChart.js
```diff
@@ -0,0 +1,66 @@
...
+    
+    MegaChart.prototype.publish("gutter", 8, "number", "Gap Between Widgets",null,{tags:["Basic"]});
+    
+    MegaChart.prototype.publish("chart",null,"widget","Chart widget to display in 'center' of MegaChart", null, {tags:["Basic"]});
+
+    MegaChart.prototype.enter = function (domNode, element) {
+        this._layout = new Border();
+        
+        this._layout.setContent("center",this.chart());
+        if(this.showLegend()){
```
GordonSmith: Needs to be in update

src/composite/MegaChart.js
```diff
@@ -0,0 +1,66 @@
...
+            this._layout.setContent("right", new Legend().targetWidget(this.chart()));
+        }
+        if(this.xAxisTitle()){
```
GordonSmith: Needs to be in update

src/composite/MegaChart.js
```diff
@@ -0,0 +1,66 @@
...
+            var xAxisTitleText = new Text().text(this.xAxisTitle());
+            this._layout.setContent("bottom", xAxisTitleText);
+        }
+        if(this.yAxisTitle()){
```
GordonSmith: Needs to be in update

GordonSmith:
One more comment:
I think I would create the X/YAxis title widgets and the legend widget in the enter method and then just add/remove them in the update method if showLegend === true and if the XYAxis text !== "" 
As it is, there will be issues when the marshaller starts to use them (and with Persist).

Also if you do it that way, you can add proxyPublish calls for X/YAxis titles.

src/common/Text.js
jbrundage: You're right... in its current state this offset needs some sort of trigonometric equation to determine its xOffset and yOffset... but that assumes that its parent isn't rotated? I'm not sure... maybe this implementation needs to be refactored?

mlzummo: XY Axis label rotation math should be applied here ?

jbrundage: Similar math, but both x and y taken into consideration this time

src/common/SVGWidget.js
```diff
@@ -131,7 +137,7 @@
     SVGWidget.prototype.postUpdate = function (domeNode, element) {
         Widget.prototype.postUpdate.apply(this, arguments);
         if (this._drawStartPos === "origin" && this._target instanceof SVGElement) {
-            this._element.attr("transform", "translate(" + (this._pos.x - this._size.width / 2) + "," + (this._pos.y - this._size.height / 2) + ")scale(" + this._scale + ")");
+            this._element.attr("transform", "translate(" + (this._pos.x - this._size.width / 2) + "," + (this._pos.y - this._size.height / 2) + ")");
```
GordonSmith: Why are you removing Scale from here?

src/common/SVGWidget.js
```diff
@@ -120,6 +120,12 @@
         return this;
     };
 
+    SVGWidget.prototype.rotation = function (_) {
+        if (!arguments.length) return this._rotation;
+        this._rotation = _;
+        return this;
+    };
+
```
GordonSmith: If rotation is not applied within Widget then the setter/getter should not be declared here (IOW it should be in Text.js)

src/common/Text.js
```diff
@@ -59,11 +62,18 @@
...
+        var r = xShift + yShift;
+        var theta = -context._rotation * Math.PI/180;
+        
+        xOffset = Math.round(r * Math.sin(theta));
+        yOffset = Math.round(r * Math.cos(theta));
+        
         this._textElement
             .style("text-anchor", this.anchor())
-            .attr("transform", function (d) { return "translate(" + xOffset + "," + yOffset + ")"; })
+            .attr("transform", function (d) { return "translate(" + xOffset + "," + yOffset + ")scale(" + context._scale + ")rotate(" + context._rotation + ")"; })
```
GordonSmith: I am not sure Scale should be here.

src/other/Table.js
jbrundage: I don't follow what you're suggesting... removing the width + height styles of this.element()?

The goal of lines 140 and 141 are to grab the width and height of the <table> and apply them to the target's <div>.

GordonSmith:
Looks good - @jbrundage Will you open a new issue to add button(s) to the Surface title to show/hide the legend?

GordonSmith:
@jbrundage Also open an issue to add composite to the builds + non amd test pages.

PR #1041, opened 2015-10-14
Allow Http/Https dyanmic switching of loading gmap

mlzummo:

### DIFF
--- a/src/map/GMap.js
+++ b/src/map/GMap.js
@@ -1,7 +1,8 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "../common/HTMLWidget", "../layout/AbsoluteSurface", "async!http://maps.google.com/maps/api/js?sensor=false", "css!./GMap"], factory);
+        var protocol = window.location.protocol !== "https" ? "http://" : "https://";
+        define(["d3", "../common/HTMLWidget", "../layout/AbsoluteSurface", "async!" + protocol + "maps.google.com/maps/api/js?sensor=false", "css!./GMap"], factory);
     } else {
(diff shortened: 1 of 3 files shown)


src/map/GMap.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "../common/HTMLWidget", "../layout/AbsoluteSurface", "async!http://maps.google.com/maps/api/js?sensor=false", "css!./GMap"], factory);
```
GordonSmith: Did search for anywhere else we may be making this mistake?

src/map/GMap.js
```diff
@@ -29,7 +29,7 @@
         this._prevMin = { x: 0, y: 0 };
         this._prevMax = { x: 0, y: 0 };
     }
-    Overlay.prototype = new google.maps.OverlayView();
+    Overlay.prototype = new google.maps.OverlayView() || null;
```
GordonSmith: ```|| null``` makes no sense...

mlzummo: I was suppose to remove that mod ... I'm trying to figure out how to fail
gracefully

mlzummo:
seems like the unit tests didnt like this too much ...

src/map/GMap.js
GordonSmith: @mlzummo does "async!//maps.google.com/maps/api/js?sensor=false" work?

mlzummo: it will fail in unit tests and when running locally if you do it with \\

GordonSmith: Ok defaulting to ```file:```  I suspect?

mlzummo:
@GordonSmith this is fixed ... can you pull?

Issue #962, opened 2015-09-25
New user feedback on long responding charts


arjunachala:
It will be nice to have some visual cue that the chart is being rendered. Sometimes you click on a chart and the child chart takes a few seconds to respond. In the meantime there is no visual cue to the user that the chart is in the process of being rendered.

mlzummo:
@arjunachala @GordonSmith  there are 2 functions that are overwritable widget.preUpdate and widget.postUpdate as well as the render(function(){}) callback that can be utilized for testing/running code when a widget is about to render and when its finished This can be handled outside the library I am not sure on Gordon's stance on something built into the library right now but DSP could use the Loading animation I've seen and plug it in.

dtsnell4:
@GordonSmith Do you want this to be its own widget, or just build it into Widget.js to run by default?

dtsnell4:
@GordonSmith rather common/Utility or api/ILoader of some sort?

mlzummo:
.on("progress   on("timeout on("error    ...

GordonSmith:
@dtsnell4 I think for any widget we should be able to show some sort of spinning "busy" image over the widget.  The following methods would be useful:
```
showBusy(boolean showHide)
pushBusy()
popBusy()
```
pushBusy would increment an internal busyCounter and show the busy image
popBusy would decrement an internal busyCounter and hide the busy image when that count hits 0

mlzummo:
i honestly don't think this belongs in the lib, and @GordonSmith you've made it clear in the past that things like this might be best handled outside the library .. also rendering widgets is not what takes time (almost every case of long loading is due to marshaller/ or dashboard as a whole not indv. widgets or even the Hipie service itself taking time to run the query) ... also this universal spinning widget might not match the style they want in which case i can bet (becuase its already being done) that they will role there own spinners which will allow much more consistency through out there apps ... i think the focus needs to be on getting progess/error and timeout handling in the marshaller .... the rest is superfluous

mlzummo:
i think this might be better served with an .on("progress", function() { (this would also pass the time that has lapsed) event system ... we have been talking about better event handling yet we avoid it at every step .... and i also think that this needs to be integrated into the marshaller which is where the focus needs to be atleast DSP (progrress/error and timeout handling) ... we really need to have a talk about why we might want to do this over that or vise versa

mlzummo:
@GordonSmith what would determine when to increment the counter vs decrement it?

[...]

PR #1052, opened 2015-10-15
GH-962 Implement "Busy" Hooks

dtsnell4:
Adds methods for starting and stopping a developer-defined "busy" indicator during long operations.

Fixes GH-962

### DIFF
--- a/src/common/Widget.js
+++ b/src/common/Widget.js
@@ -63,6 +63,7 @@
         if (window.__hpcc_theme) {
             this.applyTheme(window.__hpcc_theme);
         }
+        this._busyCounter = 0;
     }
     Widget.prototype._class = "common_Widget";
 
@@ -578,6 +579,25 @@
     };
 
     //  DOM/SVG Node Helpers  ---
+    
+    Widget.prototype.showBusy = function () {
+        this._busyIndicator.showHide(this._busyCounter);
+    };
+
+    Widget.prototype.pushBusy = function () {
+        this._busyCounter++;
+        if (this._busyCounter > 0) {
+            this.showBusy();
+        }
+    };
+
+    Widget.prototype.popBusy = function () {
+        this._busyCounter--;
+        if (this._busyCounter <= 0) {
+            this.showBusy();
+        }
+    };
+
     Widget.prototype.toWidget = function (domNode) {
         if (!domNode) {
             return null;
--- a/apps/SampleSite/test1.html
+++ b/apps/SampleSite/test1.html
@@ -1,0 +1,100 @@
+﻿<!doctype html>
+<html>
+<head>
+    <title>HPCC Systems - Popup Widget Examples</title>
+    <meta charset="utf-8">
+    <link rel="shortcut icon" href="favicon.ico" type="image/x-icon">
+    <link rel="icon" href="favicon.ico" type="image/x-icon">
+    <link href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.4/css/bootstrap.min.css" rel="stylesheet" type="text/css">
+    <link href="//cdnjs.cloudflare.com/ajax/libs/font-awesome/4.3.0/css/font-awesome.min.css" rel="stylesheet" type="text/css">
+    <link href="css/bootstrap_theme.css" rel="stylesheet" type="text/css">
+    <link href="css/prettify.css" rel="stylesheet" type="text/css">
+    <link href="//fonts.googleapis.com/css?family=PT+Sans:400,700" rel="stylesheet" type="text/css">
+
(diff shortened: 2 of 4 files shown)


dtsnell4:
@GordonSmith @mlzummo  Is there anything beyond this that needs to be done now?

src/common/Widget.js
```diff
@@ -578,6 +579,29 @@
     };
 
     //  DOM/SVG Node Helpers  ---
+    
+    Widget.prototype.showBusy = function (show) {
+        if (show) {
+            //show the indicator
+        } else {
+            //hide the indicator
```
dtsnell4: Do we want any default code here?

GordonSmith: Yup - this is where it gets subjective - FWIW here is the loading gif I use in ECL Watch:  https://github.com/GordonSmith/HPCC-Platform/blob/master/esp/src/eclwatch/img/loadingBar.gif

But I also like the way feedly.com does its progress (at the top of the screen):  http://feedly.com/i/index

src/common/Widget.js
```diff
@@ -578,6 +579,29 @@
...
+        }
+    };
+
+    Widget.prototype.pushBusy = function () {
+        this._busyCounter++;
+        if (this._busyCounter > 0) {
```
GordonSmith: ```this._busyCounter > 0``` will keep the showBusy code simpler which is good as it may be overriden by someone

src/common/Widget.js
```diff
@@ -578,6 +579,29 @@
...
+            this.showBusy(true);
+        }
+    };
+
+    Widget.prototype.popBusy = function () {
+        this._busyCounter--;
+        if (this._busyCounter <= 0) {
```
GordonSmith: ```this._busyCounter === 0``` will keep the showBusy code simpler which is good as it may be overriden by someone

dtsnell4: The only reason I did this is if there is a mistake by the developer and busyCounter gets <0 the loader will get stuck on the screen.  With <= it will still be removed.  I've seen similar things happen in other projects I've worked. Let me know which is the lesser of the two evils.

src/common/Widget.js
dtsnell4: @mlzummo @jbrundage Here is my .02¢ on the subject:

We could make a Loader widget that utilizes the built in FontAwesome spinner icons https://fortawesome.github.io/Font-Awesome/examples/ (scroll down to "Animated Icons").  With this approach we simply utlilize the FaCar widget, its all code, there are no extra files to worry about and track.  There are several different options and speed/size etc. can all be configured.  We can allow user to over ride with their own image if they choose to.

src/common/Widget.js
mlzummo: i have to agree with dan on this

src/common/Widget.js
mlzummo: sounds good ... it seems this project is moving towards also being a UI library ....

src/common/Widget.js
jbrundage: I agree, "<=" can prevent headaches

src/common/Widget.js
jbrundage: FontAwesome loading icon would be an ideal default

src/common/Widget.js
GordonSmith: Ok - the onus will be on the showBusy function not to create two busy images so.

GordonSmith:
@dtsnell4 we still need a basic busy implementation.

dtsnell4:
Yes, see https://github.com/hpcc-systems/Visualization/pull/1052#discussion_r42238627

src/common/Widget.js
GordonSmith: Agreed

dtsnell4:

I finished the Busy widget, it was a bit more complicated than I expected.  Using the FaChar widget prevents us from putting the code directly in Widget/SVGWidget/HTMLWidget.js so I created a stand alone widget for this.  You can play with the publish parameters here https://rawgit.com/dtsnell4/Visualization/GH-962_Busy_Indicator/demos/dermatology.html?src/common/Busy  just increment the busyCounter.

I also put together a quick test page here https://rawgit.com/dtsnell4/Visualization/GH-962_Busy_Indicator/apps/SampleSite/test1.html  I just used some timeouts to simulate the widget being busy.

mlzummo:
@dtsnell4 how come HTMLWidget/SVGWidget isnt consuming this and it being all handled within there?

dtsnell4:
You can't fire up another widget that uses them from within them (in this case FaChar & Text)

mlzummo:
look at not extending HTML/SVGWidget for the busy widget

Issue #835, opened 2015-09-01
Axis Tick Formatting for Google Charts


GordonSmith:

mlzummo:
@GordonSmith @dtsnell4 not possible ... yAxisFormat is already exposed and is limited to what google allows.

PR #1053, opened 2015-10-16
GH-835 Google Axis Ticks Formatting

dtsnell4:
Changes the publish param for axis formatting from "string" to "set".

Fixes GH-835

### DIFF
--- a/src/google/Area.js
+++ b/src/google/Area.js
@@ -54,8 +54,8 @@
     Area.prototype.publish("xAxisInversed", false, "boolean", "The Direction In Which The Values Along The Horizontal Axis Grow.",null,{tags:["Advanced"]});
     Area.prototype.publish("yAxisInversed", false, "boolean", "The Direction In Which The Values Along The Vertical Axis Grow.",null,{tags:["Advanced"]});
 
-    Area.prototype.publish("xAxisFormat", "", "string", "Format String For Numeric Axis Labels", ["","decimal","scientific","currency","percent","short","long"],{tags:["Intermediate"]});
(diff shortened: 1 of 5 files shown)


dtsnell4:
@GordonSmith @jbrundage @mlzummo Matt is correct, Google does not support d3.format, however it does have its own formatting options.  I noticed however the publish param was set to "string" instead of "set", so I fixed that.  Should it be a different issue?

dtsnell4:
Although, we could make the publish param set the equivalent d3.format specifiers, and just translate them in the code to Google's format.  If the objective is to keep the interface pure d3 for the devs.

mlzummo:
the problem with that is d3 supports alot more than what google offers in the formatting area prolly?

jbrundage:
The publish parameters xAxisFormat and yAxisFormat should be xAxisTickFormat and yAxisTickFormat to match the other chart libraries

dtsnell4:
@mlzummo I agree, which is why I suggested making it a set, so they only have the corresponding options.  Probably irrelevant, just throwing it out there.  @jbrundage I will change the names.

jbrundage:
On second thought, since those two parameters are sets of predefined format types it may be a better idea to name them something along the lines of xAxisFormatType and yAxisFormatType.

dtsnell4:
@jbrundage ok I'll wait...  @GordonSmith @mlzummo your thoughts?

mlzummo:
don't name them the same as the rest of the charts becuase that will cause confusion when switching charts and them wondering why its not working ....

GordonSmith:
@mlzummo @jbrundage Please re-review

mlzummo:
looks good

PR #1062, opened 2015-10-19
GH-1058 Adds Show/Hide Legend to MegaChart

jbrundage:
Fixes GH-1058

@GordonSmith please review

http://rawgit.com/jbrundage/Visualization/GH-1058-MC-ToggleLegend/demos/dermatology.html?src/composite/MegaChart

### DIFF
--- a/src/composite/MegaChart.js
+++ b/src/composite/MegaChart.js
@@ -38,6 +38,12 @@
     
     MegaChart.prototype.publish("chartType", "LINE", "set", "Chart Type", _chartTypes.map(function (item) { return item.id; }),{tags:["Basic"]});
     
+    MegaChart.prototype.click = function(obj) {
+        if(obj.id === this.id()+"_toggle_legend"){
+            this.showLegend(!this.showLegend()).render();
+        }
+    };
+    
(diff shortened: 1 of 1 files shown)


GordonSmith:
Clicking the icon hides the titles as well as the legend...

GordonSmith:
@jbrundage Needs rebase (and see above comment)

jbrundage:
@GordonSmith After some investigation here's what I've found:

Any call to MegaChart.prototype.update causes the same issue (the titles vanish).

The Cell widgets within Border are being recreated on each call to MegaChart.prototype.update. But the X/YAxis labels Text() widgets remain the same **except they're losing their _target elements** so they are never displayed.

Any idea why the Border Cell widgets aren't updating the Text _target when the Cell is recreated?

src/composite/MegaChart.js
```diff
@@ -86,7 +110,7 @@
         } else {
             this._layout.clearContent("bottom");
         }
-        
+        this._layout.render();
```
GordonSmith: Does this need a test for null?  You have one above...

PR #1085, opened 2015-10-22
Amchart axis refactor (proof of concept)

mlzummo:


mlzummo:
serilaization works but running render() directly after an xAxis PP mod does not re-render chart but setting something like 3ddepth does and then it correcylu reads the param and sets it ... (could be context issue)

mlzummo:
@GordonSmith I know this is a mess right now, im going to clean it up ... but i'd like to know your thoughts on what i did here ... so far everything is working great with it and i would like to roll it out accross the board ... i just have a few questions on it

src/amchart/CommonSerial.js
```diff
@@ -126,8 +82,77 @@
...
+            var axis = new Axis();
+            axis._context = this;
+            var currentAxes = CommonSerial.prototype._origXAxis.call(this);
+            currentAxes.push(axis);
+            CommonSerial.prototype._origXAxis.call(this, currentAxes)
+        } 
+        return axis;
+    };
+
+    //CommonSerial.prototype._origYAxis = CommonSerial.prototype.yAxis;
```
mlzummo: a cleaner version of above (depending on a few things i will remove the prototype calls)

src/amchart/CommonSerial.js
```diff
@@ -126,8 +82,77 @@
...
+            var axes = CommonSerial.prototype._origXAxis.call(this);
+            axes.forEach(function(axis) {
+                axis._context = context;
+            });
+            return axes;
+            //return  // might need to override _context b4 passing ...  but so far dont need to
+        }
+        if (CommonSerial.prototype._origXAxis.call(this)[_]) {
+             var axis = CommonSerial.prototype._origXAxis.call(this)[_]
+             axis._context = this;
```
mlzummo: after deserialization you must overwrite the context again with the correct one

src/amchart/CommonSerial.js
```diff
@@ -260,27 +285,27 @@
...
-        this._chart.valueAxes[0].axisColor = this.yAxisBaselineColor();
-        this._chart.valueAxes[0].axisAlpha = this.axisAlpha();
-        this._chart.valueAxes[0].fillColor = this.yAxisFillColor();
-        this._chart.valueAxes[0].fillAlpha = this.yAxisFillAlpha();
-
-        this._chart.valueAxes[0].gridAlpha = this.yAxisGridAlpha();
-        this._chart.valueAxes[0].dashLength = this.yAxisDashLength();
-        this._chart.valueAxes[0].boldPeriodBeginning = this.yAxisBoldPeriodBeginning();
-        this._chart.valueAxes[0].axisTitleOffset = this.yAxisTitleOffset();
+        this._chart.valueAxes[0].title = this.yAxis()[0].yAxisTitle();
```
mlzummo: eventually loop through the axis to set the values .... axis can be delete like so : this.xAxis().slice(1,1);

src/common/Widget.js
```diff
@@ -246,7 +247,8 @@
                     this["__prop_" + id] = _;
                 }
             }
-            return this;
+
+            return this._context ? this._context : this;
```
mlzummo: this is the magic that makes it all possible

src/other/Persist.js
```diff
@@ -233,7 +235,14 @@
                             });
                             break;
                         default:
+                            console.log(item.id)
```
mlzummo: random debug code

mlzummo:
can't override the WidgetArray param like the others and still have it work ... wondering how to fix that

mlzummo:
usage:    widget.xAxis(0).fontSize(10).render();   it can also be read via .xAxis()[0] ... this will also for amcharts atleast let me do the individual graph options as well for the graph object (which is similiar to rendering diff charts on an XYAxis concept you have for HPCC Charts)

mlzummo:
@GordonSmith bump

Issue #1090, opened 2015-10-26
Table widget fixedSize param causes problems in Dermatology


dtsnell4:
http://rawgit.com/hpcc-systems/Visualization/master/demos/dermatology.html?src/other/Table

In Dermatology and Playground, when the Table widget first updates, the width and height are initially 0 which causes the table not to be displayed.

Issue #890, opened 2015-09-08
Pagination + Narrow Space


GordonSmith:
We need some scheme where the pagination display is prevented from wrapping to a new line:
![image](https://cloud.githubusercontent.com/assets/966863/9730533/142cd5e8-560f-11e5-8392-599d6f3646d0.png)

Maybe a narrow mode which would look like:
[<<][<][[1] of 20][>][>>]

(the main issue with the wrapping is it affects the pagination calulation thus changing the page of number and is confusing to use).

mlzummo:
@GordonSmith what about using a nowrap/css to make sure that the paginator does not wrap?

GordonSmith:
I briefly tried that and failed miserably - but if it fixes it, its good with me.

mlzummo:
also dont allow negative total page number count

dtsnell4:
I'll work on this

PR #1091, opened 2015-10-26
GH-890 Pagination + Narrow Space

dtsnell4:
Adds CSS so that the paginator does not wrap to another line.
Also added checks for fixedSize() param so the size is not set to 0.

Fixes GH-890
Fixes GH-1090

### DIFF
--- a/src/other/Paginator.css
+++ b/src/other/Paginator.css
@@ -1,16 +1,12 @@
 .other_Paginator {
   display: block;
   position:absolute;
-  bottom: 20px;
-  right: 20px;
+  white-space: nowrap;
 }
 
 .other_Paginator .pagination {
   display: inline-block;
-  padding-left: 0;
-  margin: 0 0;
-  border-radius: 4px;
-  float: left;
+  white-space: nowrap;
 }
 
 .other_Paginator .pagination > li {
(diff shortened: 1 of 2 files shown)


dtsnell4:
@GordonSmith @jbrundage @mlzummo Please review

This adds css that prevents the paginator from wrapping.  

At that point I believe it should be up to the developer, if necessary, to add their own css for font, padding, etc. to make the paginator itself smaller.  I got it to look like like with a few changes:

paginator

Additionally, there are really two paginators there.  The number input in the second one has a spin button to scroll through the page numbers, its really all that is necessary.  Dev can just hide the first one with css.

dtsnell4:
@GordonSmith @mlzummo @jbrundage Don't bother reviewing, paginator stopped working, need to fix.

dtsnell4:
@GordonSmith @mlzummo @jbrundage Fixed, please review (note above comment)

dtsnell4:
http://rawgit.com/dtsnell4/Visualization/GH-890_Table_Narrow_Pagination/demos/dermatology.html?src/other/Table

GordonSmith:
@mlzummo Please Review

mlzummo:
looks good

Issue #1101, opened 2015-10-28
Table column totals


dtsnell4:
Add the option to total a column of numbers at the bottom of the table.

PR #1103, opened 2015-10-28
GH-1101 Table column totals

dtsnell4:
Adds params to sum total the contents of a numeric column.

Fixes GH-1101

### DIFF
--- a/src/other/Table.css
+++ b/src/other/Table.css
@@ -85,3 +85,9 @@
     white-space: nowrap;
     box-sizing:border-box;
 }
+
+.other_Table tfoot td, .rows-wrapper tfoot td {
+    background-color: #ADDFF3;
+    font-weight: bold;
+    border-top-width: 3px
+}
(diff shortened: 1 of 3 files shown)


dtsnell4:

http://rawgit.com/dtsnell4/Visualization/GH-1101_Table_Column_Totals/demos/dermatology.html?src/other/Table.totalled

mlzummo:
looks good

GordonSmith:
1 - I think visually the totals are not obviously totals enough - they need to look more different - maybe they should not have a grid around them?
2 - They only total the current page when pagination is on (is this a feature or an issue?)

dtsnell4:
1.  I would expect the user to be adding their own CSS to differentiate it, but I can certainly add a default look.
2.  After speaking with Matt and DSP, from what we gathered it is a feature.

dtsnell4:
@GordonSmith updated based on comments, with the exception of #1 above, as I understand it, it is correct.

Issue #1113, opened 2015-10-29
Refactor column + data


GordonSmith:
Create a new "database" or "grid" object which can encapsulate the current data + column methods.

This new object would then be _the_ place for setting things like formatters, pattern masks, rollups etc.

PR #1114, opened 2015-10-29
GH-1113 Refactor widget.data/column

GordonSmith:
Add Database File with a Grid Class.

Fixes GH-1113


GordonSmith:
@jbrundage @mlzummo @dtsnell4 Please Review
The new DB is used in every widget now, but here is my main test page:  http://rawgit.com/GordonSmith/Visualization/GH-1113/demos/DataAnalyst/index.html

src/chart/Column.js
```diff
@@ -92,7 +92,6 @@
 
                 if (isHorizontal) {
                     columnRect.transition()
-                        .attr("class", "columnRect")
```
mlzummo: why did you remove the class?

src/common/Database.js
```diff
@@ -0,0 +1,436 @@
...
+    if (typeof define === "function" && define.amd) {
+        define(["d3"], factory);
+    } else {
+        root.common_Database = factory(root.d3);
+    }
+}(this, function (d3) {
+    function Grid() {
+        this.clear();
+    }
+    Grid.prototype.constructor = Grid;
```
mlzummo: this isnt in reference to the layout grid we have ... should we use the term DataMatrix? (this is my first pass through of this code so my questions might not make sense)

src/common/Surface.js
```diff
@@ -264,7 +264,9 @@
             ;
         }
 
-        this._menu.element().node().parentNode.appendChild(this._menu.element().node()); // Make sure menu is on top (Z-Order POV)
+        if (this._menu.element() && this._menu.element().node() && this._menu.element().node().parentNode) {
```
mlzummo: this is to just check th menu exists and is a child before doing the Z order fix?

src/common/Database.js
```diff
@@ -0,0 +1,436 @@
...
+                    case "SUM":
+                        return d3.sum(leaves, function (d) { return d[rollupValueIdx[0]]; });
+                    case "AVE":
+                        return d3.mean(leaves, function (d) { return d[rollupValueIdx[0]]; });
+                    case "MIN":
+                        return d3.min(leaves, function (d) { return d[rollupValueIdx[0]]; });
+                    case "MAX":
+                        return d3.max(leaves, function (d) { return d[rollupValueIdx[0]]; });
+                }
+                return "???";
```
mlzummo: ??? .. whats that?

GordonSmith: Should be console warning I suspect.

src/common/Database.js
```diff
@@ -0,0 +1,436 @@
...
+            var mapping = columns[rollupField];
+            var params = [];
+            for (var param in mapping.params) {
+                params.push(mapping.params[param]);
+            }
+            var nested = this.rollup(rollupBy, function (leaves) {
+                switch (mapping.function) {
+                    case "SUM":
+                        return d3.sum(leaves, function (d) { return d[rollupValueIdx[0]]; });
+                    case "AVE":
```
dtsnell4: Should this be "AVG"?

GordonSmith: Alas no (and MEAN would have been my preference) - its from the DDL.

src/common/Surface.js
GordonSmith: Yea I hot a scenario where there was no menu and just wrapped it in the if.

src/common/Database.js
GordonSmith: You will only ever see in this context:  ```var dataGrid = new Database.Grid();``` so I think its ok.

src/chart/Column.js
GordonSmith: Its set on the enter() method above so is not needed.

dtsnell4:
This all looks good to me.  Looking forward to trying it out.

jbrundage:
Yes, looks good to me too

mlzummo:
im ok with merging it

GordonSmith:
Fixed the "???" Issue

Issue #1116, opened 2015-10-29
integrate mouseover - highlight chart indication on filter/click events - into marshaller


mlzummo:
when a widget updates another widget when you mouse over it should indicate which charts it updates. the code is there to support this and needs to be integrated into marshaller

GordonSmith:
Merged

PR #1121, opened 2015-10-30
GH-1116 Highlight Update Targets

GordonSmith:
HTML Marshaller will highlight which widgets update other widgets.
Fixed bug in Grid.getCell (broken when cell density was introduced).
Partial fix for performance issue (opening new issue(s) for my thoughts).

### DIFF
--- a/src/layout/Grid.js
+++ b/src/layout/Grid.js
@@ -100,6 +100,18 @@
         return retVal;
     };
 
+    Grid.prototype.getWidgetCell = function (id) {
+        var retVal = null;
+        this.content().some(function (cell) {
+            if (cell.widget()._id === id) {
+                retVal = cell;
+                return true;
+            }
+            return false;
+        });
+        return retVal;
+    };
+
     Grid.prototype.getContent = function (id) {
         var retVal = null;
         this.content().some(function (cell) {
--- a/src/common/Widget.js
+++ b/src/common/Widget.js
@@ -344,7 +344,7 @@
     };
 
     Widget.prototype.broadcast = function (key, newVal, oldVal) {
-        if (this._watchArr) {
+        if (this._watchArr && newVal !== oldVal) {
             this._watchArr.forEach(function (func) {
                 if (func) {
                     setTimeout(function () {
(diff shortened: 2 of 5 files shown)


GordonSmith:
@jbrundage Please Review

mlzummo:
@GordonSmith do you have a link to show this working?

src/layout/Grid.js
```diff
@@ -87,7 +87,9 @@
         return this;
     };
 
-    Grid.prototype.getCell = function (row, col) {
+    Grid.prototype.getCell = function (_row, _col) {
+        var row = _row * this.cellDensity();
+        var col = _col * this.cellDensity();
```
jbrundage: @GordonSmith This might cause undesired effects... this assumes that _row and _col aren't already adjusted for cellDensity.

jbrundage: ...but everything seems to work normally

jbrundage: @GordonSmith Actually this does break the ability to drop a widget ontop of another widget to swap their locations in designMode.

When the cellDensity is changed it is automatically applied to each cell's gridRow, gridCol, gridRowSpan and gridColSpan. Lines 91 and 92 shouldn't be necessary.

jbrundage: ...it also prevents the user from being able to "select" a cell

GordonSmith: That means that setContent(5, 5, widget) is not symmetrical with a getCell(5, 5)?

jbrundage: Valid point, yes I should have considered getCell might be used outside of the layout/Grid file.

I suppose a new method **getDensityCell()** should be created/used inside the layout/Grid file and getCell should be as you have it in this PR.

GordonSmith: For now I will add a getWidgetCell = function(Widget) method.

GordonSmith:
http://rawgit.com/GordonSmith/Visualization/GH-1116/demos/dermatology.html?src/marshaller/HTML

@jbrundage Changes made.

GordonSmith:
REbased @jbrundage Please Review

mlzummo:
looks good

Issue #1123, opened 2015-10-30
Table Widget - allow for size of 100%


dtsnell4:
For MegaChart/Border Widget shrinkWrap feature.  Previous the table would only accept dimensions in pixels.

PR #1124, opened 2015-10-30
GH-1123 Table Widget - allow for size of 100%

dtsnell4:
For MegaChart/Border Widget shrinkWrap feature. Previous the table would only accept dimensions in pixels. 

Fixes GH-1123

### DIFF
--- a/src/other/Table.js
+++ b/src/other/Table.js
@@ -136,12 +136,6 @@
     Table.prototype.update = function (domNode, element) {
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
-        var box = d3.select(".tableDiv > table").node().getBoundingClientRect();
-        if(this.fixedSize() && box.width !== 0 && box.height !== 0){
-            element.attr({width: box.width + "px",height: box.height + "px"});
-        } else {
-            element.attr({width: "100%",height: "100%"});
-        }
         
         if (!this.showHeader()) {
             this.fixedHeader(false);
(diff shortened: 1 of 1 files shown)


dtsnell4:
@GordonSmith @jbrundage @mlzummo please review.

Also if this should target 1.10 just let me know

mlzummo:
keep it at 1.8.2

src/other/Table.js
```diff
@@ -30,7 +30,7 @@
     Table.prototype.publish("fixedHeader", false, "boolean", "Enable or disable fixed table header",null,{tags:["Private"]});
     Table.prototype.publish("fixedColumn", false, "boolean", "Enable or disable fixed first column",null,{tags:["Private"]});
     
-    Table.prototype.publish("fixedSize", false, "boolean", "Fix Size to Min Width/Height");
+    Table.prototype.publish("fixedSize", true, "boolean", "Fix Size to Min Width/Height");
```
GordonSmith: To match the existing behavior is this the correct setting?

dtsnell4: This is what @jbrundage had it set to when he introduced it, I had changed it when I originally tried to fix this, but it broke it the table/legend in the Megachart.  I'm not sure what the default should be, or if it really makes a difference. Do you want me to change it? Any thoughts Jay?

jbrundage: It should default **false** (allowing 100% width and 100% height by default). The **tableDiv** is what this publish param ultimately controls, which should usually adapt to its parent's dimensions.

dtsnell4: OK, then in the MegaChart it needs to be set to true for the legend to be visible.

dtsnell4:
@GordonSmith updated to false

jbrundage:
Looks good

Issue #1131, opened 2015-11-02
AM Charts Click Events Missing Selection


GordonSmith:
AM Charts are not passing in true/false for the click event to indicate if the click resulted in a selection or a deselection.

PR #1132, opened 2015-11-02
GH-1131 AM Charts Click Missing Selection Flag

GordonSmith:
Pass true/false based on selection / de-selection.
Improved Surface Titles (params).
Fixed resize issue with marhaller demo page.
Added "endsWith" to Utility.js

Fixes GH-1131

### DIFF
--- a/demos/marshaller/index.html
+++ b/demos/marshaller/index.html
@@ -1,4 +1,4 @@
-<!doctype html>
+﻿<!doctype html>
 <html>
 <head>
     <meta charset="utf-8">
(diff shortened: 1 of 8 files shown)


GordonSmith:
@mlzummo Please Review

mlzummo:
i don't believe hipiie supports multi section yet? so this all looks good ... how will we support multi with hipie in the function since amcharts does support it?

GordonSmith:
Mutli-selection is only a possibility in the future - once HIPIE does support it then I would imagine passing in an array of "selected" items rather than a single item will work?

Issue #1144, opened 2015-11-03
amcharts lose selection on resize


mlzummo:

mlzummo:
@GordonSmith loos like even with the coloring its still an issue will have to reapply selection after an update event (i dont think there is a manual way of calling slide out on a pie chart without firing the event again)

GordonSmith:
@mlzummo I know I have to do that in the hpcc charts (recolor selection on update).

mlzummo:
fixed as of 1.10

Gor:
@mlzummo Gor its me man. Stop writing my nickname in your discussions please )

PR #1151, opened 2015-11-04
GH-1144 Amchart select persist 

mlzummo:
@GordonSmith Please Review ... if your ok with how i did this ill apply to the rest of amcharts 

Fixes GH-1144

### DIFF
--- a/src/amchart/Pie.js
+++ b/src/amchart/Pie.js
@@ -1,17 +1,20 @@
 "use strict";
 (function(root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "../common/HTMLWidget", "amcharts.pie", "../api/I2DChart", "require"], factory);
+        define(["d3", "../common/HTMLWidget", "amcharts.pie", "../api/I2DChart", "require", "../common/Utility"], factory);
     } else {
-        root.amchart_Pie = factory(root.d3, root.common_HTMLWidget, root.AmCharts, root.api_I2DChart, root.require);
+        root.amchart_Pie = factory(root.d3, root.common_HTMLWidget, root.AmCharts, root.api_I2DChart, root.require, root.common_Utility);
     }
-}(this, function(d3, HTMLWidget, AmCharts, I2DChart, require) {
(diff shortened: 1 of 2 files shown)


src/amchart/Pie.js
```diff
@@ -154,7 +167,13 @@
                     context._selected = {
                         field: field,
                         data: data
-                    };
+                    };
```
mlzummo: remove this block

src/amchart/Pie.js
```diff
@@ -85,13 +88,20 @@
...
         } else {
-        	sortingMethod = function(a,b){ return a[1] > b[1] ? 1 : -1; };
+            sortingMethod = function(a,b){ return a[1] > b[1] ? 1 : -1; };
         }
         this.data(this.data().sort(sortingMethod));
 
         this._chart.colorField = "sliceColor";
-
-        this._chart.dataProvider = this.formatData(this.data());
+
```
mlzummo: do we want updates to persist after data has been changed (added or removed)

mlzummo:
@GordonSmith Can you comment on this?

GordonSmith:
Seems to be too much going on inside the setting of the color array.  I would suggest you need:
1.  A way to add/remove rows to this._selections, based on user interaction
2.  Inside the creation of the color map simply have a ```this._selections.indexOf(row) >= 0``` test.

mlzummo:
@GordonSmith we will need to run .update on every click then ... also indexOf does a strict comparison so unless its a direct ref it wont work

mlzummo:
running a rerender is going to be more expensive than computing a hash

GordonSmith:
1.  If one of the chart elements has changed color due to a mouse action then you have to re-render?
2.  I was assuming your going to store the original row and not a clone, so indexof should work fine, if not, or if you need to persist the selection between complete data re-loads (which is not unlikely) then storing a hash of the row would be ideal (we will want to make sure we pick a suitable hashing algo. though).

mlzummo:
@GordonSmith should selections persist even after a data update or should that invalidate a selection (if it invalidates it would make things so much easier)

mlzummo:
closing in favor of: GH-1201

PR #1152, opened 2015-11-05
Refactor Widget Base Class

GordonSmith:

### DIFF
--- a/src/common/PropertyExt.js
+++ b/src/common/PropertyExt.js
@@ -1,0 +1,263 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["d3"], factory);
+    } else {
+        root.common_PropertyExt = factory(root.d3);
+    }
+}(this, function (d3) {
+    var __meta_ = "__meta_";
+    var __prop_ = "__prop_";
+
+    function Meta(id, defaultValue, type, description, set, ext) {
+        this.id = id;
+        this.type = type;
+        this.origDefaultValue = defaultValue;
+        this.defaultValue = defaultValue;
+        this.description = description;
+        this.set = set;
+        this.ext = ext || {};
+
+        switch (type) {
+            case "set":
+                this.checkedAssign = function (_) {
+                    if (!set || set.indexOf(_) < 0) {
+                        console.log("Invalid value for '" + id + "':  " + _ + " expected " + type);
+                    }
+                    return _;
+                };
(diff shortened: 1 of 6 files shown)


GordonSmith:
@mlzummo @jbrundage @dtsnell4 Please review.  Just a refactor of the Widget base class to break up the different functional pieces into their own source files.
http://rawgit.com/GordonSmith/Visualization/PLANE_REFACTOR/test/runner_rawgit.html

mlzummo:
this is what we needed ... Axis can now extend PropertyExt

src/common/PropertyExt.js
```diff
@@ -0,0 +1,254 @@
...
+                    if (!(_ instanceof Object)) {
+                        console.log("Invalid value for '" + id + "':  " + _ + " expected " + type);
+                    }
+                    return _;
+                };
+                break;
+            case "widget":
+            case "widgetArray":
+                this.checkedAssign = function (_) {
+                    //  TODO  ---
```
mlzummo: ???

mlzummo: can we add one for propertyExt?

GordonSmith: The TODO is for me to add a test that _ === WidgetArray (I don't have a clean way yet)

mlzummo:
@GordonSmith can we add

GordonSmith:
@mlzummo Can you elaborate, I am not sure what your thinking of?

mlzummo:
@GordonSmith for the Axis obj that we are working on instead of using a whole widget .. with this refactor we would be able to just extends propertyExt class wouldnt we? and if thats the case then widgetarray wouldnt accept a propertyExy Obj we would need a new one and then be able to serialize that ...

src/common/PropertyExt.js
mlzummo: check that its a widget and that its an Instance of Widget?

mlzummo:
@GordonSmith ... when will this be finished and pulled?

GordonSmith:
Its waiting on a final review.

src/common/PropertyExt.js
GordonSmith: If I do that it will be in a __hpcc_debug section...

GordonSmith:
@mlzummo Added type checking for widget + WidgetArray (can't use instanceof due to circular references...)

mlzummo:
looks good

PR #1154, opened 2015-11-05
Adds AmCharts Cursor

mlzummo:


src/amchart/CommonSerial.js
```diff
@@ -369,7 +378,31 @@
         var initObj = {
             type: "serial",
             addClassNames: true,
-            chartScrollbar: {}
+            chartScrollbar: {},
+            autoMargins: true,
+            valueAxes: [
+                {position:"bottom",title:" "},
```
mlzummo: remove this if not needed .. and need to test if we can update after first render

mlzummo:
remove pie updated code doesnt belong

Issue #1176, opened 2015-11-09
Deselect a selected row in the Table Widget


dtsnell4:
Add code to deselect a selected row(s) in the table.

Issue #1175, opened 2015-11-09
Refactor Table Widget


dtsnell4:
Refactor Table Widget to remove all copying javascript for the fixed header and first column.  Make widget all d3js.

Issue #1140, opened 2015-11-03
Table widget default to paginate if more than X num of rows


jbrundage:
After speaking with Jo Prichard regarding Table widget performance issues. It would make sense to allow the Table to paginate if a table contains more than X num of rows.

DSP regularly returns very large tables resulting in performance issues. If we create a publish param containing the desired threshold for Table row count (beyond which the Table would default to paginate), this would help resolve performance issues.

GordonSmith:
I saw the same thing with the "DataAnalyst" and defaulted pagination to on for that project.

PR #1177, opened 2015-11-09
GH-1175 Refactor Table Widget

dtsnell4:
Refactors Table Widget to remove all copying javascript for the fixed header and first column. Makes widget all d3js.  Adds code to deselect a selected row(s) in the table.  Adds paginationLimit() publish parameter to automatically turn on pagination when a developer-defined number of rows is reached.

Fixes GH-1140
Fixes GH-1175
Fixes GH-1176

### DIFF
--- a/src/other/Table.js
+++ b/src/other/Table.js
@@ -23,11 +23,12 @@
 
     Table.prototype.publish("renderHtmlDataCells", false, "boolean", "enable or disable HTML within cells",null,{tags:["Private"]});
     Table.prototype.publish("pagination", false, "boolean", "Enable or disable pagination",null,{tags:["Private"]});
+    Table.prototype.publish("paginationLimit", null, "number", "Maximum number of rows allowed before pagination defaults to on",null,{tags:["Private"]});
     Table.prototype.publishProxy("itemsPerPage", "_paginator");
     Table.prototype.publishProxy("pageNumber", "_paginator", "pageNumber",1);
     Table.prototype.publishProxy("adjacentPages", "_paginator");
     Table.prototype.publish("showHeader", true, "boolean", "Show or hide the table header",null,{tags:["Private"]});
-    Table.prototype.publish("fixedHeader", false, "boolean", "Enable or disable fixed table header",null,{tags:["Private"]});
+    Table.prototype.publish("fixedHeader", true, "boolean", "Enable or disable fixed table header",null,{tags:["Private"]});
     Table.prototype.publish("fixedColumn", false, "boolean", "Enable or disable fixed first column",null,{tags:["Private"]});
     
     Table.prototype.publish("fixedSize", false, "boolean", "Fix Size to Min Width/Height");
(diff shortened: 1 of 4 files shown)


mlzummo:
@dtsnell4 can you provide a demo link?

dtsnell4:
@GordonSmith @mlzummo @jbrundage Please review.  This should be all d3 at this point, but if I am missing anything, please let me know.

http://rawgit.com/dtsnell4/Visualization/GH-1175_Refactor_Table_Widget/demos/dermatology.html?src/other/Table.totalled

src/other/Table.js
```diff
@@ -23,11 +23,12 @@
 
     Table.prototype.publish("renderHtmlDataCells", false, "boolean", "enable or disable HTML within cells",null,{tags:["Private"]});
     Table.prototype.publish("pagination", false, "boolean", "Enable or disable pagination",null,{tags:["Private"]});
+    Table.prototype.publish("paginationLimit", null, "number", "Maximum number of rows allowed before pagination defaults to on",null,{tags:["Private"]});
```
dtsnell4: @jbrundage this is in response to https://github.com/hpcc-systems/Visualization/issues/1140.  Based on your discussion with Jo, should this default to a certain number, or leave it up to the dev?

mlzummo: why don't we just default pagination to on in general?

mlzummo:
fixedSize toggling breaks the table (cant get access or scroll to the bottom rows)

dtsnell4:
That's what fixedSize does, it makes the container for the table the exact size as the table, thus no scroll bars.  This is really only intended for the Table in the Legend for the MegaChart to fix an issue there.

src/other/Table.js
jbrundage: Does the Table have the ability to force a specific number of rows per page? (I remember we decided to allow the paginator to strictly set the number of rows per page in order to allow room for the page controls at the bottom) 

If we can't set the rows-per-page specifically then I think paginationLimit is a better approach because it allows the table to grow reasonably large without forcing pagination.

dtsnell4: @jbrundage Interesting, there is a itemsPerPage() param but it doesn't appear to do anything in dermatology.  Will look into that.

I personally do not like the idea of defaulting to pagination... I have not seen a lot of use cases, but what I have seen suggests that most tables are displayed in rather small containers (e.g. Legend) and the pagination controls are just too clunky.

GordonSmith:
The fixed header is moving around on IE.
Also totals seem to be erratic:  
image

dtsnell4:
@GordonSmith I fixed the header not scrolling problem in IE, unfortunately it is VERY jumpy now.  http://rawgit.com/dtsnell4/Visualization/GH-1175_Refactor_Table_Widget/demos/dermatology.html?src/other/Table.totalled

This is a fairly common problem with certain elements in IE.  Apparently absolute positioned table heads are one of them.  It scrolled fine in IE in the last iteration when I was copying the head into a div, the div scrolled smoothly.  I was trying to keep the table head intact and just position that to keep the code cleaner, but i guess I will have revert back to the div to solve this issue...

dtsnell4:
@jbrundage Regarding the issue above https://github.com/hpcc-systems/Visualization/pull/1177#issuecomment-156383399

Regarding the totalledColumns() pp array.  This is an array of integers that tells the code which columns it needs to add up totals for.  If you inspect the console here http://rawgit.com/dtsnell4/Visualization/GH-1175_Refactor_Table_Widget/demos/dermatology.html?src/other/Table.totalled  You will see the array outputted.  

Now, if you go into the property editor and edit this array, you will notice in the console that it is changed to an array of strings.  That is why it is failing.  Is there any way to prevent this?  I see that it does not happen when you edit the data array.  Or should I just iterate through the array and parseInt each one on update?

dtsnell4:
@GordonSmith @jbrundage Bump

src/other/Legend.js
```diff
@@ -82,6 +82,7 @@
         Table.prototype.enter.apply(this, arguments);
         this.renderHtmlDataCells(true);
         this.fixedHeader(false);
+        this.fixedSize(true);
```
mlzummo: i dont think we want that as the default behavior

src/other/Table.js
```diff
@@ -232,9 +213,9 @@
         }
 
         var totalRow = [this.totalledLabel() ? this.totalledLabel() : null];
-
         if (context.totalledColumns().length !== 0) {
             var totArr = context.totalledColumns();
+            console.log(totArr);
```
mlzummo: remove comment

src/other/Legend.js
dtsnell4: No, I caught that as well...

dtsnell4:
@GordonSmith I updated this per comments.  What are your thoughts about this PR, do you not want to use it?  I think its much better that it was, even if its not perfect.  If you have other ideas or want it to go in another direction let me know.  I am working on other enhancements to the table (I have widgets working in table cells now) but I do not want to create PR's for them until this refactor is resolved.

dtsnell4:
@GordonSmith @mlzummo OK, I updated this yet again.  This should be it unless you find some errors.

Issue #1148, opened 2015-11-04
Add 'bottom' tab location support to Tabbed Layout Widget


mlzummo:

dtsnell4:
@GordonSmith Assign to me please

PR #1183, opened 2015-11-10
GH-1148 Tabs on Bottom

dtsnell4:
Add a publish parameter to place the tabs underneath the widgets instead of on top.

Fixes GH-1148

### DIFF
--- a/src/layout/Tabbed.js
+++ b/src/layout/Tabbed.js
@@ -20,7 +20,7 @@
     Tabbed.prototype.publish("activeTabIdx", 0, "number", "Index of active tab", null, {});
 
     Tabbed.prototype.publish("labels", [], "array", "Array of tab labels sharing an index with ", null, { tags: ["Private"] });
-    Tabbed.prototype.publish("tabLocation", "top", "set", "Position the tabs at the bottom of the widget", ["top", "bottom"], { tags: ["Private"] });
+    Tabbed.prototype.publish("tabLocation", "bottom", "set", "Position the tabs at the bottom of the widget", ["top", "bottom"], { tags: ["Private"] });
     Tabbed.prototype.publish("widgets", [], "widgetArray", "widgets", null, { tags: ["Private"] });
 
     Tabbed.prototype.clearTabs = function () {
@@ -63,16 +63,30 @@
     Tabbed.prototype.enter = function (domNode, element) {
         HTMLWidget.prototype.enter.apply(this, arguments);
         this._contentContainer = element.append("div");
+        this._contentContainer.attr("id", "content_container");
     };
 
     Tabbed.prototype.update = function (domNode, element) {
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
 
-        if (this._tabContainer) {this._tabContainer.remove();}
+        // if (this._tabContainer) {this._tabContainer.remove();}
+        // this._tabContainer = element.insert("div", this.tabLocation() === "bottom" ? null : "div");
+        // this._tabContainer.attr("class", "on_" + this.tabLocation());
 
-        this._tabContainer = element.insert("div", this.tabLocation() === "bottom" ? null : "div");
-        this._tabContainer.attr("class", "on_" + this.tabLocation());
+        this._tabContainer = element.selectAll("#tab_contanier").data([this.tabLocation()], function(d) {return d;});
+        this._tabContainer
+            .enter()
+            .insert("div", function(d1) {
+                var x = d1 === "top" ? "#content_container" : null;
+                console.log(x, d1);
+                return x;
+            })
+            .attr("class", function(d2) {
+                return "on_" + d2;
+            })
+            .attr("id", "tab_contanier")
+        ;
 
         element.style("padding", this.surfacePadding() + "px");
 
--- a/src/layout/Tabbed.js
+++ b/src/layout/Tabbed.js
@@ -20,6 +20,7 @@
     Tabbed.prototype.publish("activeTabIdx", 0, "number", "Index of active tab", null, {});
 
     Tabbed.prototype.publish("labels", [], "array", "Array of tab labels sharing an index with ", null, { tags: ["Private"] });
+    Tabbed.prototype.publish("tabLocation", "top", "set", "Position the tabs at the bottom of the widget", ["top", "bottom"], { tags: ["Private"] });
     Tabbed.prototype.publish("widgets", [], "widgetArray", "widgets", null, { tags: ["Private"] });
 
     Tabbed.prototype.clearTabs = function () {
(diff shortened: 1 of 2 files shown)


dtsnell4:

src/layout/Tabbed.js
```diff
@@ -61,7 +62,6 @@
 
     Tabbed.prototype.enter = function (domNode, element) {
         HTMLWidget.prototype.enter.apply(this, arguments);
-        this._tabContainer = element.append("div");
         this._contentContainer = element.append("div");
```
mlzummo: class it

src/layout/Tabbed.js
```diff
@@ -69,6 +69,14 @@
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
 
+        if (this._tabContainer) {this._tabContainer.remove();}
+        if (this.tabsOnBottom()) {
+            this._tabContainer = element.append("div");
```
mlzummo: class this?

src/layout/Tabbed.js
mlzummo: class this?

src/layout/Tabbed.js
dtsnell4: Its technically not necessary, I only classed it below because the css is different when its on the bottom.

src/layout/Tabbed.js
```diff
@@ -20,6 +20,7 @@
     Tabbed.prototype.publish("activeTabIdx", 0, "number", "Index of active tab", null, {});
 
     Tabbed.prototype.publish("labels", [], "array", "Array of tab labels sharing an index with ", null, { tags: ["Private"] });
+    Tabbed.prototype.publish("tabsOnBottom", false, "boolean", "Position the tabs at the bottom of the widget", null, { tags: ["Private"] });
```
GordonSmith: make this a set called "tabLocation",   ["top", "bottom"] with "top" the default (in case we need left/right in the future

src/layout/Tabbed.js
GordonSmith: I would set the class to match the tabLocation as per my comment above (note:  you will need to use attr("class", this.tabLocation())

src/layout/Tabbed.js
dtsnell4: @GordonSmith Should I just be proactive and add left/right now?

mlzummo: might as well

dtsnell4:
@GordonSmith updated

mlzummo:
looks good

src/layout/Tabbed.js
```diff
@@ -69,6 +69,11 @@
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
 
+        if (this._tabContainer) {this._tabContainer.remove();}
+
+        this._tabContainer = element.insert("div", this.tabLocation() === "bottom" ? null : "div");
+        this._tabContainer.classed("on_" + this.tabLocation(), true);
```
GordonSmith: This doesn't work as you will end up with several classes if the user changes tab location at runtime, you need:
```
this._tabContainer.attr("class", "on_" + this.tabLocation());
```

dtsnell4:
@GordonSmith updated

src/layout/Tabbed.js
```diff
@@ -20,6 +20,7 @@
     Tabbed.prototype.publish("activeTabIdx", 0, "number", "Index of active tab", null, {});
 
     Tabbed.prototype.publish("labels", [], "array", "Array of tab labels sharing an index with ", null, { tags: ["Private"] });
+    Tabbed.prototype.publish("tabLocation", "top", "set", "Position the tabs at the bottom of the widget", ["top", "bottom"], { tags: ["Private"] });
```
GordonSmith: Should be "basic" not private.

src/layout/Tabbed.js
```diff
@@ -69,6 +69,11 @@
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
 
+        if (this._tabContainer) {this._tabContainer.remove();}
```
GordonSmith: Removing + inserting the tab container every update cycle would be bad...

dtsnell4: What would be the alternative?

dtsnell4: I could check the class and only remove it if its changed...?

src/layout/Tabbed.js
```diff
@@ -61,14 +62,32 @@
...
+        // this._tabContainer.attr("class", "on_" + this.tabLocation());
+
+        this._tabContainer = element.selectAll("#tab_contanier").data([this.tabLocation()], function(d) {return d;});
+        this._tabContainer
+            .enter()
+            .insert("div", function(d1) {
+                var x = d1 === "top" ? "#content_container" : null;
+                console.log(x, d1);
+                return x;
+            })
```
dtsnell4: @GordonSmith I tried this approach which seemed logical, however it is not working as I expected.  When the tabLocation is set to "bottom" it works when the insert function returns "null".  But when it is set to "top" (use the property editor) and returns "#content_container" it fails.  If I remove the function and hard code "#content_container" it works as expected.  Any thoughts?

http://rawgit.com/dtsnell4/Visualization/GH-1148_Bottom_Tabs/demos/dermatology.html?src/layout/Tabbed

src/layout/Tabbed.js
```diff
@@ -61,14 +62,32 @@
...
     };
 
     Tabbed.prototype.update = function (domNode, element) {
         HTMLWidget.prototype.update.apply(this, arguments);
         var context = this;
 
+        // if (this._tabContainer) {this._tabContainer.remove();}
+        // this._tabContainer = element.insert("div", this.tabLocation() === "bottom" ? null : "div");
+        // this._tabContainer.attr("class", "on_" + this.tabLocation());
+
```
GordonSmith: Remove commented code.

src/layout/Tabbed.js
GordonSmith: You can't use tab_container ID in this way - mainly because you might have tabbed containers within tabbed containers and the selectAll will select All of them.  Look at the Grid example where it does things like:
```
var rows = this.contentDiv.selectAll(".cell_." + this._id).data(this.content(), function (d) { return d._id; });
```

src/layout/Tabbed.js
```diff
@@ -61,14 +62,32 @@
...
+            .insert("div", function(d1) {
+                var x = d1 === "top" ? "#content_container" : null;
+                console.log(x, d1);
+                return x;
+            })
+            .attr("class", function(d2) {
+                return "on_" + d2;
+            })
+            .attr("id", "tab_contanier")
+        ;
```
GordonSmith: I would set the class to be "tab_w11 top" or "tab_w11 bottom" depending - that way you can still do the selectAll on selectAll(".tab_." + this._id) for the data join and have the second class for styling...

dtsnell4:
@GordonSmith this was my first attempt at this issue, and I later issued this PR which seemed to work better.  https://github.com/hpcc-systems/Visualization/pull/1191   I updated that one per your comments, I'm just waiting to see if it gets merged, at which time I will close this PR.

PR #1201, opened 2015-11-17
GH-1144 AmChart Select Persist Fix

mlzummo:
@GordonSmith @dtsnell4 @jbrundage  - Please Review

Allows amcharts to persist selections on update if data has not been updated

Fixes GH-1144

### DIFF
--- a/src/amchart/CommonFunnel.js
+++ b/src/amchart/CommonFunnel.js
@@ -14,6 +14,10 @@
         this._chart = {};
 
         this._selected = null;
+        this._selections = [];
+        
+        this._dataUpdated = 0;
+        this._lastDataUpdated = -1;
     }
     CommonFunnel.prototype = Object.create(HTMLWidget.prototype);
     CommonFunnel.prototype.constructor = CommonFunnel;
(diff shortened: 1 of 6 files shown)


src/amchart/CommonSerial.js
```diff
@@ -288,6 +291,7 @@
                 this._chart.valueAxes[0].logarithmic = false;
                 if (this.sortDates()) {
                     this.data(Utility.naturalSort(this.data(), "ascending", 0));
+                    this._dataUpdated--; // decrament the counter becuase its not a true data update
```
mlzummo: @GordonSmith i could use your 2cents here ... im doing a data update but is there a way i can just change the underlying without calling it with this.data(x); can i use do _data? here without any consequence?

mlzummo: perhaps i should just assign to a new variable ... just trying to use ness memory

src/amchart/CommonFunnel.js
```diff
@@ -77,10 +81,13 @@
         if(this.reverseDataSorting()){
             sortingMethod = function(a,b){ return a[1] < b[1] ? 1 : -1; };
         }
-        this.data(this.data().sort(sortingMethod));
+        this.data().sort(sortingMethod);
 
         // DataProvider
-        this._chart.dataProvider = this.formatData(this.data());
+        if (this._dataUpdated > this._lastDataUpdated) {
+            this._chart.dataProvider = this.formatData(this.data());
```
mlzummo: would be nifty to maybe use the d3 array adapter to "monitor" for changes in data ... not sure if im thinking right about that though @GordonSmith thoughts?

GordonSmith:
This still seems too complex, how about this pseudo logic:
onClick: 
1.  hash the row that was clicked
2.  if the ```hash === this._selectedHash``` then ```delete this._selectedHash``` else ```this._selectedHash = hash;```
3. trigger a new render

onUpdate:
1.  Color any data row whose hash equals the this._selectedHash differently.

mlzummo:
this will mess up the color rotation i believe which is what i am trying to avoid as well

mlzummo:
@GordonSmith please merge

PR #1212, opened 2015-11-19
GH-1211 Fixes column/data call in MegaChart

mlzummo:
@GordonSmith Please Review

Fixes GH-1211

### DIFF
--- a/src/composite/MegaChart.js
+++ b/src/composite/MegaChart.js
@@ -71,8 +71,8 @@
         
             this._layout
                 .getContent("center")
-                    .data(this.data())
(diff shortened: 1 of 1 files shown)


src/composite/MegaChart.js
```diff
@@ -71,8 +71,8 @@
         
             this._layout
                 .getContent("center")
-                    .data(this.data())
-                    .columns(this.columns());
+                    .columns(this.columns())
+                    .data(this.data());
```
GordonSmith: The order shouldn't matter - as it should not be acted on until render is called...

Issue #1214, opened 2015-11-20
Google empty chart displays an error


GordonSmith:

GordonSmith:
Fix merged

PR #1215, opened 2015-11-20
GH-1214 Empty Google Chart Error

GordonSmith:
Create / Remove Google Charts on demand.

### DIFF
--- a/src/google/Common.js
+++ b/src/google/Common.js
@@ -12,11 +12,6 @@
 
         this._tag = "div";
 
-        this.columns([]);
-        this.data([]);
-        this._data_google = google.visualization.arrayToDataTable([["", { role: "annotation" }],["",""]]);
-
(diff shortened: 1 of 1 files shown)


mlzummo:
so far with my tests this looks ok ...

mlzummo:
if you delete the columns first and then the data it errors but not the other way around which is good and when u delete both its fine ...

mlzummo:
should prolly check that there is a correct number of columns matching data?

GordonSmith:
@mlzummo:  If there is bad column/data info then we should still throw it at the google chart and let it render it or display an error (as it might still render something?)

Also in other charts we will have extra "columns" in the data structure which is used in the click event, but not used in the render process...  So we can't check generically (its basically an embedded payload).

mlzummo:
ok looks good then

Issue #1161, opened 2015-11-05
Table:  Ability to embed a widget into a cell.


GordonSmith:
Common use case will be embedding another table inside a single cell (a child dataset).

mlzummo:
Also form inputs

mlzummo:
part of GH-724

dtsnell4:
While this seems straight forward enough, I do not know what the data will look like providing the widget for a particular cell.  Can anyone provide a real world example of what the data() is going to look like?  How does the widget get passed in?  In regards to GH-724, what is a child dataset in this context?

GordonSmith:
```
table.data([["math", 22, new TextBox().text("Hello"), 33]]);
```

GordonSmith:
(we may not be able to support pagination with this option - not easily anyway)

PR #1237, opened 2015-11-30
GH-1161 Ability to embed a widget into a Table Widget cell

dtsnell4:
Adds the ability to include other widgets in the data to be displayed in a table cell.

Fixes GH-1161


dtsnell4:
@GordonSmith @mlzummo @jbrundage This is functioning well, but there are definitely some issues with it.  I will comment on them in the code itself.  Please post your comments/suggestions

src/other/Table.js
```diff
@@ -1,12 +1,13 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "../common/HTMLWidget", "./Paginator", "../common/Utility", "css!./Table"], factory);
+        define(["d3", "../common/HTMLWidget", "./Paginator", "../common/Utility", "../common/Widget", "css!./Table"], factory);
```
dtsnell4: This is the first issue, I realize this cannot be the right way to do this, requiring HTMLWidget and Widget at the same time since HTMLWidget extends Widget.   But in order to accomplish this task I need to be able to test if a data item is widget, obviously, for example here https://github.com/hpcc-systems/Visualization/pull/1237/files#diff-164179c53aa545afdc2035238236796aR345.  What would be the appropriate way to accomplish that?

mlzummo: yes you should be able to do instance of Widget even if you only include HTMLWidget

src/other/Table.js
```diff
@@ -178,6 +186,20 @@
...
+                var d = this._db._data[iDx1];
+                for (var iDx2 = 0; iDx2 < d.length; iDx2 ++) {
+                    if (d[iDx2] instanceof Widget) {
+                        context.pagination(false);
+                        break;
+                    }
+                }
+                if (context.pagination() === false) {break;}
+            }
+        }
```
dtsnell4: As we already discussed, pagination obviously breaks.  For now I just test the data to see if it includes a widget, then disable pagination if it does.  I will add a separate issue to to work on the pagination, as that will be a task in itself. FYI, I believe all the other features (fixed headers, totals, formatting etc.) all work at this point.

src/other/Table.js
```diff
@@ -307,27 +325,40 @@
...
                     .style("text-align", alignment)
+                    .classed("tr-"+trIdx+"-td-"+tdIdx,true)
                 ;
+                if (tdContents instanceof Widget) {
+                    tdContents.target(this);
+                    tdContents._parentWidget = context;
+                    if (tdContents._class.indexOf("childWidget") < 0) {
+                        tdContents._class += " childWidget";
+                    }
+                    tdContents.render();
```
dtsnell4: I know the child widgets will render themselves via the widgetArray code, but the problem here is that if they are not rendered now, they will not have a size unless it is explicitly assigned, and the DOM will not know how big to make the table cell containing the widget.  Most charts will need a specific size, but this becomes a problem when you have something like a small nested table, that you want to be displayed in it s full size without scroll bars.  Suggestions?

test/DataFactory.js
```diff
@@ -435,6 +435,52 @@
...
                 ],
                 columns: ["Label", "Lat", "Long", "Pin", "Forth Column", "Fifth Column", "sixth Column", "Seventh Column", "eighth Column", "Nineth Column", "Tenth Column"]
+            },
+            widget: {
+                data: [
+                    ["Label1", 37.665074, -122.384375, "green-dot.png", 4525, 4243545, 24354, 54, 2552345, 2455, 245435],
+                    ["Label2", 32.690680, -117.178540, new Input().name("checkbox-test").label("Checkbox Test").type("checkbox").value(true), 4525, 423545, 24354, 354, 2552345, 2455, 245435],
+                    ["Label3", 39.709455, -104.969859, new Table().fixedSize(true).fixedHeader(false).columns( ["Subject","Year 1", "Year 2", "Year 3"]).data([["Math",76,65,87],["History",65,87,67],["Science",88,91,78]]), 4525, 423545, 24354, 524, 2552345, 2455, 245435],
+                    ["Label4", 41.244123, -95.961610, new Column().size({width: 200, height: 200}).columns( ["Subject","Year 1", "Year 2", "Year 3"]).data([["Math",76,65,87],["History",65,87,67],["Science",88,91,78]]), 4525, 423545, 24354, 564, 2552345, 2455, 245435],
```
dtsnell4: Note the column needs to have a size assigned.  The table, in this instance, you want to have its natural size, whatever that may be.

src/other/Table.js
mlzummo: common widget types such as input/checkbox will more than likely have static heights that we can make assumptions with (that or we can loop thrugh the data set and get each widgets height and get the max height for that row

src/other/Table.js
dtsnell4: It did not work when I tried.

src/other/Legend.js
```diff
@@ -82,6 +82,7 @@
         Table.prototype.enter.apply(this, arguments);
         this.renderHtmlDataCells(true);
         this.fixedHeader(false);
+        this.fixedSize(true);
```
GordonSmith: The legend and mega chart need some fixing and I suspect any fixes there will affect this as well.

src/common/Widget.js
```diff
@@ -452,7 +452,13 @@
     Widget.prototype.rowToObj = function (row) {
         var retVal = {};
         this.columns().forEach(function(col, idx) {
-            retVal[col] = row[idx];
+            var item;
+            if (row[idx] instanceof Widget) {
+                item = row[idx]._class;
```
GordonSmith: Why not just include the widget?

src/other/Table.js
GordonSmith: @dtsnell4 unfortunately for instanceof to work it needs to include the Widget definition as you have done above.  So its technically correct (even if it does not feel like it).

src/other/Table.js
```diff
@@ -1,12 +1,13 @@
...
     if (typeof define === "function" && define.amd) {
-        define(["d3", "../common/HTMLWidget", "./Paginator", "../common/Utility", "css!./Table"], factory);
+        define(["d3", "../common/HTMLWidget", "./Paginator", "../common/Utility", "../common/Widget", "css!./Table"], factory);
     } else {
         root.other_Table = factory(root.d3, root.common_HTMLWidget, root.other_Paginator, root.common_Utility);
     }
-}(this, function (d3, HTMLWidget, Paginator, Utility) {
+}(this, function (d3, HTMLWidget, Paginator, Utility, Widget) {
     function Table() {
+        Widget.call(this);
```
GordonSmith: Remove this - not needed.

src/other/Table.js
```diff
@@ -17,6 +18,7 @@
         this._selectionPrevClick = null;
         this._paginatorTableSpacing = 4;
     }
+    Table.prototype = Object.create(Widget.prototype);
```
GordonSmith: Remove this not needed.

src/other/Table.js
```diff
@@ -17,6 +18,7 @@
...
     Table.prototype = Object.create(HTMLWidget.prototype);
```
GordonSmith: Just to elaborate:
```
Table.prototype = Object.create(HTMLWidget.prototype);
```
Is basically saying - let Table.prototype equal HTMLWidget.prototype.  The effect of this is that TAble now has all the same methods that HTMLWidget had defined for itself (which in turn included all the Widget.prototype methods).

src/other/Table.js
```diff
@@ -66,6 +68,7 @@
     Table.prototype.publish("columnPatterns", [], "array", "Array of formatting rules for each column", null, { tags: ["Basic"], optional: true });
     Table.prototype.publish("stringAlign", "left", "set", "Array of alignment positions for strings", ["left","right","center"], { tags: ["Basic"], optional: true });
     Table.prototype.publish("numberAlign", "right", "set", "Array of alignment positions for numbers", ["left","right","center"], { tags: ["Basic"], optional: true });
+    Table.prototype.publish("childWidgets", [], "widgetArray", "Array of child widgets", null, { tags: ["Basic"], optional: true });
```
GordonSmith: (I am sure I will see further down, but I don't see why this would be needed, since the widgets are coming in via the "data" method)?

src/other/Table.js
```diff
@@ -178,6 +186,20 @@
         ;
         th.order();
 
+        //Disable Pagination if a widget is in the data
+        if (context.pagination()) {
+            for (var iDx1 = 0; iDx1 < this._db._data.length; iDx1 ++) {
```
GordonSmith: This breaks the "Class encapsulation" - IOW you should _not_ be reaching inside "_db".  Also this is potentially expensive...

You could have achieved the same thing with the following:
```
var hasEmbeddedWidget = this.data().some(function(row) {
    return row.some(function(cell) {
        return cell instanceof Widget;
    });
});
```

src/other/Table.js
GordonSmith: Rather than testing upfront I think I would simply spit out a console warning IF pagination is on and you happen to hit an embedded widget.

src/other/Table.js
```diff
@@ -290,11 +305,14 @@
             })
         ;
         rows
-            .attr("class", function (d) {
+            .classed("selected", function (d) {
                 if (context._selectionBag.isSelected(context._createSelectionObject(d))) {
-                    return "selected";
+                    return true;
```
GordonSmith: Can now be:
```
.classed("selected", context._selectionBag.isSelected(context._createSelectionObject(d)))
```

src/other/Table.js
```diff
@@ -290,11 +305,14 @@
...
                 }
             })
+            .classed("trId" + context._id, function(){
+                return true;
+            })
```
GordonSmith: function not needed.

src/other/Table.js
```diff
@@ -307,27 +325,40 @@
             .append("td")
         ;
         cells[this.renderHtmlDataCells() ? "html" : "text"](function (d, idx) { 
-            var retVal = context.columnPatterns()[idx] ? context.getColumnFormatting(d, idx) : d;
+            var retVal;
+            if (!(d instanceof Widget)) {
+                retVal = context.columnPatterns()[idx] ? context.getColumnFormatting(d, idx) : d;
+            }
             return retVal;
```
GordonSmith: Your returning "undefined" for the cell when its a widget?  Should it be ""?

src/other/Table.js
GordonSmith: I suspect that removing the "childWidgets" PP completely is the way to go (especially when it comes to serialization...).

src/other/Table.js
```diff
@@ -469,18 +503,22 @@
             }
             this.fixedCol
                 .style("position", "absolute")
-                .style("margin-top", -this.tableDiv.node().scrollTop + tableMarginHeight + "px")
+                .style("margin-top", -this.tableDiv.node().scrollTop + parseInt(tableMarginHeight) + "px")
```
GordonSmith: I would move the parseInt up to where the value was originally assigned (if possible).

src/other/Table.js
```diff
@@ -469,18 +503,22 @@
...
             ;
             fixedColTd
                 .style("width", fixedColWidth + "px")
             ;
             this.fixedColHead
                 .style("position", "absolute")
-                .style("margin-top", (this.fixedHeader() ? this.tableDiv.node().scrollTop: 0) - tableMarginHeight + "px")
+                .style("margin-top", (this.fixedHeader() ? this.tableDiv.node().scrollTop: 0) - parseInt(tableMarginHeight) + "px")
```
GordonSmith: I would move the parseInt up to where the value was originally assigned (if possible) - then this instance would not have been needed....

src/other/Table.js
```diff
@@ -508,7 +546,7 @@
                 calcWidth();
                 calcHeight();
             } else {
-                if (box.height - tableMarginHeight <= context.tableDiv.node().offsetHeight ) {
+                if (box.height - parseInt(tableMarginHeight) <= context.tableDiv.node().offsetHeight ) {
```
GordonSmith: (see above comments re parseInt)

src/other/Table.js
```diff
@@ -529,27 +567,35 @@
...
+            if (element.classed("childWidget")) {
+                context._parentElement
+                    .style("width", finalWidth)
+                    .style("height", newTableHeight)
+                ;
+                context.tableDiv
+                    .style("overflow", "hidden")
+                ;
+            }
+            context.size({width:parseInt(finalWidth), height: parseInt(newTableHeight)});
```
GordonSmith: Minor:  Have the calcWidth function return just the width and no "px"...  The reason I prefer this is it means you add the "px" inline when you set the style, thus keeping it "grouped" logically.

src/other/Table.js
```diff
@@ -529,27 +567,35 @@
...
-            maxWidth = context.table.node().offsetWidth - fixedColWidth + context.getScrollbarWidth();
+            var newTableWidth = box.width;
+            maxWidth = context.table.node().offsetWidth + 1;
             finalWidth = newTableWidth > maxWidth ? maxWidth : newTableWidth;
             finalWidth = finalWidth + "px";
         }
 
         function calcHeight() {
-            newTableHeight = box.height - tableMarginHeight + context.getScrollbarWidth();
+            newTableHeight = context.tbody.node().offsetHeight + parseInt(tableMarginHeight) + 1;
```
GordonSmith: Can you access "offsetrHeight" via context.tbody.attr("offsetHeight") or context.tbody.property("offsetHeight")?  If so it would be "safer".

src/other/Table.js
```diff
@@ -557,8 +603,41 @@
         HTMLWidget.prototype.exit.apply(this, arguments);
     };
         
+    Table.prototype.setColumnWidths = function() {
+        var context = this;
+        var tableMarginHeight = this.fixedHeader() ? this.thead.node().offsetHeight : 0;
+
+        var totalWidth = 0;
+        var tds = this.table.select("tbody").selectAll("tr").selectAll("td");
+        var th = this.thead.selectAll("th").data(this.showHeader() ? this.columns() : [], function (d) { return d; });
```
GordonSmith: Are you sure you want ```, function (d) { return d; }```?

src/other/Table.js
```diff
@@ -557,8 +603,41 @@
...
+        th.each(function(d, i) {
+            var thwidth = this.offsetWidth;
+            var tdwidth = tds.empty() || !tds[0][i] ? 0 : tds[0][i].offsetWidth;
```
GordonSmith: You should not be reaching inside the tds[0] structure...

jbrundage: Are "widgets-within-cells" actual children of the td elements? (As opposed to overlays)

If they're child elements then this selector will break when a Table exists within a Table.
`var tds = this.table.select("tbody").selectAll("tr").selectAll("td")`

src/other/Table.js
```diff
@@ -307,27 +325,40 @@
...
             return retVal; 
         });
         cells.exit()
             .remove()
-        ;      
+        ;
+
         rows.each(function(tr,trIdx){
-            d3.select(this).selectAll("td").each(function(tdContents,tdIdx){
+            d3.select(this).selectAll("td").each(function(tdContents,tdIdx){
```
jbrundage: @dtsnell4 This selector will grab unexpected elements when a Table widget exists within another Table widget.

src/other/Table.js
```diff
@@ -307,27 +325,40 @@
...
             });
             var dis = d3.select(this);
             context.applyStyleToRows(dis);
         });  
 
-        var tableMarginHeight = this.thead.node().offsetHeight;
+        var tableMarginHeight = this.thead.style("height");
 
-        var tbodyRows = context.table.select("tbody tr");
+        var tbodyRows = this.table.select("tbody").select("tr");
```
jbrundage: @dtsnell4 This selector will grab unexpected elements when a Table widget exists within another Table widget.

src/other/Table.js
```diff
@@ -469,18 +503,22 @@
...
             this.fixedColHead
                 .style("position", "absolute")
-                .style("margin-top", (this.fixedHeader() ? this.tableDiv.node().scrollTop: 0) - tableMarginHeight + "px")
+                .style("margin-top", (this.fixedHeader() ? this.tableDiv.node().scrollTop: 0) - parseInt(tableMarginHeight) + "px")
             ;
             fixedColTh
                 .style("width", fixedColWidth + "px")
             ;
+            rows.each(function(d, i) {
+                var height = d3.select(this).select("td").node().offsetHeight;
```
jbrundage: @dtsnell4 This selector will grab unexpected elements when a Table widget exists within another Table widget.

src/other/Table.js
```diff
@@ -557,8 +603,41 @@
         HTMLWidget.prototype.exit.apply(this, arguments);
     };
         
+    Table.prototype.setColumnWidths = function() {
+        var context = this;
+        var tableMarginHeight = this.fixedHeader() ? this.thead.node().offsetHeight : 0;
+
+        var totalWidth = 0;
+        var tds = this.table.select("tbody").selectAll("tr").selectAll("td");
```
jbrundage: @dtsnell4 This selector will grab unexpected elements when a Table widget exists within another Table widget.

src/other/Table.js
```diff
@@ -557,8 +603,41 @@
...
+            .style("margin-top", (context.fixedHeader() ? tableMarginHeight : 0) + "px" )
+        ;
+        this.tbody
+            .style("width", totalWidth + "px" )
+        ;
+    };
+
     Table.prototype.getBodyRow = function(i) {
-        return this.table.selectAll("tbody tr")
+        return this.table.selectAll("tbody tr.trId" + this._id)
```
jbrundage: @dtsnell4 This selector will grab unexpected elements when a Table widget exists within another Table widget.

src/other/Table.js
dtsnell4: And just let the pagination break?  That's fine with me, it will hopefully be updated soon and it will be a non-issue.

src/other/Table.js
dtsnell4: Yes, of course... I had more code in these functions earlier that I removed, did not think to condense it

dtsnell4: Actually I stand corrected, the function is needed for this one because `_createSelectionObject()` needs "d".  Though I will rewrite it as 

`.classed("selected", function (d) { return context._selectionBag.isSelected(context._createSelectionObject(d))})`

It is not needed for the next one though.

src/other/Table.js
dtsnell4: @GordonSmith it returns undefined here because the widget gets rendered in the cell here https://github.com/hpcc-systems/Visualization/pull/1237/files#diff-164179c53aa545afdc2035238236796aR351

Is it better to just return ""?  Does it matter?

src/other/Table.js
dtsnell4: Done

src/other/Table.js
GordonSmith: I think he gets away with it because it will grab the first one - if it was SelectAll there would be trouble...

src/other/Table.js
GordonSmith: I think this one is also ok.

src/other/Table.js
GordonSmith: Yup - try this:  ``` var tds = this.table.selectAll("tbody >> tr >> td");```

GordonSmith: Yup

src/other/Table.js
GordonSmith: I think this is ok as its using the local ID?

src/other/Table.js
dtsnell4: OK, very good.  I was careful (to at least try) not to create any circular references.  If this is the best way, I'll leave it ass is except for the following comments...

src/other/Table.js
GordonSmith: Your correct - my bad...

src/other/Table.js
GordonSmith: undefined and null are probably handled the same way - so returning "" would probably be the wrong thing to do?

src/other/Table.js
dtsnell4: Technically it can't since in this case I'm using "select" instead of "selectAll", so it only returns the first `td`.  You can see an embedded table in the demo http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget

src/other/Table.js
dtsnell4: This should have been `select("tr")` instead of `selectAll` as per my comment above, thanks for catching.  It is interesting to note however, it does still work.  This selection is used to match the size of the body cells to their corresponding header cells.  The code below takes this selection and gets the size of the cells, compares them, and uses the largest value as the size for both (header and body).  Since a child table can't really have a `td` with a larger size than the cell that's containing it, its value will never be used.  That said, its a mute point because "select" is the correct method here.

src/other/Table.js
dtsnell4: @GordonSmith If a td is empty, I set the width to 0.  The value obtained here is then compared to the width of the corresponding header cell (thWidth), and the larger of the two values is used as the width for both.

src/other/Table.js
jbrundage: Yes, looks okay on second thought

dtsnell4:
@GordonSmith @mlzummo @jbrundage OK, I updated everything that was commented on.  But unfortunately its still failing the Serialization test and travis, and I see why.  I just don't *understand* why.  I left a console log in where the error is occurring.  If you could take a look [here](http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget) and view the console, you can see that when a child widget gets gets created in the serialization test, it is a reference to the child widget in the first table.  When HTMLWidget tries to assign it a target cell in the serialization test table, it gives the "Target can only be assigned once" error.  

I really don't understand serialization well enough to see whats going on here.  I already clear the target and childwidget array at the beginning of update() [here](https://github.com/hpcc-systems/Visualization/pull/1237/files#diff-164179c53aa545afdc2035238236796aR131) so I'm not sure why this is happening.  Any comments will be appreciated

jbrundage:
@dtsnell4 @mlzummo @GordonSmith It looks like **serializeToObject** needs to "clone" any widgets found within .data()

src/common/Widget.js
dtsnell4: @GordonSmith I was just about to ask you about this...  I did this for testing purposes, because in the click event JSON.stringify returns a circular reference when a widget is included.  Obviously this needs to return the widget itself, so the question is really how should I handle the logging of the click event?

src/other/Table.js
```diff
@@ -759,8 +805,8 @@
         }
      };
 
-    Table.prototype.click = function (row, column) {
-        console.log("Click:  " + JSON.stringify(row) + ", " + column);
+    Table.prototype.click = function (row, column, selected) {
+        console.log("Click:  " + JSON.stringify(row) + ", " + column + "," + selected);
```
dtsnell4: @GordonSmith I replied to your comment in rowToObj() Widget.js, and removed my changes, so the comment is hidden.  The reason I made that change was for just for testing.  In the click event JSON.stringify returns a circular structure error when a widget is included. Obviously rowToObj() needs to return the widget itself, so the question is really how should I handle the logging of the click event?

src/common/Widget.js
GordonSmith: You can add an extra function to JSON.stringify which will let you skip Objects (this issue is probably only happening in a console.log call?).

GordonSmith:
Serializing data() + columns is something that we don't really need/use?  its nice to have for the dermatology testing, but its not a requirement (I think)?

GordonSmith:
@dtsnell4 - if you open:  http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/marshaller/HTML.databomb 

It will point you at one of the failing issues (I suspect most of them)...

dtsnell4:
@GordonSmith I fixed the NEW Travis issues, but the original one still exists ("Target can only be assigned once" error in serialization test).  Not sure how to handle that still.

I also updated JSON.stringify to handle the presence of widgets with a replacer function (was not aware of that feature).

mlzummo:
@GordonSmith @dtsnell4 this is becuase its reusing the same objects inside data ... we should prolly just set null first becuase if were serializing anyways were never going to have a copy like that on the same page ... atleast i hope not ... if thats the case then we need to clone the objects in data everytime and then delete the ones passed in to free up memory

dtsnell4:
@GordonSmith with Matt's help it looks like I got the serialization test working finally.

GordonSmith:
@dtsnell4 What determines the column width?  Looking at this:  
image
I would expect the column chart to use that space?

Also I suspect we need to open an issue to allow the user to resize columns (fun I know)...

dtsnell4:
@GordonSmith With any of the chart widgets, the user will have to set its size.  I do not see any way for the widgets to "know" what size to be automatically.  The table widget has the "fixedSize" parameter already (which sets it to its natural size), so if you know you have a small enough child table, you just set it to true.  The reason there is blank space next to the column chart is that I explicitly set to to 200x200 pixels.  The child table is bigger so the main table column is bigger, thus the blank space. Am I making sense?

dtsnell4:
If I do not size the column widget and set the table widget fixedSize(false), this is what it would look like

capture

GordonSmith:
I would suggest adding a PP for "minWIdgetWIdth", then during render if a column has a widget and its width is less than the min, set it (the column) to the min width?  Then let the widgets just resize upwards?

dtsnell4:
Ok, it will need a min row height as well.

dtsnell4:
@GordonSmith I implemented what you suggested.  I assumed the min width property went in the Table widget, and would be applied to all child widgets.  That introduces another set of problems as you can see here.

http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget

How would  go about this on an individual widget basis?

dtsnell4:
@GordonSmith Bump

GordonSmith:
I see what you mean...  I _think_ its better that way and in the real world each column will probably only ever have 1 type of widget.  

Also I suspect we simply might need different logic for the table in table scenario?

dtsnell4:
@GordonSmith @mlzummo @jbrundage The table does have its own logic for when its a child widget.  I have updated everything any found a couple of edge case bugs that I fixed.  Aside from the pagination not working correctly with child widgets, I believe it is good to go unless you see something else.   

http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget

dtsnell4:
@GordonSmith @jbrundage @mlzummo I updated this with a temporary fix to the Property editor issue https://github.com/hpcc-systems/Visualization/issues/1282#issuecomment-164045942.  This is not really a long term solution, but I just wanted to get it working again.  Honestly, I'm not sure what the solution is, we will probably need to serialize the widgets in the data to be displayed in the property editor so that the data can be updated there.  

Secondly, I realized I could make the temporary pagination significantly more usable just by adding a scrollbar when there are widgets in the data.  It may not be exactly what the user expects to see, but all the data is there and view-able and accessible.  Turn on pagination to see http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget&pagination=true&itemsPerPage=22

GordonSmith:
I think the pagination "improvement" is a good idea, we will find it hard to get it 100% correct and simply not supporting pagination for data with widgets is ok for the foreseeable future...

Also I don't see any reason why we need to support persisting data (let alone widgets in data) at the moment, except that it is useful for us in the dermatology test page.  Having said that, rather than simply doing a JSON.stringify, we could simply do a Persist.serialize and we would be half way there to supporting it?

src/other/PropertyEditor.js
```diff
@@ -426,7 +426,12 @@
         switch (param.type) {
             case "array":
             case "object":
-                element.property("value", JSON.stringify(val));
+                element.property("value", JSON.stringify(val, function replacer(key, value) {
+                    if (value instanceof Widget) {
+                        return "Widget with class: " + value.classID();
+                    }
+                    return value;
+                }));
```
GordonSmith: How about:
```
element.property("value", val instanceof Widget ? Persist.serialize(val) : JSON.stringify(val));
```

GordonSmith:
@mlzummo I think this is ready to be pulled (once the final comments have been addressed) do you agree?

src/other/PropertyEditor.js
dtsnell4: Perfect

dtsnell4:
@GordonSmith Yes this whole JSON.stringify issue is simply an exercise in keeping the widget functioning in Dermatology.  Your solution is perfect until we figure out what direction this will go in the future.

dtsnell4:
@GordonSmith @mlzummo @jbrundage  OK The last update has been made.  If it passes travis checks Im good to go.

jbrundage:
This table cell has inline styles of **width:200px;height:200px;** but the table within that cell is forcing the dimensions to be **width:312px;height:200px;**, this can/will cause unpredictable behavior in different browsers.

dtsnell4:
This is necessary based on other requirements.  You can have width and height styles on a table td all day, but they are the first thing the browser will ignore.  They are always over ridden based on the content of the cell.  A lot of times that is a pain, but here it actually works in our favor.  Also the 200x200 px is a publish param.  Just leave it blank and they wont be added. This will generally never even be an issue unless the user is putting different types of widgets in the same column, which I have been told is not likely not happen.

That said, thanks for bringing this up because the table is now broken in IE for some other reason.  It was working...  I will look into it.

GordonSmith:
@dtsnell4 Are you going to fix the IE issue in this branch or another one?  Or is this ready to be pulled?

src/other/Table.js
```diff
@@ -307,55 +305,67 @@
...
+                    console.log('el.node().offsetHeight: ', el.node().offsetHeight);
+                    tdContents.target(null);
+                    tdContents.target(this);
+                    tdContents._parentWidget = context;
+                    if (tdContents._class.indexOf("childWidget") < 0) {
+                        tdContents._class = "childWidget " + tdContents._class;
+                    }
+                    tdContents.render();
+                    console.log("Widget._size.height: ", tdContents._size.height);
+                    console.log(" ")
```
dtsnell4: @GordonSmith referring to the IE problems...   I'm not sure what is happening there...  I added the above console logs to debug.  Can you look at them in Chrome and IE? 
http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget
In Chrome all of those evaluate to the proper height (200), however in IE the `el.style("height")` evaluates to 0 (and thus the `Widget._size.height`).  This must be why the widgets have no height in the DOM, but I have no idea why or what to do about it...  

Again this is just an issue with the widgets, regular tables are unaffected.

dtsnell4:
@GordonSmith @jbrundage 
OK, while I hate to admit it, an IE bug actually helped helped me out for once.  While trying to figure out why IE was misbehaving, I realized a much better way for handling the sizing of the widgets in the cells.  The cells no longer forced to be the minWidgetHeight, the will shrink to fit the widget when appropriate.  And a widget will not be forced to the minWidgetWidth, it will expand to the size of the cell if there is room.  Take a look at it with a [minWidgetWidth(200)](http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget&minWidgetWidth=200&minWidgetHeight=200) and a [minWidgetHeight(400)](http://rawgit.com/dtsnell4/Visualization/GH-1161_Table_Child_Widgets/demos/dermatology.html?src/other/Table.widget&minWidgetWidth=400&minWidgetHeight=200).

I know what you might be inclined to say...  "The table does not expand to fit the cell like the Column does".  Well, that's how the table widget is, it only gets as big as its content, see candidate-1.6 http://rawgit.com/hpcc-systems/Visualization/candidate-1.6.10/demos/dermatology.html?src/marshaller/HTML.  Expanding to fit would be a new feature.

At the moment I think its working really well.  With the exception of pagination showing a scrollbar when child widgets are in the table, It seems to working as a user would expect it to.  I think its good to go if there aren't any more issues.

Issue #766, opened 2015-08-21
Add second Y Axis to C3


GordonSmith:

jbrundage:
@mlzummo @dtsnell4 @GordonSmith Do we have a test/xxxxxxFactory example of the standard/expected columns+data format that would result in a chart with two Y-Axes?

GordonSmith:
Not that I know off...

GordonSmith:
Closed "timed-out", please reopen if needed.

PR #1239, opened 2015-12-01
GH-766 Adds C3 2nd Y-Axis capability

jbrundage:
Fixes GH-766

http://rawgit.com/jbrundage/Visualization/GH-766/demos/dermatology.html?src/c3chart/Combo.multiY

http://rawgit.com/jbrundage/Visualization/GH-766/demos/dermatology.html?src/c3chart/Column.multiY

http://rawgit.com/jbrundage/Visualization/GH-766/demos/dermatology.html?src/c3chart/Area.multiY

### DIFF
--- a/src/c3chart/CommonND.js
+++ b/src/c3chart/CommonND.js
@@ -66,6 +66,8 @@
     CommonND.prototype.publish("yAxisTypeTimePattern", "%Y-%m-%d", "string", "Time Series Pattern", null, {});
 
     CommonND.prototype.publish("axisTickLabelMultiLine", false, "boolean", "Show Y Grid",null,{tags:["Intermediate"]});
+    
+    CommonND.prototype.publish("y2", [], "array", "Array of columns to be associated with a Second Y-Axis",null,{tags:["Intermediate"]});
 
     CommonND.prototype.enter = function (domNode, element) {
         if (this.subchart()) {
@@ -104,10 +106,10 @@
         };
         this._config.grid = {
             x: {
-                show: this.showXGrid(),
+                show: this.showXGrid()
             },
             y: {
-                show: this.showYGrid(),
+                show: this.showYGrid()
             }
         };
 
@@ -131,6 +133,23 @@
             this._config.data.x = this.columns()[0];
             this._config.data.xFormat = this.xAxisTypeTimePattern();
             break;
+        }
+
+        if(this.y2() instanceof Array && this.y2().length > 0){
+            var axesObj = {};
+            var y2arr = this.y2();
+            this._config.axis.y2 = {
+                show: y2arr.length > 0
+            };
+            if(this.columns().length > 1){
+                var cols = this.columns();
+                for(var i = 0;i<y2arr.length;i++){
+                    for(var j = 1;j<cols.length;j++){
+                        axesObj[cols[j]] = y2arr.indexOf(cols[j]) !== -1 ? "y2" : "y";
+                    }
+                }
+                this._config.data.axes = axesObj;
+            }
         }
 
         Common.prototype.enter.apply(this, arguments);
--- a/test/c3chartFactory.js
+++ b/test/c3chartFactory.js
@@ -16,6 +16,15 @@
                     );
                 });
             },
+            multiY: function (callback) {
+                require(["test/DataFactory", "src/c3chart/Bar"], function (DataFactory, Widget) {
+                    callback(new Widget()
+                        .y2(["C2","C3"])
+                        .columns(DataFactory.ND.multiY.columns)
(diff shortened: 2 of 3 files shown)


jbrundage:
This implementation simply adds **y2** array-type publish parameter. The "columns" that should belong to the Y2 axis are specified in the y2 publish parameter array.

src/c3chart/CommonND.js
```diff
@@ -108,7 +110,7 @@
             },
             y: {
                 show: this.showYGrid(),
-            }
+            },
```
GordonSmith: remove trailing comma

src/c3chart/CommonND.js
```diff
@@ -108,7 +110,7 @@
             },
             y: {
                 show: this.showYGrid(),
```
GordonSmith: remove trailing comma - I would have though lint would have caught these?

jbrundage: I don't think lint checks for trailing commas

mlzummo:
@jbrundage is there a way to configure the positions of the Axes?

mlzummo:
not sure if this is a general issue but looks like the default behavior for c3 selection is multi selection?

mlzummo:
also should have capability of coloring the axis line the same color as the datapoints they belong to or some kind of legend?

dtsnell4:
As for configuring the axes, my assumption was just changing the y2 publish param to "C1" would do it, but that has no affect on the widget at all

mlzummo:
@dtsnell4 im not sure i follow what your saying

src/c3chart/CommonND.js
```diff
@@ -133,6 +135,23 @@
...
+            if(this.columns().length > 1){
+                var cols = this.columns();
+                for(var i = 0;i<y2arr.length;i++){
+                    for(var j = 1;j<cols.length;j++){
+                        axesObj[cols[j]] = y2arr.indexOf(cols[j]) !== -1 ? "y2" : "y";
+                    }
+                }
+                this._config.data.axes = axesObj;
+            }
+        }
```
dtsnell4: @mlzummo Sorry, I mean changing the y2 publish param in the property editor.   This array is what determines what is associated with the second axis.  In the test data that is [c2, c3] (second and third columns).  I am assuming changing this to [c1] would associate the first column of data with the second y axis.  

The second point of my comment was that doing this in the _Property Editor_ has not effect, which is probably an issue of the Property editor itself, not the chart, but I don't know that for sure.

src/c3chart/CommonND.js
```diff
@@ -133,6 +135,23 @@
             break;
         }
 
+        if(this.y2() instanceof Array && this.y2().length > 0){
```
mlzummo: if y2 is of type array we dont need to check if its an array?  (though i know sometimes we set things null)

mlzummo:
should we basing the second y axis off the columns or should we use the whole data row?

mlzummo:
same goes for combo chart ...?

src/c3chart/CommonND.js
jbrundage: @dtsnell4 It's an issue with the chart. As far as I'm aware... C3 doesn't allow adding a second Y axis to a chart that's already initialized.

src/c3chart/CommonND.js
jbrundage: Yeah, I just coded that with "null" in mind... to prevent this.y2().length from breaking

GordonSmith:
@mlzummo, @jbrundage can you compare the test factory instantiation code for your two implementations and see how similar the API calls are (I would expect them to be the same)?

test/c3chartFactory.js
```diff
@@ -37,6 +46,15 @@
...
                 });
+            },
+            multiY: function (callback) {
+                require(["test/DataFactory", "src/c3chart/Column"], function (DataFactory, Widget) {
+                    callback(new Widget()
+                        .y2(["C2","C3"])
+                        .columns(DataFactory.ND.multiY.columns)
+                        .data(DataFactory.ND.multiY.data)
+                    );
+                });
```
jbrundage: @GordonSmith Here is the API call for c3 column

jbrundage: @GordonSmith And here is the API call for AM column
https://github.com/hpcc-systems/Visualization/blob/master/test/amchartFactory.js#L48-L65

mlzummo:
@jbrundage is this the final code?

jbrundage:
@mlzummo I suppose the yAxis publish parameters should be applied to the y2 axis as well... though this differs from the AM Charts implementation

mlzummo:
You can add the axis obj logic to c3 charts

Issue #1005, opened 2015-10-07
Export Option on Table Widget (or any widget where export might be appropriate)


jprichard:
When the Table widget is displaying, since we use them consistently to filter and show details, many times the user wants to export to CSV to go over to another system (or EXCEL) to perform other tasks.

jchambers-ln:
Required for https://github.com/hpcc-systems/DSR/issues/3

PR #1251, opened 2015-12-04
GH-1005 Export Option on Table Widget

dtsnell4:
Adds an Exporter widget so that  any widget can export its data to CSV, TSV, or JSON formats, and save as a file.

Fixes GH-1005

### DIFF
--- a/src/common/Utility.js
+++ b/src/common/Utility.js
@@ -237,6 +237,53 @@
                     return node;
                 }
             };
+        },
+        downloadData: function (format, data, id) {
+            id = id || "data";
+            var currentdate = new Date(); 
+            var timeFormat =  d3.time.format("%Y-%m-%dT%H_%M_%S");
+            var now = timeFormat(currentdate);
+
+            var filename = id + "_" + now + "." + format.toLowerCase();
+            
+            var mimeType = "";
+            switch (format) {
+                case "TSV":
+                    mimeType = "text/tab-seperated-values";
+                    break;
+                case "JSON":
+                    mimeType = "application/json";
+                    break;
+                default:
+                    mimeType = "text/csv";
+            }
+
+            var a = document.createElement('a');
+            if (navigator.msSaveBlob) { // IE10+
+                a = null;
+                return navigator.msSaveBlob(new Blob([data], { type: mimeType }), filename);
+            } else if ('download' in a) { //html 5
+                a.href = 'data:' + mimeType + ',' + encodeURIComponent(data);
+                a.setAttribute('download', filename);
+                document.body.appendChild(a);
+                setTimeout(function() {
+                    a.click();
+                    document.body.removeChild(a);
+                }, 10);
+                return true;
+            } else { //old chrome and FF:
+                a = null;
+                var frame = document.createElement('iframe');
+                document.body.appendChild(frame);
+                frame.src = 'data:' + mimeType + ',' + encodeURIComponent(data);
+
+                setTimeout(function() {
+                    document.body.removeChild(frame);
+                }, 100);
+                return true;
+            }
+
+            return false;
         }
 
     };
--- a/src/common/Widget.js
+++ b/src/common/Widget.js
@@ -47,6 +47,21 @@
     Widget.prototype.mixin(Platform);
     Widget.prototype.mixin(PropertyExt);
 
+    Widget.prototype.export = function (_) {
+        var formattedData;
+        switch(_) {
+            case "TSV":
+                formattedData = this._db.tsv();
+                break;
+            case "JSON":
+                formattedData = this._db.json();
+                break;
+            default:
+                formattedData = this._db.csv();
+        }
+        return formattedData;
+    };
+
     Widget.prototype.leakCheck = function (newNode) {
         var context = this;
         var watchArray = [newNode];
--- a/src/layout/Surface.css
+++ b/src/layout/Surface.css
@@ -7,8 +7,6 @@
 .layout_Surface > h3 {
     margin: 0px;
     padding: 2px;
-    text-align: center;
-    font-weight: bold;
     background-color: #e5e5e5;
 }
 .layout_Surface .html-button-container {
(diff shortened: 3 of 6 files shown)


src/common/Exporter.js
```diff
@@ -0,0 +1,80 @@
...
+                formattedData = this._widget._db.json();
+                break;
+            default:
+                formattedData = this._widget._db.csv();
+        }
+        return formattedData;
+    };
+
+
+    Exporter.prototype.downloadData = function (format) {
```
GordonSmith: Can we move this into Utility?   We will pass in the pre-formatted data and a type.

src/common/Exporter.js
```diff
@@ -0,0 +1,80 @@
...
+        define([], factory);
+    } else {
+        root.common_Exporter = factory();
+    }
+}(this, function () {
+    function Exporter(widget) {
+        this._widget = widget;
+    }
+
+    Exporter.prototype.formatData = function (_) {
```
GordonSmith: Expose an "export" function in Widget.

src/common/Exporter.js
```diff
@@ -0,0 +1,80 @@
...
+        var mimeType = "";
+        switch (format) {
+            case "TSV":
+                mimeType = "text/tab-seperated-values";
+                break;
+            case "JSON":
+                mimeType = "application/json";
+                break;
+            default:
+                mimeType = "text/csv";
```
GordonSmith: What other mimeTypes are there (that might be interesting)?

src/common/Exporter.js
```diff
@@ -0,0 +1,80 @@
...
+        } else { //old chrome and FF:
+            var frame = document.createElement('iframe');
+            document.body.appendChild(frame);
+            frame.src = 'data:' + mimeType + ',' + encodeURIComponent(data);
+
+            setTimeout(function() {
+                document.body.removeChild(frame);
+            }, 100);
+            return true;
+            }
```
GordonSmith: Does this "trick" work on all browsers?  If not we will need to disable the export button if its not supported (and if it does I have a use for it in some places!)

dtsnell4:
http://www.sitepoint.com/web-foundations/mime-types-complete-list/

src/common/Exporter.js
dtsnell4: This should work on everything IE 9+.  I have IE 10, and used the dev tools to emulate IE9 and it works (but that is not always valid).

dtsnell4: @GordonSmith @mlzummo @jbrundage DEMO: http://rawgit.com/dtsnell4/Visualization/GH-1005_Exporter_Widget/demos/dermatology.html?src/common/Exporter

There are basically two functions, one to format the data and one to download it.

src/common/Exporter.js
dtsnell4: @GordonSmith Do you mean get rid of Exporter.js and just put the formatting logic in Widget.js as Widget.prototype.export() and the download function in Utility.js?

If that is not what you mean, please elaborate.

src/common/Exporter.js
dtsnell4: http://www.sitepoint.com/web-foundations/mime-types-complete-list/

dtsnell4:
@GordonSmith Do you mean get rid of Exporter.js and just put the formatting logic in Widget.js as Widget.prototype.export() and the download function in Utility.js?

If that is not what you mean, please elaborate.

src/common/Exporter.js
GordonSmith: Is that all<g>?

src/common/Exporter.js
GordonSmith: Yes I suspect that would be the end result of those two changes?  thoughts?

dtsnell4: That would be fine with me, I originally thought about just doing all of it in Widget.js, but wasn't sure if that would be acceptable. I will update it.

dtsnell4:
@GordonSmith updated per comments

src/common/Utility.js
```diff
@@ -237,6 +237,55 @@
                     return node;
                 }
             };
+        },
+        downloadData: function (format, data, id) {
+            // var data = this.export(format);
+
+            var a = document.createElement('a');
```
GordonSmith: This is not always used or removed.

src/common/Utility.js
```diff
@@ -237,6 +237,55 @@
...
+
+            var currentdate = new Date(); 
+            var now = currentdate.getDate() + "."+ 
+                (currentdate.getMonth()+1)  + "." + 
+                currentdate.getFullYear() + "_"  + 
+                currentdate.getHours() + ":"  + 
+                currentdate.getMinutes() + ":" + 
+                currentdate.getSeconds()
+            ;
```
GordonSmith: Use d3.time.format to format the date.

src/common/Utility.js
```diff
@@ -237,6 +237,55 @@
...
+                var frame = document.createElement('iframe');
+                document.body.appendChild(frame);
+                frame.src = 'data:' + mimeType + ',' + encodeURIComponent(data);
+
+                setTimeout(function() {
+                    document.body.removeChild(frame);
+                }, 100);
+                return true;
+            }
         }
```
GordonSmith: Should have a return false at the end?

GordonSmith:
@dtsnell4 Can you add a link to the test in this thread?

dtsnell4:
@GordonSmith http://rawgit.com/dtsnell4/Visualization/GH-1005_Exporter_Widget/demos/dermatology.html?src/common/Exporter

I put this in the "Common" menu when there was an Exporter.js file.  There is not now.  Should it go somewhere else?

src/common/Utility.js
```diff
@@ -237,6 +237,53 @@
                     return node;
                 }
             };
+        },
+        downloadData: function (format, data, id) {
+            id = id || "data";
+            var currentdate = new Date(); 
+            var timeFormat =  d3.time.format("%m-%d-%Y_%H.%M.%S");
```
GordonSmith: Can you switch this to:  ```%Y-%m-%dT%H:%M:%S``` (sorry for being picky) - but:
1.  Its sortable
2.  It matches time stamps we use elsewhere
3.  The "." in the time could mess with the file extension.

dtsnell4: @GordonSmith You can't have a colon : in a file name (or / \ | " * ? < > ).  That's why I went with the period, I've never known it to cause an issue on modern OSes (although it did back in the day). And I wanted to go with something different than the hyphen in the date.  Any other suggestions?

I will change the order though.

GordonSmith: How about:  ```%Y-%m-%dT%H_%M_%S```?

dtsnell4:
@GordonSmith Updated

PR #1256, opened 2015-12-07
GH-1255 Makes layout/Cell dragHandle ID-specific

jbrundage:
Fixes GH-1255

@GordonSmith @mlzummo @dtsnell4 Please Review

### DIFF
--- a/src/layout/Cell.js
+++ b/src/layout/Cell.js
@@ -41,7 +41,7 @@
         var context = this;
         var offsetMultiple;
 
-        var dragHandles = element.selectAll(".dragHandle").data(this._dragHandles, function (d) { return d; });
+        var dragHandles = element.selectAll("#"+this.id()+" > .dragHandle").data(this._dragHandles, function (d) { return d; });
         dragHandles.enter().append("div")
             .attr("class", function (d) { return "dragHandle dragHandle_" + d; })
             .style("position", "absolute")


src/layout/Cell.js
```diff
@@ -41,7 +41,7 @@
         var context = this;
         var offsetMultiple;
 
-        var dragHandles = element.selectAll(".dragHandle").data(this._dragHandles, function (d) { return d; });
+        var dragHandles = element.selectAll("#"+this.id()+" > .dragHandle").data(this._dragHandles, function (d) { return d; });
```
GordonSmith: We should be using this pattern more often - do you think the following would be worthwhile:
```
Widget.prototype.selectChildren = function(selector) {
    return this._element.selectAll("#" + this.id() + " > " + selector);
};
```
And then for completeness:
```
Widget.prototype.selectAll = function(selector) {
    return this._element.selectAll(selector);
};
```

jbrundage: I think the **selectChildren** is worthwhile, but for completeness did you mean to suggest this?
```javascript
Widget.prototype.selectAll = function(selector) {
    return this.selectChildren(selector);
});
```

GordonSmith: No it would be as is, just to replace```this._element.selectAll``` or ```this.element().selectAll``` or ```element.selectAll``` with a ```this.selectAll```

jbrundage:
@GordonSmith This is ready to be merged

PR #1268, opened 2015-12-09
GH-1258 Adds sub-prop table span to table head

jbrundage:
Fixes GH-1258

http://rawgit.com/jbrundage/Visualization/GH-1258/demos/dermatology.html?src/layout/Grid


jbrundage:
These simple changes really help in simplifying the PropertyEditor interface

mlzummo:
looks good but maybe we should use a space instead of an underscore in the names? and maybe make it customizable to an extent that we can make the displaying of all that not reflect stuff like other_PropertyEditor but just PropertyEditor

Issue #1267, opened 2015-12-09
common\Database.js hipieMapSortArray throws error when fieldByLabel returns undefined


jbrundage:
fieldByLabel returns undefined if this.fields() is an empty array

jbrundage:
https://github.com/hpcc-systems/Visualization/blob/master/src/common/Database.js#L39

GordonSmith:
It may be hipieMapSortArray should check that fields exist before trying to sort them or upstream should check if data has been loaded before trying to sort?

jbrundage:
That sounds reasonable, though I'm not sure what hipieMapSortArray is being used to sort

mlzummo:
ill take a look at that

mlzummo:
@GordonSmith i think this is fixed withyour fix

PR #1276, opened 2015-12-10
GH-1267 FILTER_BY_LABEL_DB_FIX

mlzummo:
Fixes the function from returning undefined (returns empty object instead)

Fixes GH-1267


GordonSmith:
I think I would prefer it to return null rather than undefined or an empty object so that the caller knows that nothing was found.

mlzummo:
@GordonSmith so we want to keep the commserror we get then?

GordonSmith:
Well that error probably means something?

mlzummo:
well the error i think occurs cause the idx its looking for isnt loaded until click ... its looking for {}.idx where {} is the object its suppose to return ... i just return an object instead of undefined and then the just the idx will be undefined which is ok in this case

GordonSmith:
The API for any field access is "get field by ..." where "..." could be "index" or "label" (or other options in the future).

If the field does not exist it *should* return null and not an empty object.

If this is causing an issue in other places, then the other places need to either be 100% sure that a field does exist or to test for its existence.

Issue #1287, opened 2015-12-14
MegaChart and Legend widgets need some improvements


jbrundage:
- [x] Legend should optionally display on any edge of MegaChart
- [x] Legend needs to support 2D charts
- [x] Legend should have vertical/horizontal display modes
- [x] MultiChart should be able to communicate what dataType its current chartType uses

PR #1288, opened 2015-12-14
GH-1287 Enhances Legend within MegaChart

jbrundage:
Fixes GH-1287

http://rawgit.com/jbrundage/Visualization/Legend_Imp/demos/dermatology.html?src/composite/MegaChart.pie

### DIFF
--- a/src/chart/MultiChart.js
+++ b/src/chart/MultiChart.js
@@ -29,7 +29,7 @@
     MultiChart.prototype._1DChartTypes = [
         { id: "SUMMARY", display: "Summary", widgetClass: "chart_Summary" },
         { id: "C3_GAUGE", display: "Gauge (C3)", widgetClass: "c3chart_Gauge" }
-    ];
+    ].map(function(item) { item.family = "1D"; return item;});
     MultiChart.prototype._2DChartTypes = [
         { id: "BUBBLE", display: "Bubble", widgetClass: "chart_Bubble" },
         { id: "PIE", display: "Pie", widgetClass: "chart_Pie" },
@@ -40,7 +40,7 @@
         { id: "AM_PIE", display: "Pie (amCharts)", widgetClass: "amchart_Pie" },
         { id: "AM_PYRAMID", display: "Area (amCharts)", widgetClass: "amchart_Pyramid" },
         { id: "WORD_CLOUD", display: "Word Cloud", widgetClass: "other_WordCloud" }
-    ];
+    ].map(function(item) { item.family = "2D"; return item;});
     MultiChart.prototype._NDChartTypes = [
         { id: "COLUMN", display: "Column", widgetClass: "chart_Column" },
         { id: "LINE", display: "Line", widgetClass: "chart_Line" },
@@ -61,10 +61,10 @@
         { id: "AM_BAR", display: "Bar (amCharts)", widgetClass: "amchart_Bar" },
         { id: "AM_LINE", display: "Line (amCharts)", widgetClass: "amchart_Line" },
         { id: "AM_SCATTER", display: "Scatter (amCharts)", widgetClass: "amchart_Scatter" },
-    ];
+    ].map(function(item) { item.family = "ND"; return item;});
     MultiChart.prototype._anyChartTypes = [
         { id: "TABLE", display: "Table", widgetClass: "other_Table" }
-    ];
+    ].map(function(item) { item.family = "any"; return item;});
     MultiChart.prototype._allChartTypes = MultiChart.prototype._1DChartTypes.concat(MultiChart.prototype._2DChartTypes.concat(MultiChart.prototype._NDChartTypes.concat(MultiChart.prototype._anyChartTypes)));
 
     MultiChart.prototype.publish("chartType", "BUBBLE", "set", "Chart Type", MultiChart.prototype._allChartTypes.map(function (item) { return item.id; }),{tags:["Basic"]});
@@ -114,6 +114,10 @@
         if (!arguments.length) return this._chartTypeProperties;
         this._chartTypeProperties = _;
         return this;
+    };
+    
+    MultiChart.prototype.getChartDataFamily = function () {
+        return this._allCharts[this.chartType()].family;
     };
 
     MultiChart.prototype.requireContent = function (chartType, callback) {
--- a/src/composite/MegaChart.js
+++ b/src/composite/MegaChart.js
@@ -41,7 +41,7 @@
     MegaChart.prototype.publishProxy("valueAxisTitle","_valueTitle","text");
     MegaChart.prototype.publishProxy("domainAxisTitle","_domainTitle","text");
     
-    MegaChart.prototype.publish("showLegend",false,"boolean","Show/Hide Legend", null, {tags:["Basic"]});
+    MegaChart.prototype.publish("legendPosition","right","set","Position of the Legend widget", ["none","top","right","bottom","left"], {tags:["Basic"]});
     
     MegaChart.prototype.publish("gutter", 0, "number", "Gap Between Cells",null,{tags:["Basic"]});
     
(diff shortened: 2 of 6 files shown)


src/chart/MultiChart.js
```diff
@@ -115,6 +115,19 @@
...
+        var types = ["1D","2D","ND"];
+        for(var i=0;i<types.length;i++) {
+            var idx = "_"+types[i]+"ChartTypes";
+            for(var j=0;j<context[idx].length;j++) {
+                if(context[idx][j].id === context.chartType()){
+                    return types[i];
+                }
+            }
+        }
+    };
```
GordonSmith: 1.  Knowing what data "family" a widget belongs to is a useful thing in general I suspect (I will open an issue for this).
2.  Double loops are always candidates for performance bottlenecks and should be avoided

At line 16 add something along the lines of:
```
newItem.chartFamily = this._1DChartType.indexOf(item) >= 0 ? "1D" : this._2DChartType.indexOf(item) >= 0 ? "2D" : this._NDChartType.indexOf(item) >= 0 ? "ND" : "???";
```
And then use that.

GordonSmith: or...
Change the prototypes to include the "family" (either by hand) or by:
```
MultiChart.prototype._1DChartTypes = [
        { id: "SUMMARY", display: "Summary", widgetClass: "chart_Summary" },
        { id: "C3_GAUGE", display: "Gauge (C3)", widgetClass: "c3chart_Gauge" }
    ].map(function(item) { item.family = "1D"; return item;});
```
Type of thing?

src/composite/MegaChart.js
```diff
@@ -52,41 +53,110 @@
...
     };
+    
+    MegaChart.prototype.setLegend = function(){
+        if(this.showLegend() && (this._layout.getContent(this.legendPosition()) === null || this._layout.getContent(this.legendPosition()).classID() !== "other_Legend")){
+            this._legend = new Legend().fixedSize(true).targetWidget(this._layout.getContent("center"));
+            this._layout.setContent(this.legendPosition(), this._legend);
+        } else if (!this.showLegend()) {
+            this._layout.clearContent(this.legendPosition());
+        }
+    };
```
GordonSmith: Looks like you create a new legend every time it relocates?  If so you should remove the previous instance first (.target(null)) to avoid future leaks (probably not leaking in this case), but will need more work if the legend is to be shown on the dermatology page?

src/composite/MegaChart.js
```diff
@@ -52,41 +53,110 @@
...
+    
+    MegaChart.prototype.removeExtraLegends = function () {
+        for(var i in this._contentClasses){
+            if(this._contentClasses[i] === "other_Legend" && i !== this.legendPosition()){
+                this._layout.clearContent(i);
+            }
+        }
+    };
```
GordonSmith: Ok this addresses my previous comment.

I would prefer is you simply used a "prevLegendPosition" pattern:

```
if (this._prevLegendPosition !== this.legendPosition()) {
  //remove old and reposition it to new location (or create a new one)
  this._prevLegendPosition = this.legendPosition();
}
```
Also you could do that inside the update method.  

Thought:  You could remove the "showLegend PP completely and include a legendPosition of "none"?

src/other/Legend.js
```diff
@@ -16,24 +16,39 @@
...
-                        ]);
-                    }
+                switch(this.dataType()){
+                    case '2D':
+                        for(var i in widget.data()){
+                            dataArr.push([
+                                _htmlColorBlock(widget._palette(widget.data()[i][0])),
+                                widget.data()[i][0]
+                            ]);
+                        }
```
GordonSmith: would be cleaner as a ```dataArr = widget.data().map(...)```

src/other/Legend.js
```diff
@@ -16,24 +16,39 @@
...
-                for(var i in widgetColumns){
-                    if(i > 0){
-                        dataArr.push([
-                            _htmlColorBlock(widget._palette(widgetColumns[i])),
-                            widgetColumns[i]
-                        ]);
-                    }
+                switch(this.dataType()){
+                    case '2D':
+                        for(var i in widget.data()){
```
GordonSmith: data() is an array and using the "in" for looping through it is bad practice.

src/other/Legend.js
```diff
@@ -16,24 +16,39 @@
...
+                            dataArr.push([
+                                _htmlColorBlock(widget._palette(widget.data()[i][0])),
+                                widget.data()[i][0]
+                            ]);
+                        }
+                        break;
+                    case 'ND':
+                        for(var j in widgetColumns){
```
GordonSmith: Don't use "in" to loop through arrays.

src/other/Legend.js
```diff
@@ -16,24 +16,39 @@
...
+                            if(j > 0){
+                                dataArr.push([
+                                    _htmlColorBlock(widget._palette(widgetColumns[j])),
+                                    widgetColumns[j]
+                                ]);
+                            }
+                        }
```
GordonSmith: should be recoded as a ```widgetColumns.filter(...).map(...)```

src/chart/MultiChart.js
jbrundage: Is it really a candidate for a performance bottleneck?

It does make more sense to move the logic to the location that the chart types are specified (...to prevent problems with future additions).

src/composite/MegaChart.js
jbrundage: clearContent sets the old legend widget target to null

src/composite/MegaChart.js
jbrundage: This makes sense. I'll do this now.

src/other/Legend.js
jbrundage: Agreed

src/other/Legend.js
jbrundage: Good catch, I thought I was beyond that mistake

jbrundage:
@GordonSmith All set, ready to be re-reviewed. What do you think about the **lazyRender** in Border layout? I had to implement it like that to allow the shirnkWrap to apply appropriately. Any suggestions regarding a better implementation?

src/chart/MultiChart.js
```diff
@@ -115,6 +115,15 @@
...
     };
+    
+    MultiChart.prototype.getChartDataFamily = function () {
+        var context = this;
+        for(var i=0;i<context._allChartTypes.length;i++) {
+            if(context._allChartTypes[i].id === context.chartType()){
+                return context._allChartTypes[i].family;
+            }
+        }
+    };
```
GordonSmith: This can now be:  ```return this._allCharts[context.chartType()].family;```

src/composite/MegaChart.js
```diff
@@ -52,41 +52,93 @@
...
-                    .data(this.data());
+        if(this.legendPosition() === "none" && this._prevLegendPosition !== this.legendPosition()){
+            if(this._prevLegendPosition !== undefined){
+                this._layout.clearContent(this._prevLegendPosition);
+            }
+        }
+        if (this.legendPosition() !== "none" && this._prevLegendPosition !== this.legendPosition()) {
+            if(this._prevLegendPosition !== undefined){
+                this._layout.clearContent(this._prevLegendPosition);
+            }
```
GordonSmith: 89->97 could be improved?

```
if (this._prevLegendPosition !== this.legendPosition()) {
  if (this._prevLegendPosition && this._prevLegendPosition !== "none") {
    //  remove
  }
  if (this.legendPosition() !== "none") {
    //  add
  }
  this._prevLegendPosition = this.legendPosition();
}
```

src/composite/MegaChart.js
```diff
@@ -52,41 +52,93 @@
...
-            this._layout.setContent("right", new Legend().fixedSize(true).targetWidget(this._layout.getContent("center")));
-        } else {
-            this._layout.clearContent("right");
+        
+        if(this._legend !== null){
+            if(["top","bottom"].indexOf(this.legendPosition()) !== -1){
+                this._legend.orientation("horizontal");
+            } else {
+                this._legend.orientation("vertical");
+            }
```
GordonSmith: I would move this block into the ```//add``` above - it only needs to be set once.

src/layout/Border.js
```diff
@@ -51,6 +51,28 @@
...
+                if (oldVal !== newVal) {
+                    context.lazyRender();
+                }
+            });
+        }
+    };
+    
+    Border.prototype.lazyRender = Border.prototype.debounce(function(){
+        this.postUpdate();
+    },0);
```
GordonSmith: You will want a debounce timeout > 0 otherwise the debounce doesn't do its thing...

GordonSmith:
@jbrundage bump

GordonSmith:
@jbrundage Can you rebase and I will pull...

dtsnell4:
Looks good to me

jbrundage:
@GordonSmith Rebased

PR #1296, opened 2015-12-16
Remove "theme" support from Publish Params.

GordonSmith:
This feature was not used.

### DIFF
--- a/src/common/PropertyExt.js
+++ b/src/common/PropertyExt.js
@@ -150,39 +150,26 @@
         }
         var meta = this[__meta_ + id] = new Meta(id, defaultValue, type, description, set, ext);
         this[id] = function (_) {
-            var isPrototype = this._id === undefined;
             if (!arguments.length) {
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jbrundage Please Review

jbrundage:
@GordonSmith I think the themes could play a big part in the acceptance of this framework in the long run. The idea was (originally) to allow themes to be the take-away from the "dermatology layer" and applied within an app to provide styling. I was using a theme within the Avenger app, but Claudio preferred to configure the publish parameter values from within the DSP.

**For apps that don't use DSP** a theme is what will allow a non-technical person style their charts.

GordonSmith:
@jbrundage I agree 100% - but this change does not affect your theme support and the way you have implemented it (which is why I am removing it) - the PR title could be better...

GordonSmith:
@jbrundage Please Review (see above comments)

jbrundage:
@GordonSmith It seems to work well enough. 
Styles are no longer reset to their default values when switching between themes, but that isn't necessarily an issue.

Issue #1336, opened 2016-01-11
DSP savelayout in 1.10 breaks deserilization


mlzummo:

mlzummo:
pulling in my 2 fixes partially fixes the issue

mlzummo:
having issues deserialzing the graph .... its matching the visualization.id and trying to reuse however ... in hippieddl function getVertex(item, origItem) { ... since its trying to reuse the graph object "vertex" and "edge" are never defined and it cant create new ones becuase vertexmap is never used ... i recommend we put a require in that section ... but @GordonSmith i'd like your input on that one ... if i try to tell it to skip reusing the persisted one it just flat out doesnt work

PR #1338, opened 2016-01-12
GH-1336 MARSHALLER_PERSIST_GRAPH_FIX

mlzummo:
Allows the graph to persist and recreate new vertexes/edges

Fixes GH-1336

### DIFF
--- a/src/graph/Graph.js
+++ b/src/graph/Graph.js
@@ -22,6 +22,9 @@
     Graph.prototype.constructor = Graph;
     Graph.prototype._class += " graph_Graph";
     Graph.prototype.implements(IGraph.prototype);
+    
+    Graph.prototype.Vertex = Vertex;
+    Graph.prototype.Edge = Edge;
 
     Graph.prototype.publish("allowDragging", true, "boolean", "Allow Dragging of Vertices", null, { tags: ["Advanced"] });
     Graph.prototype.publish("layout", "Circle", "set", "Default Layout", ["Circle", "ForceDirected", "ForceDirected2", "Hierarchy", "None"], { tags: ["Basic"] });
--- a/src/marshaller/HipieDDL.js
+++ b/src/marshaller/HipieDDL.js
@@ -6,8 +6,6 @@
         root.marshaller_HipieDDL = factory(root.d3, root.common_Database, root.common_Utility, root.other_Comms, root.common_Widget, root.require);
     }
 }(this, function (d3, Database, Utility, Comms, Widget, require) {
-    var Vertex = null;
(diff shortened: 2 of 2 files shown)


mlzummo:
@GordonSmith @jbrundage Please Review (i can show you what this fixes ... gordon if you have a better solution let me know)

GordonSmith:
@mlzummo - i am pretty sure this will work fine (as the Vertex + Edge are loaded if the Graph has been loaded) - but it might be simpler to just add the following to Graph.js:

```
Graph.prototype.Vertex = Vertex;
Graph.prototype.Edge = Edge;
```

And the we can simply do:
```
var v = new graph.Vertex();
```
or
```
var v = new Graph.prototype.Vertex();
```

mlzummo:
@GordonSmith but how will we access this graph object from within GraphMapping.prototype.doMapAlll which is where we need this?

mlzummo:
i think i see a way one sec

mlzummo:
@GordonSmith made suggested changes

mlzummo:
@GordonSmith can you pull this?

src/graph/Graph.js
```diff
@@ -558,9 +561,6 @@
         edgeElements.enter().append("g")
             .attr("class", "graphEdge")
             .style("opacity", 1e-6)
-            .on("click.selectionBag", function (d) {
-                context._selection.click(d, d3.event);
-            })
```
GordonSmith: Why is this removed?

GordonSmith:
Can you clean up the other places where we were using the global Vertex and global Edge?

src/graph/Graph.js
mlzummo: idk i was getting something weird with that ... where it looked like i needed it to be removed .... but im going to put it back

mlzummo:
@GordonSmith other places in the marshaller or in general throughout the lib?

GordonSmith:
These items can be removed and then replaced with the new declarations:
https://github.com/mlzummo/Visualization/blob/MARSHALLER_PERSIST_GRAPH_FIX/src/marshaller/HipieDDL.js#L9-L10

src/graph/Graph.js
```diff
@@ -558,8 +561,8 @@
         edgeElements.enter().append("g")
             .attr("class", "graphEdge")
             .style("opacity", 1e-6)
-            .on("click.selectionBag", function (d) {
```
mlzummo: this is why i removed it in the first place idk why its red over here

mlzummo:
@GordonSmith should be ready to pull now

Issue #1342, opened 2016-01-12
fieldbyLabel returns undefined causing comms error


mlzummo:
when it cant find an idx that hasn't been pulled yet from an ajax datasource

PR #1343, opened 2016-01-12
GH-1342 DB_FIELDBYLABEL_FIX

mlzummo:
Returns empty object and console log if it can't find index

Fixes GH-1342

### DIFF
--- a/src/common/Database.js
+++ b/src/common/Database.js
@@ -183,7 +183,13 @@
     };
 
     Grid.prototype.fieldByLabel = function (_, ignoreCase) {
-        return this.fields().filter(function (field, idx) { field.idx = idx; return ignoreCase ? field.label().toLowerCase() === _.toLowerCase() : field.label() === _; })[0];
+        var retVal = this.fields().filter(function (field, idx) { field.idx = idx; return ignoreCase ? field.label().toLowerCase() === _.toLowerCase() : field.label() === _; })[0];
+        if (!retVal) {
+            console.log("Not able to find field: " + _);
+            console.log(this.fields());
+            return {};
+        }
+        return retVal;
     };
 
     Grid.prototype.data = function (_, clone) {


mlzummo:
we could also add some exceptions in comms for certain errors (this was specific to something @jbrundage was working on) ...Jay can you link us to that page?

jbrundage:
It was on AVP an internal app. I can't link to it from here.  

Your code change patches the error, but it may be appropriate to return *undefined*. The real bug here is that the response from fieldByLabel is assumed to be an object, but it should be checked before accessed.

mlzummo:
@GordonSmith looks like since the function is used in a lot of places this is prolly the best way to fix this ... if you would like the console logs to display diff let me know other than that ... can we pull?

src/common/Database.js
```diff
@@ -183,7 +183,13 @@
     };
 
     Grid.prototype.fieldByLabel = function (_, ignoreCase) {
-        return this.fields().filter(function (field, idx) { field.idx = idx; return ignoreCase ? field.label().toLowerCase() === _.toLowerCase() : field.label() === _; })[0];
+        var retVal = this.fields().filter(function (field, idx) { field.idx = idx; return ignoreCase ? field.label().toLowerCase() === _.toLowerCase() : field.label() === _; })[0];
+        if (!retVal) {
+            console.log("Not able to find field: " + _);
+            console.log(this.fields());
```
GordonSmith: The console log messages do not belong here - they belong where the call is being made.

src/common/Database.js
```diff
@@ -183,7 +183,13 @@
...
+            return {};
```
GordonSmith: Should return null (which it was) if there was no field found.

mlzummo: then we need to change every place that function is used so that when its looking for {}.idx it throws a console log and does a check ... just seems a bit redundant when we can handle it in one place

GordonSmith: You need to do that for all the places which request a field that might not exist yes.  You can put this here - but only inside a __hppcDebug__

GordonSmith: You could add a method "fieldExists" if you like.

mlzummo: @GordonSmith we need to fix the DDL/ or something then so that a field isnt requested just yet before its fetched? basically we dont want to be throwing an error ... the error is fixed after the data and index is fetched ... @jbrundage can you email us the DDL this fails on?

GordonSmith:
```
var field = fieldByLabel(...)
if (field) {
...
}
```

mlzummo: //  Hipie Helpers  ---
    Grid.prototype.hipieFieldsByMappings = function (mappings) {
        return mappings.map(function (mapping) {
            return this.fieldByLabel(mapping, true);
        }, this);
    };

    Grid.prototype.hipieFieldIdxByMappings = function (mappings) {
        return this.hipieFieldsByMappings(mappings).map(function (field) {
            return field.idx;
        });
    };


the problems are happening here ... and its looking for undefined.idx

mlzummo: mapping.params.forEach(function (params) {
                            rollupValueIdx.push(this.fieldByLabel(params.param1, true).idx);
                        }, this);

mlzummo: var idx = this.fieldByLabel(params.param1, true).idx;

mlzummo: im not sure how to handle this for the marshaller

GordonSmith:
Closing in favor of:  https://github.com/hpcc-systems/Visualization/pull/1369

PR #1389, opened 2016-01-23
GH-1388 Prevent duplicate PP serialization

GordonSmith:
Fixes GH-1388


GordonSmith:
@mlzummo @jbrundage Please Review

jbrundage:
@GordonSmith Great, looks good

mlzummo:
@GordonSmith looks good... but i have a few things im doing right now... what cuased us to have to do this?

mlzummo:
in ext we could set it so we could override props or does this compeltely delete all props?

GordonSmith:
In this case the same widget was being "published" by more than one PP.

PR #1394, opened 2016-01-23
Add bundles definition file to assist dynamic loading.

GordonSmith:


GordonSmith:
@jbrundage @mlzummo Please Review.

This adds an extra file during the build process which will allow us to load 1.6.10 and 1.8.4 and 1.10.0 at the same time (which in turn will allow DSP to select which version of the framework to use at runtime...) - I will be adding a separate loader function shortly...

mlzummo:
looks good... but @GordonSmith can you explain the usage of this in a bit more detail to us ... such as this loader function?

GordonSmith:
I will have a worked example shortly - this just makes life a bit easier (which is why its targeting 1.6.10)

Issue #1306, opened 2015-12-28
The Maps for ACC need pins that can have icons/text applied to them


jbrundage:

GordonSmith:
@jbrundage - @mlzummo discussed this a bit before I went on vacation.  There are some "speed" considerations especially with the new maps, as they can zoom + scroll without re-rendering for the most part!  The new pins will need to do the same.

jbrundage:
@GordonSmith Take a look at my code change and example in #1307. If we keep these things rendering in SVG it should address your above concerns.

mlzummo:
need to add support for multiline text

jbrundage:
@GordonSmith Can we target a closer milestone for this issue? I'll need it in the first steps of integration with ACC

GordonSmith:
FYI 1.12.0-rc1 was released today and it only has one change the dynamic loader.

jbrundage:
@GordonSmith This is the needed enhancement we discussed on the phone this morning.

PR #1395, opened 2016-01-25
GH-1306 ACC Pins

mlzummo:
Adds Pablo's SVG pins to src/map/Pins.js
Fixes GH-1306

Signed-off-by: jay <mlzummo@gmail.com>


mlzummo:
@GordonSmith the code for doing auto text wrapo is what was in text.js you mentioned in the other PR ... however if we want to leave it so that wrap only happens on \n and not based on width let me know ... i messed with it a bit ... but i couldnt get the width calcs correct so it was being a pain... so let me know

@jbrundage ... thoughts

src/map/Pins.js
```diff
@@ -29,47 +33,143 @@
...
-                var pos = base.project(d[0], d[1]);
-                if (!pos) {
-                    return "none";
+        this.pinsPaths.enter().append("g").each(function(d) {
+
+            d3.select(this).style("opacity", d[2] && d[2].opacity ? d[2].opacity : context.opacity()); // might need to rework this from where it was using pinstransform?
+
+            if (!d[2] || d[2].pinShape !== "none") {
+                var path = d3.select(this).append("path")
+                    .attr("class", "data")
```
mlzummo: @GordonSmith is this really such a good class name for this .... maybe pin-path or pin-shape ?

src/map/Pins.js
```diff
@@ -29,47 +33,143 @@
...
+                    })
+                    .style("stroke", function (d) {
+                        if (d[2] && d[2].strokeColor) {
+                            return d[2].strokeColor;
+                        }
+                        return d[2] && d[2].strokeColor ? d[2].strokeColor : context.pinColor();
+                    })
+                ;
+
+                //var pathBBox = path.node().getBBox();
```
mlzummo: im leavin this for now if thats ok?

GordonSmith:
@mlzummo In the future I am going to auto-close any duplicate PRs like this, its just too painful having switch back and forwards all the time.
I would suggest that It would be far simpler if you just kept committing to the same branch - having lots of commits won't be an issue until we are ready to do the pull (IOW the comment will be "please rollup to single commit to be puller)?

mlzummo:
ok sounds good

GordonSmith:
@mlzummo @jbrundage What is the status of this PR?

GordonSmith:
Please re-submit PR for 1.14.0

PR #1446, opened 2016-02-03
GH-1445 AMCHARTS_PIE_LABEL_FIX

mlzummo:
(labels wont get clipped)

Fixes GH-1445

### DIFF
--- a/src/amchart/Pie.js
+++ b/src/amchart/Pie.js
@@ -58,13 +58,12 @@
         this._chart.type = "pie";
         
         this._chart.labelsEnabled = true;
-
         if (this.labelPosition()==="inside") {
             this._chart.radius = "50%";
             this._chart.labelRadius = -40;
             this._chart.pullOutRadius = "20%";
         } else {
-            this._chart.radius = "45%";
+            this._chart.radius = this.calcRadius();
             this._chart.labelRadius = 20;
             this._chart.pullOutRadius = "20%";
         }
@@ -72,10 +71,11 @@
         this._chart.labelFunction = function(d) {
             return d.title;
         };
-        if (this.marginRight()) { this._chart.marginRight = this.marginRight(); }
-        if (this.marginLeft()) { this._chart.marginLeft = this.marginLeft(); }
-        if (this.marginTop()) { this._chart.marginTop = this.marginTop(); }
-        if (this.marginBottom()) { this._chart.marginBottom = this.marginBottom(); }
+
+        this._chart.marginRight = this.marginRight();
+        this._chart.marginLeft = this.marginLeft();
+        this._chart.marginTop = this.marginTop();
+        this._chart.marginBottom = this.marginBottom();
 
         this._chart.depth3D = this.Depth3D();
         this._chart.angle = this.Angle3D();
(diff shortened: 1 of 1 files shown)


mlzummo:
@GordonSmith @jbrundage - Please Review

mlzummo:
http://rawgit.com/mlzummo/Visualization/AMCHARTS_PIE_LABEL_FIX_3/demos/dermatology.html?src/amchart/Pie

mlzummo:
currWidget.data([["Some Label",45,55,52],["Science",66,60,72],["Geography",75,68,65],["Here is a super long string of text",98,92,90]]).render()    with long labes if the width of the chart is not big enough will still get clipped (but the pie chart need to stay center ... ) there may be a way to set the poistion of pie wedges to fix this?

src/amchart/Pie.js
```diff
@@ -191,10 +191,75 @@
 
         this._chart.validateNow();
         this._chart.validateData();
+        console.log("width: " + this.width());
+        console.log("height: " +this.width());
```
GordonSmith: remove console logging.

src/amchart/Pie.js
```diff
@@ -191,10 +191,75 @@
...
+         if (leftOffset < 0){
+            smallerOffset += leftOffset;
+         }
+
+         if (this.calcRadius() + smallerOffset - 20 < this.calcRadius()) {
+            this._chart.radius = this.calcRadius() + smallerOffset - 20; // 20 for the label line lengths
+         }
+
+         this._chart.validateNow();
+         this._chart.validateData();
```
GordonSmith: You should only need to validate _if_ the _chart.radius is changed?

src/amchart/Pie.js
```diff
@@ -191,10 +191,75 @@
...
+            var labelLeftPos = d3.select(this).node().getBoundingClientRect().left;
+            if (labelLeftPos < containerBoundingClientRect.left) {
+                left.push(labelLeftPos - containerBoundingClientRect.left);
+            }
+        });
+
+         var topOffset = top.length ? Math.min.apply(Math, top) : 0;
+         var bottomOffset = bottom.length ? Math.min.apply(Math, bottom) : 0;
+         var rightOffset = right.length ? Math.min.apply(Math, right) : 0;
+         var leftOffset = left.length ? Math.min.apply(Math, left) : 0;
```
GordonSmith: Is ```Math.min.apply(Math, left)``` a safe hack?  while arguments are Array "like" they are not quite the same thing...

src/amchart/Pie.js
```diff
@@ -191,10 +191,75 @@
...
+
+            }
+
+            var labelRightPos = d3.select(this).node().getBoundingClientRect().right;
+            if (labelRightPos > containerBoundingClientRect.right) {
+                right.push(containerBoundingClientRect.right - labelRightPos);
+            }
+
+
+            var labelLeftPos = d3.select(this).node().getBoundingClientRect().left;
```
GordonSmith: Only call "getBoundingClientRect" once at the top of the each (its not a "cheap" operation) - also I typically use ```getBBox``` for SVG nodes - I didn't think the bounding rect worked with them (I am not saying its wrong - just thinking out loud).

mlzummo: getBoundingClientRect gets a def value for me then the BBox i believe

mlzummo: and i need to get all the values

src/amchart/Pie.js
```diff
@@ -72,10 +71,11 @@
         this._chart.labelFunction = function(d) {
             return d.title;
         };
-        if (this.marginRight()) { this._chart.marginRight = this.marginRight(); }
-        if (this.marginLeft()) { this._chart.marginLeft = this.marginLeft(); }
-        if (this.marginTop()) { this._chart.marginTop = this.marginTop(); }
-        if (this.marginBottom()) { this._chart.marginBottom = this.marginBottom(); }
+
```
mlzummo: @GordonSmith is this ok to do this way? cause margin can be 0 which can relate to false or should i do if this.marginRight() !== false (but idk how the publish param handles a null/blank etc or i figured i would test for all

mlzummo:
@GordonSmith i think this is ready

src/amchart/Pie.js
GordonSmith: How about ```if (!isNaN(this.marginRight()))```?

src/amchart/Pie.js
GordonSmith: Yes but you only need to call it once not four times.

src/amchart/Pie.js
```diff
@@ -195,6 +195,66 @@
...
+        var containerBoundingClientRect = context._element.select("svg").node().getBoundingClientRect(); // might need to change this selection (could get a bit more accurate results?)
+        this.d3LabelSelection = element.selectAll(".amcharts-pie-label");
+
+        var left = [];
+        var right = [];
+        var top = [];
+        var bottom = [];
+
+        this.d3LabelSelection.each(function(d, i) {
+            var labelTopPos = d3.select(this).node().getBoundingClientRect().top;
```
GordonSmith: change to:  ```var boundingRect = d3.select(this).node().getBoundingClientRect();```

PR #1441, opened 2016-02-02
GH-1399 C3Charts can corrupt

GordonSmith:
If render is called while the previous render is still "transitioning", then the corruption can occur.

Fixes GH-1399

Signed-off-by: Gordon Smith <gordon.smith@lexisnexis.com>

Issue #1399, opened 2016-01-25
Any chart with a transition can get corrupted if called in quick succession  


mlzummo:

GordonSmith:
When you day double data - do you mean duplicate rows?

mlzummo:
@GordonSmith will be looking into this

GordonSmith:
Reopening this issue as its not just C3 Pie that is affected - its potentially any chart with a transition.

PR #1473, opened 2016-02-06
Revert "GH-1399 C3Charts can corrupt"

GordonSmith:
Reverts hpcc-systems/Visualization#1441


mlzummo:
@GordonSmith what broke, that makes you want to revert this?

GordonSmith:
Nothing broke - its just that I can confirm that any D3 based chart which is using transitions is prone to this problem, so I going to do a broader "fix".

mlzummo:
looks good

PR #1482, opened 2016-02-08
GH-1399 Prevent marshaller from rendering during D3 transition

GordonSmith:
Fixes GH-1399

### DIFF
--- a/src/marshaller/HipieDDL.js
+++ b/src/marshaller/HipieDDL.js
@@ -1017,9 +1017,13 @@
                 delete this.request[key];
             }
         }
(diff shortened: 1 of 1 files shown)


GordonSmith:
@mlzummo Please Review

mlzummo:
looks good ... so if im reading this right ... it waits at least 500 milliseconds and makes sure the marshaller did not load everything that quick

GordonSmith:
Yes - in the event that the results come back < 500ms (which is slightly more than the default transition time - I should double check that) it will ensure it waits the full 500ms before calling the callback.

Issue #1517, opened 2016-02-24
DSP - Remove time pattern custom option after saving the layout


prakash-sukumar:
Steps to reproduce
1. Create a widget with time (apply time format)in chart attributes and run the dashboard
![screenshot 14](https://cloud.githubusercontent.com/assets/7846016/13278362/ef7bddae-daf5-11e5-964a-be51b9f909c9.png)
![screenshot 15](https://cloud.githubusercontent.com/assets/7846016/13278372/007da29a-daf6-11e5-81f0-cbcb6bdae990.png)
2.Change the position of the widget and save the layout
![screenshot 16](https://cloud.githubusercontent.com/assets/7846016/13278390/27418a7c-daf6-11e5-885e-06ddf0b51d2f.png)
3. Edit the widget and delete the custom time pattern given in chart attributes
![screenshot 17](https://cloud.githubusercontent.com/assets/7846016/13278401/45e5826c-daf6-11e5-9872-50484b627e60.png)
4.The custom options for the widget are restored from dermatology, which is not expected.
![screenshot 18](https://cloud.githubusercontent.com/assets/7846016/13278438/7af5a716-daf6-11e5-8285-879ed928b30c.png)

GordonSmith:
@prakash-sukumar - can you show include the DDL before and after the %c has been removed in this issue?  
Note:  This issue will be true for all "_"

prakash-sukumar:
**Sample DDL before**
https://216.19.105.2:18010/WsWorkunits/WUResult.json?Wuid=W20160225-023844&ResultName=testtimeonemore_Comp_Ins012_DDL&SuppressXmlSchema=true

{"WUResultResponse": {"Wuid": "W20160225-023844", "Sequence": 0, "LogicalName": null, "Cluster": null, "Name": "testtimeonemore_Comp_Ins012_DDL", "Start": 0, "Requested": 1, "Count": 1, "Total": 1, "Result": {
"testtimeonemore_Comp_Ins012_DDL": {"Row": [
[...]

PR #1521, opened 2016-02-24
GH-1517 "_" properties should be set as default values

GordonSmith:
Fixes GH-1517


GordonSmith:
@prakash-sukumar This will need some testing, before I pull it into 1.10.2

PR #1524, opened 2016-02-26
Fixes GH-1523 - Sorts data() with dataObjArr

jbrundage:
Fixes GH-1523

I added this code change to the IDM/AVP/DASH application accessed from within Accurint.

@GordonSmith Please review

### DIFF
--- a/src/amchart/CommonSerial.js
+++ b/src/amchart/CommonSerial.js
@@ -252,6 +252,9 @@
             dataObjArr.sort(function (a, b) {
                 return a[sortField] - b[sortField];
             });
+            this.data().sort(function (a, b) {
+                return a[0] - b[0];
+            });
         }
 
         return dataObjArr;


src/amchart/CommonSerial.js
```diff
@@ -252,6 +252,9 @@
             dataObjArr.sort(function (a, b) {
                 return a[sortField] - b[sortField];
             });
+            this.data().sort(function (a, b) {
```
mlzummo: i think if we just sort the data before its used at the top of the function we can eliminate sorting twice?

src/amchart/CommonSerial.js
```diff
@@ -252,6 +252,9 @@
...
+                return a[0] - b[0];
+            });
```
GordonSmith: I can't see how this will work, when your dates look like "03 Jan 2015", "01 July 2016" etc...

GordonSmith:
Closing in favor of:  https://github.com/hpcc-systems/Visualization/pull/1527

Issue #1546, opened 2016-03-02
Grid Layout designMode needs ability to prevent cell dragging unless surfaceTitle is grabbed


jbrundage:
Widgets currently can't be interacted with while in designMode. Adding a publish param to specify whether or not cell dragging should be draggable only by the title bar will allow the best of both worlds.

GordonSmith:
@jbrundage The PR needs to use the title bar in the MegaChart not Surface.

GordonSmith:
Closed "timed-out", please reopen if needed.

PR #1547, opened 2016-03-02
GH-1546 Adds Grid moveByTitleOnly

jbrundage:
Fixes GH-1546

@GordonSmith @mlzummo Please test/review

http://rawgit.com/jbrundage/Visualization/grid_title_drag/demos/dermatology.html?src/layout/Grid.simple&designMode=true&moveByTitleOnly=true


GordonSmith:
With the mega chart we don't really use the cell title anymore:  
http://rawgit.com/jbrundage/Visualization/grid_title_drag/demos/dermatology.html?src/marshaller/HTML.databomb&designMode=true&moveByTitleOnly=true&surfacePadding=0

I think you need to add an extra region to drag (similar to the resize bars) but at the top (and it should be shaded to make it obvious?  I don't think I would make this optional either?

PR #1683, opened 2016-05-13
Temp PR for comments

GordonSmith:


src/kendo/Bar.js
```diff
@@ -0,0 +1,95 @@
+"use strict";
+(function (root, factory) {
+    if (typeof define === "function" && define.amd) {
+        define(["jquery","kendo.all.min","../common/HTMLWidget"], factory);
+    } else {
+        root.kendo_Bar = factory();
+    }
+}(this, function (jquery, kendo, HTMLWidget) {
```
GordonSmith: Is JQuery actually needed?  Or is it just a dependency of kendo (if so we can probably move this to a shim, or let bower/kendo handle it)?

src/kendo/Bar.js
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+    function Bar() {
+        HTMLWidget.call(this);
+
+        var context = this;
+        this._tag = "div";
+        this._type = "unknown";
+        this._chartType = "BarChart";
```
GordonSmith: ```this._type, this._chartType```  Aren't really needed at this point (but you may end up wanting something similar in a Kendo base class when you add more charts)...

src/kendo/Bar.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+    Bar.prototype = Object.create(HTMLWidget.prototype);
+    Bar.prototype.constructor = Bar;
+    Bar.prototype._class += " kendo_Bar";
+
+
+    Bar.prototype.init = function (domNode, element) {  };
```
GordonSmith: Not needed

src/kendo/Bar.js
```diff
@@ -0,0 +1,95 @@
...
+
+
+    Bar.prototype.update = function (domNode, element) {
+
```
GordonSmith: You should call the base class update at this point:  ```HTMLWidget.prototype.update.apply(this, arguments);```

src/kendo/Bar.js
```diff
@@ -0,0 +1,95 @@
...
+                        template: "#= series.name #: #= value #"
+                    }
+                };
+
+                $("#"+domNode.id).kendoChart(kendoChartOptions);
+
+        };
+
+        $(document).ready(create);
+        $(document).bind("kendo:skinChange", create);
```
GordonSmith: The creation code for kendo should go into ```Bar.prototype.enter``` which has the same signature as update.

In general when writing a new widget, there are three methods of interest:
* enter
* update
* exit

Enter:  This is called once when the widget "enters" the web page - so is a good place to create and append your charts.
Exit:  This is called once when the widget "exits" the web page - so a good place to clean up any pieces before the hosting dome node is destroyed.
Update:  This is typically called when the "data",  "published property" or "size" is changed.  So typically its where you update the data within the kendo chart and update any visual settings that you may have "published" to the DSP user.

PR #1693, opened 2016-05-18
Sankey chart

GordonSmith:


GordonSmith:
@jbrundage Please Review

bower.json
```diff
@@ -42,6 +42,7 @@
     "topojson": "~1.6.19",
     "dagre": "https://raw.githubusercontent.com/cpettitt/dagre/v0.7.3/dist/dagre.js",
     "d3-cloud": "~1.2.1",
+    "d3-plugins": "*",
```
jchambers-ln: does * include all plugins here?  and if so do we really want/need to do that

src/graph/Sankey.js
```diff
@@ -0,0 +1,232 @@
...
+        this._palette = this._palette.switch(this.paletteID());
+
+        var sankeyData = this.sankeyData();
+        this._d3Sankey
+            .size([this.width(), this.height()])
+            .nodeWidth(this.vertexWidth())
+            .nodePadding(this.vertexPadding())
+            .nodes(sankeyData.vertices)
+            .links(sankeyData.edges)
+            .layout(32)
```
jchambers-ln: not sure what layout(32) is but flagging it as having a hard set value being passed in

src/graph/Sankey.js
```diff
@@ -0,0 +1,232 @@
...
+                d.y = Math.max(0, Math.min(context.height() - d.dy, d3.event.y));
+            }
+            gElement.attr("transform", "translate(" + d.x + "," + d.y + ")");
+            context._d3Sankey.relayout();
+            link.attr("d", context._d3SankeyPath);
+
+            gElement.select("text")
+                .attr("x", -6)
+                .attr("y", function (d) { return d.dy / 2; })
+                .attr("dy", ".35em")
```
jchambers-ln: hardset text size-- allow?

src/graph/Sankey.js
```diff
@@ -0,0 +1,232 @@
...
+            })
+            .style("stroke", function (d) {
+                return d3.rgb(context._palette(d.name)).darker(2);
+            })
+            .style("cursor", (this.xAxisMovement() || this.yAxisMovement()) ? null : "default")
+        ;
+        node.select("text")
+             .attr("x", -6)
+             .attr("y", function (d) { return d.dy / 2; })
+             .attr("dy", ".35em")
```
jchambers-ln: hardset text size-- allow?

jchambers-ln:
@GordonSmith a couple of questions put into the code can you double check if they concern you?

bower.json
GordonSmith: No - it means pull in latest version.  Because the plugins don't have semantic versioning, we are stuck with using the latest (which is what * is).
FYI - we do pull down all the plugins (but they are all small and self contained, but we don't pull them into the library because they are there, only the ones which are used get auto-included.

src/graph/Sankey.js
GordonSmith: This is ok - its an approximation to vertical align text (as the baseline is about a thirs of the way up from the bottom).  And 1 em is 1 Char Height.

src/graph/Sankey.js
GordonSmith: See above.

src/graph/Sankey.js
GordonSmith: Checked the docs and its the number of iterations to perform during the layout - all the demos use 32...

Issue #1743, opened 2016-06-09
HIPIE Responses may omit data fields.


GordonSmith:
When a response field is missing our tables get out of alignment.

PR #1744, opened 2016-06-09
GH-1743 HIPIE Responses may omit data fields

GordonSmith:
Insert place-holders when response data is missing.

Fixes GH-1743

### DIFF
--- a/src/common/Database.js
+++ b/src/common/Database.js
@@ -336,7 +336,11 @@
         }, this).filter(function(d) { return d.idx >= 0; });
     };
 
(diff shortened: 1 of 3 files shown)


GordonSmith:
@mlzummo Please Review

mlzummo:
@GordonSmith looks good but we need a way of specifying a value to return as well, at the marshaler level would be preferable so that it would be global for a dashboard.

GordonSmith:
@mlzummo Added - please re-review.

mlzummo:
looks good

Issue #1758, opened 2016-06-16
Graph: Filtering the same widget on the same field from two separate widgets should filter on a single value.


jprichard:
It should be OR. In the case below I have 2 widgets that filter the 3rd on the year column, the title has both filter values and only the second value from the second widget is actually applied to the filter.
http://10.241.100.159:8010/WsWorkunits/WUResult.json?Wuid=W20160616-145742&ResultName=coremultiwidgetfilter_Comp_Ins49326_DDL&SuppressXmlSchema=true

![gordon2](https://cloud.githubusercontent.com/assets/908650/16129560/287f75d0-33d3-11e6-8f2c-5d4ea55533ed.png)

PR #1762, opened 2016-06-17
GH-1758 Duplicate filter input issue

GordonSmith:
When two widgets update a third by the same field ID, the wrong filter value
can be used.

Fixes GH-1758

### DIFF
--- a/src/marshaller/HipieDDL.js
+++ b/src/marshaller/HipieDDL.js
@@ -805,20 +805,25 @@
         visitor.visit(this);
     };
 
-    Visualization.prototype.update = function (msg) {
-        var updatedBy = this.getInputVisualizations();
-        function formatParams() {
-            var paramsArr = [];
+    Visualization.prototype.update = function (params) {
+        if (!params) {
+            var validParams = {};
+            var updatedBy = this.getInputVisualizations();
             updatedBy.forEach(function (viz) {
                 for (var key in viz._eventValues) {
-                    if (viz._eventValues[key]) {
-                        paramsArr.push(viz._eventValues[key]);
-                    }
+                    validParams[key] = true;
                 }
             });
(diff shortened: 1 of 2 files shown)


GordonSmith:

src/marshaller/HipieDDL.js
```diff
@@ -806,19 +806,25 @@
     };
 
     Visualization.prototype.update = function (msg) {
-        var updatedBy = this.getInputVisualizations();
-        function formatParams() {
-            var paramsArr = [];
+        var params = msg;
```
jbrundage: Why don't you just rename the *msg* variable to *params*?

src/marshaller/HipieDDL.js
```diff
@@ -894,11 +900,17 @@
                     updatedViz.getInputVisualizations().forEach(function (inViz, idx) {
                         if (inViz._eventValues) {
                             for (var key in inViz._eventValues) {
+                                var changed = inViz === context;
```
jbrundage: I'm really not sure what's going on here. This seems like a strange place to compare these two objects?

src/marshaller/HipieDDL.js
GordonSmith: Yup will do.

src/marshaller/HipieDDL.js
GordonSmith: This "fix" is a hack for some missing information from HIPIE ("by design" - but will be fixed in a later version).  So essentially I am guessing what input params need to be defaulted to their existing values and which values need to be marked as "changed" - this is so the "updates" array and "request" params or correct (I hope)

GordonSmith:
@jbrundage Change made (and my attempt at an explanation).

jbrundage:
Looks good

Issue #1735, opened 2016-06-02
Security enhancement 


jchambers-ln:
External products require xss prevention by embedding session tokens into the header for all requests going back to the server.  Visualization needs to add a hook to allow the embedding application to set this token.

See CSRF tokens

GordonSmith:
Should expose the entire comms request/response to the custom marshaller, then they can plugin whatever comms library functions they may have.

GordonSmith:
Resolved by https://github.com/hpcc-systems/Visualization/issues/1729

PR #1768, opened 2016-06-20
GH-1735 Hookable JSONP

GordonSmith:
Allow hosting app to hook the JSONP call to ESP.

### DIFF
--- a/src/other/Comms.js
+++ b/src/other/Comms.js
@@ -195,27 +195,9 @@
         return str.join("&");
     };
 
-    Comms.prototype.jsonp = function (url, request, callback) {
-        var context = this;
+    var jsonp = function (url, request, timeout) {
         return new Promise(function (resolve, reject) {
-            for (var key in context._proxyMappings) {
-                var newUrlParts = url.split(key);
-                var newUrl = newUrlParts[0];
-                if (newUrlParts.length > 1) {
-                    var espUrl = new ESPUrl()
-                        .url(url)
-                    ;
-                    url = newUrl + context._proxyMappings[key];
-                    request.IP = espUrl._hostname;
-                    request.PORT = espUrl._port;
-                    if (newUrlParts.length > 0) {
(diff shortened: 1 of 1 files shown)


GordonSmith:
@jbrundage, @jchambers-ln Please Review.

Sample usage:
```
require(["src/other/Comms"], function(Comms) {
        Comms.hookJsonp(function (url, request, timeout) {
            function serialize(obj) {
                var str = [];
                for (var key in obj) {
                    if (obj.hasOwnProperty(key)) {
                        str.push(encodeURIComponent(key) + "=" + encodeURIComponent(obj[key]));
                    }
                }
                return str.join("&");
            };

            return new Promise(function (resolve, reject) {
                var uri = url;
                if (request) {
                    uri += "?" + serialize(request);
                }
                var xhr = new XMLHttpRequest();
                xhr.onload = function (e) {
                    if (this.status >= 200 && this.status < 300) {
                        resolve(JSON.parse(this.response));
                    }
                    else {
                        reject(Error(this.statusText));
                    }
                };
                xhr.onerror = function () {
                    reject(Error(this.statusText));
                };
                xhr.open("POST", uri);
                xhr.setRequestHeader("X-Requested-With", "XMLHttpRequest");
                xhr.send();
            });
        });
});

```

lrenn:
This LGTM, but does it help with the original issue?  DSP wants to (for example) add a CSRF token to the url.  Using this method, they would lose and thus have to reimplement your existing jsonp code.  I wonder if something like middleware would be better here...an array of methods thats that can modify the url, request, response that gets called by the existing code.  It's also possible not reading this correctly.

GordonSmith:
This isn't for DSP, its for the web apps that host a dashboard that was created in DSP.  Specifically they wanted to add some security tokens to the headers.  Also since they have access to full frameworks like JQuery, they can now use a consistent ajax call throughout their entire web app.

Issue #1774, opened 2016-06-22
Expose "download as CSV" as function


GordonSmith:
For Mega + Multi CHart

PR #1777, opened 2016-06-22
GH-1774 Add procedure downloadCSV to MegaChart

GordonSmith:
Fixes GH-1774

### DIFF
--- a/src/composite/MegaChart.js
+++ b/src/composite/MegaChart.js
@@ -88,6 +88,11 @@
         return this;
     };
 
+    MegaChart.prototype.downloadCSV = function () {
+        Utility.downloadBlob("CSV", this._chart.export("CSV"));
+        return this;
+    };
+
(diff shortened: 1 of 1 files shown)


GordonSmith:

src/composite/MegaChart.js
```diff
@@ -88,6 +88,11 @@
         return this;
     };
 
+    MegaChart.prototype.downloadCSV = function () {
+        Utility.downloadBlob("CSV", this._chart.export("CSV"));
```
jchambers-ln: does downloadBlob handle user response if for some reason the data stream isn't available?

jchambers-ln:
one note otherwise good-- i

src/composite/MegaChart.js
GordonSmith: no idea - but because this is a client side download I suspect that cannot happen?

PR #1808, opened 2016-07-12
GH-1802 Refactor Persist.discover

GordonSmith:
Move into PropertyExt

Fixes GH-1802

### DIFF
--- a/src/common/PropertyExt.js
+++ b/src/common/PropertyExt.js
@@ -154,14 +154,34 @@
     };
 
     // Publish Properties  ---
-    PropertyExt.prototype.publishedProperties = function (includePrivate) {
+    PropertyExt.prototype.publishedProperties = function (includePrivate, expandProxies) {
         var retVal = [];
         for (var key in this) {
             if (isMeta(key) && (includePrivate || !isPrivate(this, key))) {
-                retVal.push(this[key]);
+                var meta = this[key];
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jbrundage Please Review

src/common/PropertyExt.js
```diff
@@ -154,16 +154,36 @@
...
-                retVal.push(this[key]);
+                var meta = this[key];
+                if (expandProxies && meta.type) {
+                    var item = this;
+                    while (meta.type === "proxy") {
+                        item = item[meta.proxy];
+                        meta = item.publishedProperty(meta.method);
+                    }
+                    if (meta.id !== this[key].id) {
+                        meta = JSON.parse(JSON.stringify(meta));  //  Clone meta so we can safely replace the id.
```
jbrundage: I think this will cause problems with "function" and potentially "array"/"object" publish parameters.

GordonSmith: @jbrundage:
1.  This is the same code that is currently in use, just refactored from Persist.js to PropertyExt.js
2. We don't don't serialize function properties that way (and our Widget Arrays shouldn't have circular references).

There probably is a more efficient way to do this, but I would rather leave the existing behavior in place and if we do change it, do it in a diff pull request?

mlzummo:
looks good

Issue #1815, opened 2016-07-12
Format field not working for choroplethstates


mounicarachakonda:
Choropleth hover over text is not formatted. When the user changes the format value it does not get reflected.

GordonSmith:
Already fixed in v1.14.x
http://rawgit.com/hpcc-systems/Visualization/v1.14.0-rc5/demos/dermatology.html?src/map/ChoroplethStates&tooltipValueFormat=%2C.0f&projection=albersUsaPr&centerLat=38.49416886517441&centerLong=-98.32631788353747&zoom=7.974003801526609

PR #1816, opened 2016-07-12
Fixed format issue for choroplethstate issue#1815

mounicarachakonda:


jpellock:
GH-1815

src/api/ITooltip.js
```diff
@@ -109,7 +109,12 @@
         if (opts.value instanceof Date) {
             opts.value = opts.value || "";
         } else {
-            opts.value = this._valueFormatter(opts.value) || "";
+            if(typeof opts.value !== undefined){
+                opts.value = this.fields()[1].transform(opts.value);
```
jbrundage: Will this.fields()[1] always return an object?

mounicarachakonda: Yes, It always returns an object.

jpellock: You should try a test with an empty input dataset.

mounicarachakonda: I added this condition only for Tooltip. Tooltip will be called only when there is a data so without data Tooltip will not be called. I have added typeof opts.value condition incase if there is a chance for empty value.

GordonSmith:
The current code uses the "tooltipValueFormat" property to format the data, which may be overkill (having two places for the formatting) - if we are going to use the fields variant (which seems like a good ides), we should remove the tooltipValueFormat property from ITooltip?

GordonSmith:
@mounicarachakonda FYI the general template we use for commit messages looks like this:
```
GH-123 Brief issue summary

Longer description of the changes made.

Fixes GH-123

Signed off by <someuser@company.com>
```
* ```GH-123``` and ```#123``` equate to the same thing.
* The brief description should be limited to about 50 chars (as it ends up in the read me notes).
* The ```fixes GH-123``` will auto close the issue, once this pull request makes it to the master branch.
* The signature is needed to track who  made what change.

GordonSmith:
Closing as there is already a fix for this in 1.14.x

Issue #1794, opened 2016-07-07
Choropleth issue with Database.js


jchambers-ln:
Database.js:458 Uncaught TypeError: func is not a function

common/Database.js

  } else {
            return this._data.map(function (row) {
                var retVal = [];
                fieldIndicies.forEach(function (func) {
                    **retVal.push(func(row));**
                });
                return retVal;
            });
        }

jchambers-ln:
needed to fix https://github.com/hpcc-systems/DSP/issues/1534

GordonSmith:
@jchambers-ln, @drealeed - this is being caused by bad DDL/Data (and on newer versions there is a warning in the console). the mappings/field "color": "Red_Yellow_Blue" is missing in the data:
DDL:
```
...
    "visualizations": [
      {
        "id": "test",
        "source": {
          "output": "View_test",
          "mappings": {
            "color": "Red_Yellow_Blue",
            "weight": [
              "airtime_AVE"
            ],
            "state": "orgin_state"
          },
          "id": "Ins002_dsOutput1"
        },
        "type": "CHORO",
        "fields": [
          {
            "id": "orgin_state",
            "properties": {
              "type": "string"
            }
          },
          {
            "id": "airtime_AVE",
            "properties": {
              "type": "real"
            }
          },
          {
            "id": "Red_Yellow_Blue",
            "properties": {
              "type": "string"
            }
          }
        ],
        "title": "test",
        "properties": {
          "charttype": "CHORO"
        }
      }
...
```
Data:
```
...
    "Result": {
      "Ins002_dsOutput1_View_test": {
        "Row": [
          {
            "orgin_state": "AR",
            "base_count": "33668",
            "airtime_sum": 2395574.0
          },
          {
            "orgin_state": "AZ",
            "base_count": "231811",
            "airtime_sum": 25372363.0
          },
          {
            "orgin_state": "CT",
            "base_count": "29449",
            "airtime_sum": 3417182.0
          },
          {
            "orgin_state": "HI",
            "base_count": "110161",
            "airtime_sum": 11052053.0
          },
...
```

PR #1821, opened 2016-07-15
GH-1794 MissingData + Aggregates

GordonSmith:
Ignore missing data and aggregate functions.

Fixes GH-1794-MissingDataFunc

### DIFF
--- a/src/common/Database.js
+++ b/src/common/Database.js
@@ -455,7 +455,7 @@
             return this._data.map(function (row) {
                 var retVal = [];
                 fieldIndicies.forEach(function (func) {
-                    retVal.push(func(row));
+                    retVal.push(typeof func === "function" ? func(row) : row);
                 });
                 return retVal;
             });


GordonSmith:
@jchambers-ln, @hamiltwp Please Review

mlzummo:
@GordonSmith this code right here is going to have direct impact on fixing current issue i see in candidate 1.14.x however this doesnt fix the issue we are going to need more code around this?

jchambers-ln:
looks good, however I don't know your code well enough here to know if it properly fixes everything

jchambers-ln:
I made the change locally and it solved the choropleth issue, so I think it is good.

GordonSmith:
The underlying issue, is that roxie can omit fields from the response when there is no data (rather than returning empty fields.  We added code which can optionally display ***MISSING*** in the table widget, but in this scenario the field was used as part of an aggregate hence the extra test.

PR #1827, opened 2016-07-20
Fixes GH-1306 Adds PP to Pins for flexibility

jbrundage:
Fixes GH-1306

@GordonSmith Please Review

### DIFF
--- a/src/map/Pins.js
+++ b/src/map/Pins.js
@@ -19,7 +19,26 @@
     Pins.prototype.publish("geohashColumn", null, "set", "Geohash column", function () { return this.columns(); }, { optional: true });
     Pins.prototype.publish("tooltipColumn", null, "set", "Tooltip column", function () { return this.columns(); }, { optional: true });
     Pins.prototype.publish("opacity", 1.0, "number", "Opacity", null, { tags: ["Advanced"] });
-    Pins.prototype.publish("fillColor", "#006ccc", "html-color", "Pin Color", null, { optional: true });
+    Pins.prototype.publish("fillColor", "#00FFDD", "html-color", "Pin Color", null, { optional: true });
+    
+    Pins.prototype.publish("strokeWidth", 0.5, "number", "Pin Border Thickness (pixels)", null, { tags: ["Basic"] });
+    Pins.prototype.publish("strokeColor", "#000000", "html-color", "Pin Border Color", null, { optional: true });
+    
+    Pins.prototype.publish("fontSize", 18, "number", "Font Size",null,{tags:["Basic","Shared"]});
+    Pins.prototype.publish("fontFamily", "Verdana", "string", "Font Name",null,{tags:["Basic","Shared","Shared"]});
+    Pins.prototype.publish("fontColor", "#000000", "html-color", "Font Color",null,{tags:["Basic","Shared"]});
+    
+    Pins.prototype.publish("pinType", "pin", "set", "Pin Type", ["pin","circle","rectangle","rectangle-pin"], { tags: ["Basic"] });
+    Pins.prototype.publish("arrowWidth", 8, "number", "Pin arrow width (pixels)", null, { tags: ["Basic"], disable: function(w) { return ["pin","rectangle-pin"].indexOf(w.pinType()) === -1; } });
+    Pins.prototype.publish("arrowHeight", 12, "number", "Pin arrow height (pixels)", null, { tags: ["Basic"], disable: function(w) { return ["pin","rectangle-pin"].indexOf(w.pinType()) === -1; } });
+    
+    Pins.prototype.publish("pinWidth", 20, "number", "Width of pin (pixels)", null, { tags: ["Basic"], disable: function(w) { return ["rectangle","rectangle-pin"].indexOf(w.pinType()) === -1; } });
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jbrundage Requires a rebase.

src/map/Pins.js
```diff
@@ -13,8 +13,28 @@
...
+    Pins.prototype.publish("fontFamily", "FontAwesome", "string", "Font Name",null,{tags:["Basic","Shared","Shared"]});
+    Pins.prototype.publish("fontColor", "#000000", "html-color", "Font Color",null,{tags:["Basic","Shared"]});
+    
+    Pins.prototype.publish("pinType", "pin", "set", "Pin Type", ["pin","circle","rectangle","rectangle-pin"], { tags: ["Basic"] });
+    Pins.prototype.publish("arrowWidth", 18, "number", "Pin arrow width (pixels)", null, { tags: ["Basic"] });
+    Pins.prototype.publish("arrowHeight", 22, "number", "Pin arrow height (pixels)", null, { tags: ["Basic"] });
+    
+    Pins.prototype.publish("rectanglePinWidth", 50, "number", "Width of pin (pixels)", null, { tags: ["Basic"] });
+    Pins.prototype.publish("rectanglePinHeight", 50, "number", "Height of pin (pixels) (not including arrow)", null, { tags: ["Basic"] });
+    Pins.prototype.publish("rectangleCornerRadius", 10, "number", "Radius of rectangular pin corners (pixels)", null, { tags: ["Basic"] });
```
GordonSmith: These can now be disabled based on "pinType":
```
Pins.prototype.publish("rectangleCornerRadius", 10, "number", "Radius of rectangular pin corners (pixels)", null, { tags: ["Basic"], disabled: function() { return this.pinType !== "rectangle"; } });

```

src/map/Pins.js
```diff
@@ -31,28 +51,48 @@
...
+            .style("dominant-baseline",this.textBaseline())
+            .attr("dx",0)
+            .attr("dy",context.pinTextDY())
+            .text(function(d){
+                return d[2] && d[2].text ? d[2].text : "";
+            });
+        this.pinsPaths.selectAll("path.data")
+            .attr("d", function (d) {
+                return context.svgPinPath();
+            })
```
GordonSmith: The following would suffice:
```
.attr("d", this.svgPinPath())
```

And rather than calling that function for each pin - you could call it once and store it in a var and then use that for each pin.

src/map/Pins.js
jbrundage: Fantastic! I've been wishing for such a feature (assuming it hides disabled parameters in the property editor).

GordonSmith: I just fixed a few typos in the example I gave.

jbrundage: In light of the *disable* option I'll rename a few of the publish parameters.

jbrundage:
@GordonSmith That **disable** function throws an error: **this.pinType is not a function**

jbrundage:
@GordonSmith Nevermind, I see how you're using it in other widgets **function(w)** where *w* is an instance of the widget.

jbrundage:
@GordonSmith I've made your requested changes and this PR is ready for a final review :+1:

src/map/Pins.js
```diff
@@ -57,13 +76,12 @@
             .style("opacity", this.opacity())
         ;
 
-        this.pinsPaths = this._pinsTransform.selectAll(".data").data(this.visible() ? this.pinsData() : [], function (d) { return d[0]; });
+        this.pinsPaths = this._pinsTransform.selectAll(".pin").data(this.visible() ? this.pinsData() : [], function (d) { return d[0]; });
         var context = this;
-        this.pinsPaths.enter().append("path")
+        var gPinEnter = this.pinsPaths.enter().append("g").classed("pin",true);
```
GordonSmith: Minor (for reference):  ```.attr("class", "pin")``` is slightly faster than the classed method.

src/map/Pins.js
```diff
@@ -73,15 +91,34 @@
...
+            .attr("text-anchor","middle")
+        ;
+        this.pinsPaths.selectAll("text")
+            .style("stroke", this.fontColor())
+            .style("fill", this.fontColor())
+            .style("font-size", this.fontSize())
+            .style("font-family", this.fontFamily())
+            .style("dominant-baseline",this.textBaseline())
+            .attr("dx",0)
+            .attr("dy",context.pinTextDY())
```
GordonSmith: Should be ```this.pintTextDY()```

src/map/Pins.js
```diff
@@ -73,15 +91,34 @@
...
+            .text(function(d){
+                return d[2] && d[2].text ? d[2].text : "";
+            });
+        this.pinsPaths.selectAll("path.data")
+            .attr("d", context.svgPinPath())
```
GordonSmith: Should be ```this.svgPinPath()``` - I think it would be worthwhile changing this to:
```
var svgPath =  this.svgPinPath();
this.pinsPaths.selectAll("path.data")
   .attr("d", svgPath)
...
```

GordonSmith:
@jbrundage Some more minor comments, then we can pull...

jbrundage:
@GordonSmith All set

GordonSmith:
Looks good.

Issue #1838, opened 2016-07-22
Regression early setting of columns


GordonSmith:
@jchambers-ln https://github.com/hpcc-systems/Visualization/pull/1829 caused a regression when the widget in question was a input only widget (like forms).

PR #1829, opened 2016-07-21
GH-1828 DDL Columns v Persist Columns

GordonSmith:
Merge the DDL columns with Persist Columns prior to initial render.

Fixes GH-1828

Signed-off-by: Gordon Smith <gordonjsmith@gmail.com>

PR #1839, opened 2016-07-22
GH-1838 Regression caused by GH-1829

GordonSmith:
Check that widget has columns to set (forms won't).

Fixes GH-1838

### DIFF
--- a/src/marshaller/HipieDDL.js
+++ b/src/marshaller/HipieDDL.js
@@ -421,7 +421,7 @@
     };
 
     Source.prototype.getColumns = function () {
-        return this.mappings.columns;
+        return this.mappings && this.mappings.columns ? this.mappings.columns : [];
     };
 
     Source.prototype.getData = function () {


GordonSmith:
@jchambers-ln @jbrundage Please Review

src/marshaller/HipieDDL.js
```diff
@@ -782,8 +782,10 @@
     Visualization.prototype.setWidget = function (widget) {
         this.widget = widget;
         this.events.setWidget(widget);
-        var columns = this.source.getColumns();
-        this.widget.columns(columns);
+        if (this.widget.columns) {
```
mlzummo: @GordonSmith are you testing that the function exists here? should we check typeof if that is the case?

GordonSmith: I am not sure this check is even needed since all our widgets have a columns method - I think I might simply remove it?

GordonSmith:
Removed unneeded test @mlzummo can you re-review.

mlzummo:
@GordonSmith looks good

PR #1843, opened 2016-07-23
GH-1842 Generalize WsECL connection

GordonSmith:
Fixes GH-1842

### DIFF
--- a/src/other/Comms.js
+++ b/src/other/Comms.js
@@ -360,7 +360,7 @@
     WsECL.prototype.url = function (_) {
         var retVal = Comms.prototype.url.apply(this, arguments);
         if (arguments.length) {
-            //  http://192.168.1.201:8010/esp/files/stub.htm?QuerySetId=roxie&Id=stock.3&Widget=QuerySetDetailsWidget
+            //  http://localhost:8010/esp/files/stub.htm?QuerySetId=roxie&Id=stock.3&Widget=QuerySetDetailsWidget
             this._port = this._port === "8010" ? "8002" : this._port;  //  Need a better way  ---
             for (var key in this._params) {
                 switch (key) {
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jbrundage @mlzummo Please Review.

src/other/Comms.js
```diff
@@ -374,25 +374,14 @@
...
-                    queryParts = pathParts[1].split("/");
-                    if (queryParts.length >= 3) {
-                        this.target(queryParts[1]);
-                        this.query(queryParts[2]);
-                    }
-                }
-            }
-            if (!this._target && !this._query) {
+            if (!this._target || !this._query) {
                 //http://10.241.100.157:8002/WsEcl/forms/default/query/roxie/wecare
```
jchambers-ln: commented out reference to an internal IP I would revome

src/other/Comms.js
```diff
@@ -374,25 +374,14 @@
...
-                if (pathParts.length >= 2) {
-                    queryParts = pathParts[1].split("/");
-                    if (queryParts.length >= 3) {
-                        this.target(queryParts[1]);
-                        this.query(queryParts[2]);
-                    }
-                }
-            }
-            if (!this._target && !this._query) {
+            if (!this._target || !this._query) {
```
jchambers-ln: changing from an && to an !! means this.query will be evaluated even if this._target is not could this introduce a null pointer?

jchambers-ln:
see in line, otherwise good-- not sure if changes are needed but called out the change in Boolean statement.

src/other/Comms.js
GordonSmith: Its the other way round - if either are null then I want to try an alternative parse.

GordonSmith:
@jchambers-ln Changes made as suggested.

Issue #1868, opened 2016-07-29
Gmap pins mouse out broken ie11 c1.14.x


mlzummo:
The refactor of gmap/pins has created an issue where mouse out no longer works on a pin path

GordonSmith:
I have seen this before with IE.  Cause is here:  https://github.com/hpcc-systems/Visualization/blame/master/src/map/Pins.js#L89-L90

PR #1872, opened 2016-08-01
GH-1868 Disable pins "bring to top" for IE

GordonSmith:
Fixes GH-1868

### DIFF
--- a/src/map/Pins.js
+++ b/src/map/Pins.js
@@ -87,7 +87,9 @@
                 context.click(context.rowToObj(d[2].origRow), "geohash", context._selection.selected(this));
             })
             .on('mouseover', function (d) {
-                this.parentNode.appendChild(this);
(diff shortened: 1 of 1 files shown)


GordonSmith:
@jbrundage @mlzummo Please Review

jbrundage:
Looks good

jbrundage:
@GordonSmith Actually... the mouseover event callback is causing the pin icon path element to display ontop of the pin icon (text element). The onClick and onMouseover event callbacks should be bound to the **g.pin** instead of the **path.data**. I'll issue a PR for this either later tonight or tomorrow.

GordonSmith:
@jbrundage That would be a different issue to this one though (no mouse out event on IE 11)?

jbrundage:
@GordonSmith You're right. This change looks good.

Issue #1424, opened 2016-02-01
HPCC Column + stacked


GordonSmith:
Change stacked paradigm to match other charts (leave current one as "overlay?")

PR #1895, opened 2016-08-07
GH-1424 Stacked Column

GordonSmith:
Stack values instead of overlaying them.
Prevent axis extend from crossing the 0 barrier
Fix tooltips on ranged column charts

Fixes GH-1424

### DIFF
--- a/src/common/Platform.js
+++ b/src/common/Platform.js
@@ -159,5 +159,13 @@
         };
     };
 
+    Math.sign = Math.sign || function (x) {
+        x = +x; // convert to a number
+        if (x === 0 || isNaN(x)) {
+            return x;
+        }
+        return x > 0 ? 1 : -1;
+    };
+
     return Platform;
 }));
--- a/src/chart/Axis.js
+++ b/src/chart/Axis.js
@@ -219,19 +219,29 @@
                 case "ordinal":
                     break;
                 default:
-                    var length, delta, newLow, newHigh;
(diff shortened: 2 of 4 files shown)


GordonSmith:
@jbrundage @mlzummo @jchambers-ln Please Review

src/common/Platform.js
```diff
@@ -159,5 +159,13 @@
         };
     };
 
+    Math.sign = Math.sign || function (x) {
+        x = +x; // convert to a number
```
jchambers-ln: maybe consider parseInt or parseFloat

src/common/Platform.js
```diff
@@ -159,5 +159,13 @@
...
+        if (x === 0 || isNaN(x)) {
```
jchambers-ln: since you did +x before is isNaN needed?

jchambers-ln:
@GordonSmith a few minor comments

src/common/Platform.js
GordonSmith: FYI - this source code came from here:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign

+x is preferable here for performance.   You want it to fail to NaN sooner rather than in a parse.

src/common/Platform.js
GordonSmith: Yes as +"abc" is NaN

Issue #1899, opened 2016-08-08
ie10 gmap pin tooltip (custom) ... moustin is not working


mlzummo:

GordonSmith:
(When the snapshot tool works, it shows IE10 ok):  thttps://developer.microsoft.com/en-us/microsoft-edge/tools/screenshots/#http://rawgit.com/GordonSmith/Visualization/future-1.14.x/demos/dermatology.html?src/map/GMap.pins

mlzummo:
@GordonSmith sorry i wasnt precise in what i said, i have some sample code we can look at in how we are handling custom pins ..

PR #1907, opened 2016-08-09
GH-1899 Moves map pin events to g.pin element

jbrundage:
Fixes GH-1899

@GordonSmith @mlzummo Please review

### DIFF
--- a/src/map/Pins.js
+++ b/src/map/Pins.js
@@ -80,9 +80,6 @@
         var context = this;
         var gPinEnter = this.pinsPaths.enter().append("g").attr("class","pin");
         gPinEnter
-            .append("path")
-            .attr("class", "data")
-            .call(this._selection.enter.bind(this._selection))
             .on("click", function (d) {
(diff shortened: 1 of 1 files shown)


src/map/Pins.js
```diff
@@ -80,17 +80,18 @@
...
             .on('mouseover', function (d) {
                 if (!this.isIE) {
                     this.parentNode.appendChild(this);
                 }
-            })
+            });
+        gPinEnter
+            .append("path")
+            .attr("class", "data")
+            .call(this._selection.enter.bind(this._selection))
```
jchambers-ln: any concerns if this is null/not selected?

GordonSmith:
Looks fine (I assume the bug was where to "title" was being appended to.

src/map/Pins.js
jbrundage: *this._selection* is initialized as an empty instance of **SimpleSelection** so it will never be null

Issue #1871, opened 2016-07-30
Marshaller Transaction Frames


GordonSmith:
If the user selects / inputs new values during a "loading" cycle, the dashboard can get visually out of sync with the data that is displayed.

GordonSmith:
(Previous fix caused regression)

PR #1911, opened 2016-08-10
GH-1871 Request/Response Sequence

GordonSmith:
Ensure most recent request if processed last.

Fixes GH-1871

### DIFF
--- a/src/marshaller/HipieDDL.js
+++ b/src/marshaller/HipieDDL.js
@@ -1372,7 +1372,12 @@
         }
     };
 
+    var transactionID = 0;
+    var transactionQueue = [];
     DataSource.prototype.fetchData = function (request, updates) {
+        var myTransactionID = ++transactionID;
+        transactionQueue.push(myTransactionID);
+
         var context = this;
         this.filter.forEach(function (item) {
             this.request[item + _CHANGED] = request[item + _CHANGED] || false;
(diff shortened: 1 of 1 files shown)


GordonSmith:
@jbrundage @jchambers-ln Take II - please review

src/marshaller/HipieDDL.js
```diff
@@ -1394,14 +1399,17 @@
...
-                var delay = 500 - (Date.now() - now);  //  500 is to allow for all "clear" transitions to complete...
-                setTimeout(function () {
-                    context.processResponse(response, request, updates).then(function () {
-                        resolve(response);
-                    });
-                    context.dashboard.marshaller.commsEvent(context, "response", context.request, response);
-                    ++context._loadedCount;
-                }, delay > 0 ? delay : 0);
+                var intervalHandle = setInterval(function () {
+                    if (transactionQueue[0] === myTransactionID && Date.now() - now >= 500) {  //  500 is to allow for all "clear" transitions to complete...
```
jchambers-ln: this will probably work on fast computers but on slower computers it may take longer than 1/2 sec.  Is it not possible to have a status object and push/pop holds to it in the transitions and put a loop here waiting until this object says all transition holds are complete?

jchambers-ln:
I don't see any hard bug/errors I put one comment on the logic, decision is up to you.

src/marshaller/HipieDDL.js
GordonSmith: @jchambers-ln The visualization transitions are guaranteed to finish within the given time frame, what changes on fast/slow hardware is the number of "frames" that can be rendered.

Having said that, I am assuming that all the transitions are at their default duration of 250ms (which I doubled to be sure).  It is not an ideal solution, but is self contained.

PR #1915, opened 2016-08-11
Added title attribute to property editor rows

TomCharmanODG:
This allows tooltips to appear in the property editor, sourced from the property descriptions.


TomCharmanODG:
Oops, ignore this, forgot to rebase my change.

Issue #1932, opened 2016-08-16
Allow default CDN to be overridden


GordonSmith:
The current default is viz.hpccsystems.com

PR #1933, opened 2016-08-16
GH-1932 Override default CDN host

GordonSmith:
Add ability to replace viz.hpccsystems.com with an alternative source.

Fixes GH-1932

### DIFF
--- a/src/loader.js
+++ b/src/loader.js
@@ -273,6 +273,7 @@
 
     if (!root.hpccsystems.redirect) {
         root.hpccsystems.redirect = (function () {
+            var cdnHost = "viz.hpccsystems.com";
             function url(opts) {
                 opts = opts || {};
                 var protocol = opts.protocol || (root.location.protocol === "https:" ? "https:" : "http:");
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jchambers-ln Please review

src/loader.js
```diff
@@ -286,13 +287,18 @@
...
                     port: "",
                     pathname: "/" + version + "/dist-amd"
                 });
             }
 
             return {
+                cdnHost: function(_) {
+                    if (!arguments.length) return cdnHost;
+                    cdnHost = _;
+                    return this;
```
jbrundage: Doesn't *this* = *window* in that context?

hamiltwp:
It appears to be working with current DSP version.

src/loader.js
GordonSmith: No "this" === the object being returned:  https://github.com/GordonSmith/Visualization/blob/a82e8d2afe2abe32a073587fd4427c49e8533110/src/loader.js#L296-L352
Its to enable the method chaining as per the unitTest

jchambers-ln:
looks good

hamiltwp:
DDL from DSP looks right.  Filter default in DDL is only table 2

Issue #1941, opened 2016-08-17
src/graph/Graph Vertex and Edge need a classes to indicate :hover status


jbrundage:
When hovering over a Vertex or Edge we're currently applying opacity:1 to the highlighted and opacity:0.33 to the others. The highlighted should have a class assigned to them so that CSS can easily override their styles if need be.

PR #1942, opened 2016-08-17
GH-1941 Adds class to hovered Graph Vertex/Edge

jbrundage:
Fixes GH-1941

Signed-off-by: jbrundage <jbrundage372@gmail.com>

@GordonSmith Please review

### DIFF
--- a/src/graph/Graph.js
+++ b/src/graph/Graph.js
@@ -686,7 +686,9 @@
     Graph.prototype.highlightVerticies = function (vertexMap) {
         var context = this;
         var vertexElements = this.svgV.selectAll(".graphVertex");
-        vertexElements.transition().duration(this.transitionDuration())
+        vertexElements
+            .classed("graphVertex-highlighted", function (d) { return !vertexMap || vertexMap[d.id()]; })
+            .transition().duration(this.transitionDuration())
             .each("end", function (d) {
                 if (vertexMap && vertexMap[d.id()]) {
                     if (d._parentElement.node() && d._parentElement.node().parentNode) {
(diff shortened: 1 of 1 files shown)


src/graph/Graph.js
```diff
@@ -694,6 +694,9 @@
                     }
                 }
             })
+            .each(function(d){
+                d3.select(this).classed("graphVertex-highlighted",!vertexMap || vertexMap[d.id()]);
+            })
             .style("opacity", function (d) {
```
GordonSmith: Your combination of "each" and "d3.select(this)" can be avoided as its the same as:
```
    .classed("graphVertex-highlighted", function (d) { return !vertexMap || vertexMap[d.id()]; })
```
I would suggest moving it above the transition:
```
        vertexElements
            .classed("graphVertex-highlighted", function (d) { return !vertexMap || vertexMap[d.id()]; })
            .transition().duration(this.transitionDuration())
            .each("end", function (d) {
                if (vertexMap && vertexMap[d.id()]) {
                    if (d._parentElement.node() && d._parentElement.node().parentNode) {
```

src/graph/Graph.js
```diff
@@ -713,7 +716,11 @@
                     return context.highlight.edge;
                 }
                 return "1px";
-            }).transition().duration(this.transitionDuration())
+            })
+            .each(function(d){
+                d3.select(this).classed("graphEdge-highlighted",!edgeMap || edgeMap[d.id()]);
+            })
```
GordonSmith: Same as above this can be written as:
```
    .classed("graphEdge-highlighted", function(d) {return !edgeMap || edgeMap[d.id()];})
```
By convention I typically set the class before I set the styles (as mentally I think of it as the default style and any additional ones are applied on top) - but it is only my convention:
```
        edgeElements
            .classed("graphEdge-highlighted", function(d) {return !edgeMap || edgeMap[d.id()];})
            .style("stroke-width", function (o) {
                if (edgeMap && edgeMap[o.id()]) {
```

Issue #1537, opened 2016-02-29
Distinctive hover text not displayed for measures of stacked chart


dineshshetye:
URL: http://172.23.40.63:9090/DSP_DEV/
DashboardName: hover_issue

**PERSIST STRING**
```
User Id :genericUser
Composition Id :da018751-a175-4417-a1c1-2076c2b278ee
Composition Version :1.0.0
Ddl :hoverissue_Comp_Ins554_DDL
Gcid :0
Modified Date :2016-02-29
Layout :
{
  "formText": null,
  "LayoutText": {
    "__class": "marshaller_HTML",
    "__version": "1.10.0",
    "__properties": {
      "ddlUrl": "https://216.19.105.2:18010/WsWorkunits/WUResult.json?Wuid=W20160226-194440&ResultName=hoverissue_Comp_Ins554_DDL&SuppressXmlSchema=true",
      "proxyMappings": {
        "https://216.19.105.2:18010/WsWorkunits/WUResult.json": "http://172.23.40.63:9090/DSP_DEV/dev-dashboard/proxy-WUResult.do",
        "https://216.19.105.2:18010/WsWorkunits/WUInfo.json": "http://172.23.40.63:9090/DSP_DEV/dev-dashboard/proxy-WUInfo.do",
        "https://216.19.105.2:18002/WsEcl/submit/query/": "http://172.23.40.63:9090/DSP_DEV/dev-dashboard/proxy-WsEcl.do"
      },
      "designMode": false,
      "surfacePadding": "0",
      "content": [
        {
          "__class": "layout_Cell",
          "__version": "1.10.0",
          "__properties": {
            "gridRow": 0,
            "gridCol": 0,
            "gridRowSpan": 3,
            "gridColSpan": 3,
            "title": "",
            "widget": {
              "__class": "composite_MegaChart",
              "__id": "test",
              "__version": "1.10.0",
              "__properties": {
                "legendPosition": "none",
                "title": "test",
                "chartType": "AM_COLUMN",
                "chart": {
                  "__class": "amchart_Column",
                  "__version": "1.10.0",
                  "__properties": {
                    "stacked": true,
                    "xAxes": [
                      {
                        "__class": "amchart_SerialAxis",
                        "__id": "_pe787",
                        "__properties": {}
                      }
                    ],
                    "yAxes": [
                      {
                        "__class": "amchart_SerialAxis",
                        "__id": "_pe788",
                        "__properties": {}
                      }
                    ],
[...]

PR #1943, opened 2016-08-18
GH-1537 AM Chart Balloon Text Issue

GordonSmith:
Use correct series from graph object.

Fixes GH-1537

### DIFF
--- a/src/amchart/CommonSerial.js
+++ b/src/amchart/CommonSerial.js
@@ -506,13 +506,10 @@
             gObj.valueAxis = "v0";
         }
 
-        gObj.balloonFunction = function(d) {
-            //var balloonText = d.category + ", " + context.columns()[d.graph.columnIndex+1]  + ": " + context.data()[d.index][d.graph.columnIndex+1];
-            if (d.graph.type === "line") {
-                return d.category + ", " + context.columns()[d.graph.index + 1]  + ": " + context.data()[d.index][d.graph.index + 1];
-            } else {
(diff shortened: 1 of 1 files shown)


GordonSmith:
@mlzummo @jchambers-ln Please Review

src/amchart/CommonSerial.js
```diff
@@ -506,13 +506,10 @@
...
-            //var balloonText = d.category + ", " + context.columns()[d.graph.columnIndex+1]  + ": " + context.data()[d.index][d.graph.columnIndex+1];
-            if (d.graph.type === "line") {
-                return d.category + ", " + context.columns()[d.graph.index + 1]  + ": " + context.data()[d.index][d.graph.index + 1];
-            } else {
-                return d.category + ", " + context.columns()[d.graph.columnIndex + 1]  + ": " + context.data()[d.index][d.graph.columnIndex + 1];   
-            }
+        gObj.balloonFunction = function(item, graph) {
+            var key = graph.valueField;
+            var data = graph.chart.dataProvider;
+            return item.category + ", " + key + ": " + data[item.index][key];
```
jchambers-ln: is it possible for those indexes not to exist in data?

GordonSmith: I don't believe so (only if there was no bar chart displayed) - I tested on as many samples as I could and it looked good.  
There is an identical issue with click I just spotted, so will open an issue for that as well.

mlzummo:
looks like it should be ok

Issue #1954, opened 2016-08-23
Omitted data + nested dataset


GordonSmith:
Causing error during render of nested table.

PR #1955, opened 2016-08-23
GH-1954 Nested Table + no data

GordonSmith:
Check for NULL data prior to instantiating the nested Table.
Field order and mapping order can be different now.

Fixes GH-1954

### DIFF
--- a/src/marshaller/HipieDDL.js
+++ b/src/marshaller/HipieDDL.js
@@ -241,15 +241,19 @@
     TableMappings.prototype.doMapAll = function (data) {
         var retVal = SourceMappings.prototype.doMapAll.apply(this, arguments);
         if (retVal instanceof Array) {
-            //var columnsRHS = this.visualization.source.getColumnsRHS();
-            this.visualization.fields.forEach(function (field, idx) {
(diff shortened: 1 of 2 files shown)


GordonSmith:
@jchambers-ln @jbrundage Please Review

jchambers-ln:
looks good

src/marshaller/HipieDDL.js
```diff
@@ -241,15 +241,19 @@
...
+            this.visualization.fields.forEach(function (field) {
                 var fieldType = (!field || !field.properties) ? "unknown" : hipieType2DBType(field.properties.type);
-                switch (fieldType) {
-                    case "dataset":
-                        retVal = retVal.map(function (row) {
-                            row.forEach(function (cell, cellIdx) {
-                                if (cellIdx === idx) {
-                                    cell = cell.Row || cell;
+                var colIdx = columnsRHSIdx[field.id];
+                if (isNaN(colIdx)) {
```
hamiltwp: Is it certain that isNaN is always sufficient here? Coerced values like isNaN("") = false?

GordonSmith: In this case it should be either a number or undefined so should be fine (I might change it to test that explicitly)

PR #1964, opened 2016-08-26
GH-1963 locateParentWidget fails when _target is not set

GordonSmith:
Fixes GH-1963

### DIFF
--- a/src/common/Widget.js
+++ b/src/common/Widget.js
@@ -296,12 +296,14 @@
     };
 
     Widget.prototype.locateParentWidget = function (domNode) {
-        domNode = domNode || this._target.parentNode;
-        var widget = this.toWidget(domNode);
-        if (widget) {
-            return widget;
-        } else if (domNode.parentNode) {
-            return this.locateParentWidget(domNode.parentNode);
+        domNode = domNode || (this._target ? this._target.parentNode : null);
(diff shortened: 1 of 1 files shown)


GordonSmith:
@jchambers-ln Please Review

jchambers-ln:
looks good

src/common/Widget.js
```diff
@@ -296,12 +296,14 @@
...
-        if (widget) {
-            return widget;
-        } else if (domNode.parentNode) {
-            return this.locateParentWidget(domNode.parentNode);
+        domNode = domNode || (this._target ? this._target.parentNode : null);
+        if (domNode) {
+            var widget = this.toWidget(domNode);
+            if (widget) {
+                return widget;
+            } else if (domNode.parentNode) {
```
hamiltwp: Is this checking this._target.parentNode.parentNode in some cases?

GordonSmith: Yes - its a recursive algorithm which walks up the dom node parentage, until it finds a node which is associated with one of our widgets.

GordonSmith: FYI - This "fix" is when this method is called prior to the widget ever being rendered on a page (in which case _target is null and it can have no parent widget).

Issue #1984, opened 2016-09-02
VizEvents Not captured for Table Click/Double Click Events


Mohammed-Idris-K-J:
Trying to save the state of the dashboard when click/double click happens on the tables of the dashboard, where it currently saves the state on click/doubleclick of other widgets.

GordonSmith:
Background:
The current design allows users to "listen" for click evens on the marshaller.  This works fine for all widgets which have side-effects (clicking on them updates another widget in the page).
Currently widgets with no side-effects, will not emit any events and the only way to "listen" to those widgets is to register a callback with them individidually.

PR #2024, opened 2016-09-22
GH-1984 Propagate DBL Click support

GordonSmith:
Fixes GH-1984

### DIFF
--- a/src/chart/MultiChart.js
+++ b/src/chart/MultiChart.js
@@ -143,7 +143,10 @@
         if (arguments.length) {
             var context = this;
             _.click = function (row, column, selected) {
-                context.click(row, column, selected);
+                context.click.apply(context, arguments);
+            };
+            _.dblclick = function (row, column, selected) {
+                context.dblclick.apply(context, arguments);
             };
             if (this._chartMonitor) {
                 this._chartMonitor.remove();
--- a/src/common/Icon.js
+++ b/src/common/Icon.js
@@ -54,11 +54,22 @@
         ;
         this._tooltipElement = element.append("title");
         var context = this;
-        element.on("click", function(el) { context.click(el); });
+        element
+            .on("click", function (el) {
+                context.click(el);
+            })
+            .on("dblclick", function (el) {
+                context.dblclick(el);
+            })
+        ;
     };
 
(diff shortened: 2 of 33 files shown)


GordonSmith:
@jbrundage @jchambers-ln @mlzummo Please Review

GordonSmith:
@jchambers-ln I would be inclined to push this back to 1.14.8?

jchambers-ln:
Agree that looks to be a larger one-- I'll still review.

src/chart/MultiChart.js
```diff
@@ -143,7 +143,10 @@
         if (arguments.length) {
             var context = this;
             _.click = function (row, column, selected) {
-                context.click(row, column, selected);
+                context.click.apply(context, arguments);
```
jchambers-ln: Is param selected used now, is it needed?

src/common/Icon.js
```diff
@@ -54,13 +54,24 @@
...
+            })
+        ;
     };
 
     Icon.prototype.click = function (domNode) {
         console.log("Clicked the icon");
     };
 
+    Icon.prototype.dblclick = function (domNode) {
+        console.log("Double clicked the icon");
```
jchambers-ln: should that log statement be controlled so it can be disabled?

src/map/GMapPinLine.js
```diff
@@ -16,6 +16,9 @@
             .on("click", function (row, col, sel) {
                 context.click(context.rowToObj(row.ext.origRow), "", sel);
             })
+            .on("dblclick", function (row, col, sel) {
```
jchambers-ln: Not consistent naming -- some of the functions in this pull have (row, col, sel) some have (row, column, selected)

src/chart/MultiChart.js
GordonSmith: ```context.click.apply(context, arguments);``` FYI this means call the "click" event on a base class passing in the current argument list - so row, column + selected all get passed to the base click event.

src/map/GMapPinLine.js
GordonSmith: True.

Issue #2040, opened 2016-09-29
Add length to auto complete text box


jpellock:
The current size of the text box is fixed. We need a parameter to determine the desired length of the text box. 

@mounicarachakonda

PR #2055, opened 2016-10-04
GH-2040 Dynamic width for AutoComplete textbox.

mounicarachakonda:
Fixes GH-2040

Signed-off by : Mounica Rachakonda mounica.rachakonda@lexisnexis.com

### DIFF
--- a/src/other/AutoCompleteText.js
+++ b/src/other/AutoCompleteText.js
@@ -15,6 +15,7 @@
     AutoCompleteText.prototype._class += " other_AutoCompleteText";
 
     AutoCompleteText.prototype.publish("label", "Label: ", "string", "Label for AutoCompleteText");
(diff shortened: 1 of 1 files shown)


src/other/AutoCompleteText.js
```diff
@@ -15,6 +15,7 @@
     AutoCompleteText.prototype._class += " other_AutoCompleteText";
 
     AutoCompleteText.prototype.publish("label", "Label: ", "string", "Label for AutoCompleteText");
+    AutoCompleteText.prototype.publish("textboxWidth", 150, "number", "width of textbox");
```
GordonSmith: Question - do you ever want it to be just a a "default size"?  You could do it like this:
```
AutoCompleteText.prototype.publish("textboxWidth", null, "number", "width of textbox", null, {optional:true});
...
this._input.style("width",this.textboxWidth_exists() ? this.textboxWidth()+"px" : null);
```

PR #2064, opened 2016-10-06
GH-2061 Add safe stringify to Utility

GordonSmith:
Fixes GH-2061

Signed-off-by: Gordon Smith gordonjsmith@gmail.com

### DIFF
--- a/src/common/Utility.js
+++ b/src/common/Utility.js
@@ -420,6 +420,18 @@
                 }
                 return dest;
             }
+        },
+        logStringify: function (obj) {
+            var cache = [];
+            return JSON.stringify(obj, function (key, value) {
+                if (typeof value === 'object' && value !== null) {
+                    if (cache.indexOf(value) !== -1) {
+                        return;
+                    }
+                    cache.push(value);
+                }
+                return value;
+            });
         }
     };
 }));


GordonSmith:

mlzummo:
@GordonSmith not sure i understand what this is really doing? can you explain?

GordonSmith:
Its a version of JSON.stringify that won't fail on circular references.

src/common/Utility.js
```diff
@@ -420,6 +420,18 @@
                 }
                 return dest;
             }
+        },
+        stringify: function(obj) {
+            var cache = [];
```
jchambers-ln: I'm not entirely following what you are doing here,  But from my Naive point perspective a few things to consider-- if it does what you need then I'm good with this.  
1. Since this cache is local to the function is it giving you any lift after 50% of your JSON is processed?
2. It seems the code is caching values disassociated with the keys - wouldn't it be reasonable to have same values under different keys.

jbrundage: I think this code assumes there will never be two nodes with the same value (of type 'object'). This doesn't seem like a safe assumption (is it?). Caching the key+value pair would resolve this.

GordonSmith: The code is written to only serialize the first instance of a given object, not the first instance of _any_ object.
As written it is over cautious as the same object could be referenced twice at the same level and that is not a circular reference - I will think about it some more (its not actually being called at the moment and I had been planning on only using it for console.log messages...) - I might rename it to logStringify

Issue #1098, opened 2015-10-28
cell null indicatetheseids console log issue


mlzummo:
if indicatetheseids has no data the loop for the array on mousenter returns invalid data because its not looping the array correctly

PR #2066, opened 2016-10-13
GH-1098 Implementation of Pivot Table via orb.js library

yangxulexis:
Demo:  http://rawgit.com/yangxulexis/Visualization/candidate-1.14.x/demos/dermatology.html?src/react/Orb

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -30,8 +30,8 @@ const cfg = {
   prefix: "hpcc-viz"
 };
 
-const libs = ["d3", "c3", "colorbrewer", "dagre", "topojson", "d3-cloud", "font-awesome", "amcharts", "es6-promise", "amcharts.funnel", "amcharts.gauge", "amcharts.pie", "amcharts.radar", "amcharts.serial", "amcharts.xy", "amcharts.gantt", "amcharts.plugins.responsive", "simpleheat", "d3-hexbin", "d3-tip", "d3-bullet", "d3-sankey", "autoComplete", "handsontable"];
-const bundles = ["common", "layout", "api", "other", "chart", "form", "c3chart", "google", "amchart", "tree", "graph", "map", "handson", "composite", "marshaller"];  //  Order is important ---
+const libs = ["d3", "c3", "colorbrewer", "dagre", "topojson", "d3-cloud", "font-awesome", "amcharts", "es6-promise", "amcharts.funnel", "amcharts.gauge", "amcharts.pie", "amcharts.radar", "amcharts.serial", "amcharts.xy", "amcharts.gantt", "amcharts.plugins.responsive", "simpleheat", "d3-hexbin", "d3-tip", "d3-bullet", "d3-sankey", "autoComplete", "handsontable", "orb", "orb-react"];
+const bundles = ["common", "layout", "api", "other", "chart", "form", "c3chart", "google", "amchart", "tree", "graph", "map", "handson", "react", "composite", "marshaller"];  //  Order is important ---
 const lintFilter = filter(["**", "!config.js", "!map/us-counties.js", "!map/us-states.js", "!map/countries.js", "!map/us-zip5.js"]);
 
 function buildModule(module, cb) {
@@ -79,7 +79,7 @@ function css(minify) {
 }
 
 function optimize(opts, cb) {
-  //opts.optimize = "none";
+  opts.optimize = "none";
   rjs.optimize(opts,
     function (text) { cb(null, text) },
     cb
@@ -157,6 +157,9 @@ const amd_modules = bundles.map(function (bundle, idx) {
     switch (bundle) {
         case "common":
             include = ["d3", "d3-bullet", "d3-sankey", "d3-cloud", "d3-hexbin", "es6-promise"].concat(include);
+            break;
+        case "react":
+            include = ["orb-react", "orb"].concat(include);
             break;
     }
     amd_bundles["src/" + name] = include;
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -79,7 +79,6 @@ function css(minify) {
 }
 
 function optimize(opts, cb) {
-  opts.optimize = "none";
   rjs.optimize(opts,
     function (text) { cb(null, text) },
     cb
--- a/src/react/Orb.js
+++ b/src/react/Orb.js
@@ -28,12 +28,6 @@
         this.rowFields = [];
         this.dataFields = [];
         this.columnFields = [];
-        this.allColumns = this.columns().map(function (column, idx) {
-                return {
-                    name: "" + idx,
-                    caption: column
-                };
-            })
 
     }
 
--- a/src/react/Orb.js
+++ b/src/react/Orb.js
@@ -140,7 +140,7 @@
 
     	});
 
-    }
+    };
 
 
     Orb.prototype.update = function (domNode, element) { 
--- a/src/react/Orb.js
+++ b/src/react/Orb.js
@@ -6,6 +6,20 @@
         root.template_Orb = factory(root.d3, root.common_HTMLWidget, root.common_Utility, root.common_PropertyExt, root.React);
     }
 }(this, function (d3, HTMLWidget, Utility, PropertyExt){
+    
+    function Mapping(owner){
+        PropertyExt.call(this);
+        this._owner = owner;
+    }
+
+    Mapping.prototype = Object.create(PropertyExt.prototype);
+    Mapping.prototype.constructor = Mapping;
+    Mapping.prototype._class += " react_Orb";
+    Mapping.prototype.publish("addField", "", "set", "Show Toolbox or not",function() { return this._owner ? this._owner.columns() : [];}, {optional: true} );
+    Mapping.prototype.publish("location", true, "set", "Data Location",['row','column','data'], { tags: ["basic"] });
+    Mapping.prototype.publish("aggregateFunc", "", "set", "Aggregate Function type",['sum','count','min','max','avg','prod','var','varp','stdev','stdevp'], {optional: true} );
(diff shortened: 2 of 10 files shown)


src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+   }
+  }
+ }
+ 
+
+ var react = React;
+ react.unmountComponentAtNode(document.getElementById(this.id() + "_orb"));
+ this._div = element.append("div").attr("id", this.id() + "_orb");
+ this._orb = new orb.pgridwidget(this.orbConfig(ds,this.orbFields,this.rowFields,this.columnFields,this.dataFields));
+ this._orb.render(document.getElementById(this.id() + "_orb"));
```
GordonSmith: The tear down and re-creation of the Orb widget is obviously expensive.  And is only needed when in the DSP dermatology editor - IOW it should probably be inside an "if" statement checking if any for the properties changed actually need the complete re-instantiation)?

IOW (assuming a change  of theme required a tear down and re-creation):
```
if (this._prevTheme !== this.theme()) {
  this._prevTheme = this.theme();
  ...recreate orb...
} 
```

Type of thing.

test/reactFactory.js
```diff
@@ -0,0 +1,23 @@
...
+}(this, function () {
+    return {
+        Orb: {
+            simple: function (callback) {
+                require(["test/DataFactory", "src/react/Orb"], function (DataFactory, Orb) {
+                    callback(new Orb()
+                        .data(DataFactory.Pivot.subjects.data)
+                        .columns(DataFactory.Pivot.subjects.columns)
+
+
```
GordonSmith: remove blank lines.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+Orb.prototype = Object.create(HTMLWidget.prototype);
+Orb.prototype.constructor = Orb;
+Orb.prototype._class += " react_Orb";
+
+Orb.prototype.mapping = mapping;
+Orb.prototype.publish("stringProp", "defaultValue", "string", "Sample Property");
+
+Orb.prototype.publish("toolbar", true, "boolean", "Show Toolbox or not", null, { tags: ["basic"] });
+Orb.prototype.publish("themeColor", "blue", "set", "Theme color", ['blue','red','black','green'], { tags: ["basic"] });
+Orb.prototype.publish("newField" ,[], "propertyArray", "Source Columns", null, { autoExpand : mapping});
```
yangxulexis: Yes. This is the way how dashboard designer decide which fields go to row, which go to column and which go to data. For end user, there is an option called "movable", in which dashboard designer can decide if the end user can move fields around or not. So addFields() should be kept here, right?

GordonSmith: I would have thought the preferable route would be to allow the user to drag and drop the fields, while in design mode and have the result be the default?  It may not be obvious how to persist that information without these fields though.
Question:  Will we ever have a situation where the end user should _not_ be allowed to drage fields around (movable false)?  If so should we even display the fields (can they be hidden?)

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+    };
+ 
+ 
+   }
+    
+  }
+}
+
+     
+ if (this.removeField()){
```
yangxulexis: Yeah. It was supposed to give dashboard designer capability to remove any field they selected first. But actually there is some bug in it---select,remove,select again---would cause some problem. I would like to remove this function now. We could add it later when we figure it out a good way for it.

src/react/Orb.js
yangxulexis: I agree. Tear down and re-creation is kinda expensive. But the orb.js was written in react.js. If we don't want to complete re-instantiation, IOW if we just want to recreate modified subjects in config object, we have to jump into react.js code. I looked up the html structure. It is pretty complicated there.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+        this.orbFields =[];
+        this.savedField = [];
+        this.rowFields = [];
+        this.dataFields = [];
+        this.columnFields = [];
+
+    }
+
+
+function mapping(owner){
```
yangxulexis: Corrected.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+ this._owner = owner;
+}
+
+mapping.prototype = Object.create(PropertyExt.prototype);
+mapping.prototype.constructor = mapping;
+mapping.prototype._class += " react_Orb";
+mapping.prototype.publish("addField", "", "set", "Show Toolbox or not",function() { return this._owner ? this._owner.columns() : [];}, {optional: true} );
+mapping.prototype.publish("location", true, "set", "Data Location",['row','column','data'], { tags: ["basic"] });
+mapping.prototype.publish("aggregateFunc", "", "set", "Aggregate Function type",['sum','count','min','max','avg','prod','var','varp','stdev','stdevp'], {optional: true} );
+mapping.prototype.publish("formatFunction","","string","Format function");
```
yangxulexis: Corrected.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+  for (var i=0;i<this.orbFields.length;i++){
+    if (this.savedField.indexOf(this.orbFields[i].caption) === -1){
+     this.savedField.push(this.orbFields[i].caption);
+    }
+    
+  }
+
+
+ for (var k=0;k<this.newField().length;k++){
+  if (this.savedField.indexOf(this.newField()[k].__prop_addField) === -1){
```
yangxulexis: Corrected.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+ }
+
+    
+ for (var h=0; h<this.newField().length; h++){
+  var columnIndex = this.columnFields.indexOf(this.newField()[h] .__prop_addField);
+  var dataIndex = this.dataFields.indexOf(this.newField()[h] .__prop_addField);
+  var rowIndex = this.rowFields.indexOf(this.newField()[h] .__prop_addField);
+
+
+  if (this.newField()[h] .__prop_addField !== null){
```
yangxulexis: Corrected.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+   switch(this.newField()[h] .__prop_location){
```
yangxulexis: Corrected.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+  var columnIndex = this.columnFields.indexOf(this.newField()[h] .__prop_addField);
+  var dataIndex = this.dataFields.indexOf(this.newField()[h] .__prop_addField);
+  var rowIndex = this.rowFields.indexOf(this.newField()[h] .__prop_addField);
+
+
+  if (this.newField()[h] .__prop_addField !== null){
+   switch(this.newField()[h] .__prop_location){        
+
+   case "row":
+    if (this.rowFields.indexOf(this.newField()[h] .__prop_addField) === -1){
```
yangxulexis: corrected.

src/react/Orb.js
```diff
@@ -0,0 +1,305 @@
...
+     }
+     if (dataIndex > -1){
+      this.dataFields.splice(dataIndex,1);
+     }
+    }
+    
+    break;
+
+   case "column":
+    if (this.columnFields.indexOf(this.newField()[h] .__prop_addField) === -1){
```
yangxulexis: corrected.

src/react/Orb.js
GordonSmith: "update" gets called for a variety of reasons, not just on property changes (page resize for example).

yangxulexis: Do you mean set up a set of variables to store previous user input? Every time there is an update, compare them to see if they have been changed? I guess so far the pivot table is "proof of concept". Complete re-instantiation would not cause huge problem. Even the data get larger, maybe using callback could handle it appropriately?

GordonSmith:
@yangxulexis I added a demo link above, but am seeing issues when I play with it:
* Data / Columns / Rows information gets lost when page is resized
* Various rendering issues (which I thought had been previously been fixed in some of the tweaks I sent you?):
image
image

Is this the current branch?

Also I thought we were going to have all the fields available up front, rather than manually picking them in the prop editor.

yangxulexis:
@GordonSmith @salimmantri @xiahx   
IE support seems to be a tough task. I checked the difference between Chrome and IE. Pivot table on Chrome displays calculated column width rendered by orb.js libary(react.js).See the pictures. IE does not display that value, which is the reason why the table on IE does not have a consistent column width. 
![orb under dermatology]
(https://cloud.githubusercontent.com/assets/20667031/21728346/b9d0a1ea-d413-11e6-8ff1-da5b4b9798da.png)
native orb
I doubt it is due to CSS, since column width is calculated by JavaScript. Even I manually add width and give them a number in IE dev tools, it is changed back whenever the page is refreshed. 

I say it is tough because if we want to make it right to IE11 (I point to version 11 since older IE version cannot even load the whole dermatology page, not mentioning to say pivot table. And newest Microsoft web browser Edge displays pivot table without any issue, like Chrome), before beginning to modify the original code, trying to find where is wrong could be somehow complicated.

gulpfile.js
```diff
@@ -79,7 +79,7 @@ function css(minify) {
 }
 
 function optimize(opts, cb) {
-  //opts.optimize = "none";
+  opts.optimize = "none";
```
GordonSmith: @yangxulexis  bump

yangxulexis: @GordonSmith Removed it in the latest commit.

GordonSmith:
Closing in favor of:  https://github.com/hpcc-systems/Visualization/pull/2170

Issue #2094, opened 2016-11-03
Wouldn't it be neat if this library was wrapped in jQuery?


jbrundage:

dabayliss:
I think the specific goal was 'zero-stack'

jbrundage:
I was just using this issue for something to reference in a pull request. I wrote a little proof of concept I've been thinking about for a while.

GordonSmith:
Having a new demo which shows how to use the framework within JQuery would be great (I have a few showing the marshaller in bootstrap / impress / mdl already:  https://github.com/hpcc-systems/Visualization/tree/master/demos/marshaller

PR #2095, opened 2016-11-03
GH-2094 Adds jQuery plugin demo

jbrundage:
```
require(['src/chart/Pie'],function(Pie){
    new Pie()
        .target('my_element_id')
        .paletteID('Set1')
        .columns(['Name','Value'])
        .data([['A',1],['B',2]])
        .render();
})
```

becomes...

```
$('#my_element_id').chart_Pie({
    paletteID: 'Set1',
    columns: ['Name','Value'],
    data: [['A',1],['B',2]]
})
```


GordonSmith:
@jbrundage Demo looks good - can you rename "vizQuery" to "JQuery" to make it easier for folks to find?
Also it should be rebased without the street view changes in it.

PR #2100, opened 2016-11-04
Fixes GH-2083 Adds support for google maps drawing tools.

jesse-morris-ln:
Fixes #GH-2083

http://rawgit.com/jesse-morris-ln/Visualization/GH_2083/demos/dermatology.html?src/map/GMap.drawing&mapDrawingTools=true

Signed-off-by: Jesse Morris jesse.morris@lexisnexis.com

### DIFF
--- a/src/map/GMap.js
+++ b/src/map/GMap.js
@@ -9,8 +9,6 @@
     }
 }(this, function (d3, HTMLWidget, AbsoluteSurface) {
 
-    var _mapDrawings = [];
-
     //  =======================================================================
     function Overlay(map, worldSurface, viewportSurface) {
         google.maps.OverlayView.call(this);
@@ -42,8 +40,7 @@
 
         this._viewportSurface
             .target(this.div)
-            .units("pixels")
-        ;
+            .units("pixels");
 
         var panes = this.getPanes();
         panes.overlayMouseTarget.appendChild(this.div);
@@ -51,8 +48,7 @@
 
     Overlay.prototype.draw = function () {
         var projection = this.getProjection();
-        if (!projection)
-            return;
+        if (!projection) return;
 
         var bounds = this._map.getBounds();
         var center = projection.fromLatLngToDivPixel(bounds.getCenter());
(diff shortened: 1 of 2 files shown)


jesse-morris-ln:
@GordonSmith @jbrundage please review.

src/map/GMap.js
```diff
@@ -238,6 +239,25 @@
         this._prevCenterLat = this.centerLat();
         this._prevCenterLong = this.centerLong();
         this._prevZoom = this.zoom();
+        
+        if (this.mapDrawingTools()) {
```
GordonSmith: This should be moved to the update method so that it can be enabled and disabled at runtime (anything in the "enter" method is only called once)?

jesse-morris-ln: Moving the initialization of the drawing tools to the update function makes sense, I'll see to that.  I'll also add a way to keep track of and get/set the drawn objects.

jesse-morris-ln: In the most recent commit, I've kept the drawing tools initialization in the enter function, but I've moved setting the map object on the drawing tools object into the update method, and this allows for toggling the drawing functionality on/off.  I did this so that the update function effectively enables/disables the drawing tools, but doesn't have to concern itself with initializing or tearing down the drawing tools object.

Additionally, added a published property which is just a running array of the objects drawn on the map.

GordonSmith:
Looking better - in the demo page, click on the "clone" option in the menu - the cloned map should be redrawing the users shapes - once that works we should be good to pull.

GordonSmith:
@jesse-morris-ln The cloning appears to be working now, Can you rebase onto latest upstream/candidate-1.16.x?

GordonSmith:
@jesse-morris-ln - your original branch seems to have disappeared - is there a new one?

jesse-morris-ln:
@GordonSmith Yes, I was having trouble rebasing to the current 1.16.x branch, I re-forked and re-cloned the repository.  I'm going to create a new pull request for the new branch.

Issue #2121, opened 2016-12-01
Google Map Duplication


GordonSmith:
Allow hosting app to set their own google script include - and prevent duplication,

PR #2122, opened 2016-12-01
GH-2121 Delay load GMap

GordonSmith:
Allow host page to specify their own load script.

Fixes GH-2121


GordonSmith:
@jbrundage @jchambers-ln Please Review (this url will look better:  https://github.com/hpcc-systems/Visualization/pull/2122/files?w=1)

The primary reason to push the loading into the render method is for none "require" based builds.  The added bonus is when the user is hosting on a pgae which already has loaded google maps.

PR #2142, opened 2016-12-16
Fixes GH-2083 Adds support for google maps drawing tools.

jesse-morris-ln:
Fixes #GH-2083

http://rawgit.com/jesse-morris-ln/Visualization/GH_2083/demos/dermatology.html?src/map/GMap.drawing&mapDrawingTools=true

Signed-off-by: Jesse Morris jesse.morris@lexisnexis.com

### DIFF
--- a/src/map/GMap.js
+++ b/src/map/GMap.js
@@ -119,7 +119,7 @@
 
         return new Overlay(map, worldSurface, viewportSurface);
     }
-    
+
     function UserShapeSelectionBag(mapObj) {
         this._userShapes = [];
         this.mapContext = mapObj;
@@ -146,14 +146,14 @@
             return UserShapeSelectionBag.prototype._saveShape(shape);
         });
     };
-    
+
     UserShapeSelectionBag.prototype.load = function(_) {
-        this._deserializeShapes(_);       
+        this._deserializeShapes(_);
     };
-    
+
     UserShapeSelectionBag.prototype._saveShape = function(_) {
         var retVal = {};
-        
+
         var createShapes = {
             "circle": function(_) {
                 retVal.type = "circle";
@@ -172,9 +172,9 @@
             },
             "polygon": function(_) {
                 retVal.type = _.__hpcc_type;
-                
+
                 var vertices = _.getPath();
-                    
+
                 retVal.vertices = [];
 
                 for (var i = 0; i < vertices.length; i++) {
@@ -185,21 +185,21 @@
                  createShapes.polygon(_);
             }
         };
-        
+
         createShapes[_.__hpcc_type](_);
         retVal.strokeWeight = _.strokeWeight;
         retVal.fillColor = _.fillColor;
         retVal.fillOpacity = _.fillOpacity;
         retVal.editable = _.editable;
         retVal.clickable = _.clickable || true;
-        
+
         return retVal;
     };
-    
+
     UserShapeSelectionBag.prototype._deserializeShapes = function(_) {
-        
+
         var shapes = JSON.parse(_);
-        
+
         var defOptions = {
             strokeWeight: 0,
             fillOpacity: 0.45,
@@ -207,7 +207,7 @@
             editable: true,
             clickable: true
         };
-        
+
         var createShapes = {
             "circle": function(_, map) {
                 var shape = new google.maps.Circle({
@@ -246,7 +246,7 @@
                     fillOpacity: _.fillOpacity || defOptions.fillOpacity,
                     editable: _.editable || defOptions.editable,
                     clickable: _.clickable || defOptions.clickable,
-                    map: map,                        
+                    map: map,
                     paths: _.vertices
                 });
                 return shape;
@@ -264,7 +264,7 @@
                 return shape;
             }
         };
-        
+
         for (var i = 0; i < shapes.length; i++) {
             var shape = createShapes[shapes[i].type](shapes[i], this.mapContext._googleMap);
             this.mapContext.onDrawingComplete({ type: shapes[i].type, overlay: shape });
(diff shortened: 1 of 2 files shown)


jesse-morris-ln:
@GordonSmith @jbrundage please have a look.

jbrundage:
@GordonSmith Looks good to me. This is needed for AVCC

src/map/GMap.js
```diff
@@ -305,6 +484,21 @@
...
+            
+            // Add drawing complete listener to maintain array of drawingState.
+            google.maps.event.addListener(
+                this._drawingManager,
+                "overlaycomplete",
+                function(){
+                    GMap.prototype.onDrawingComplete.apply(context, arguments);
+                });
+        } else {
+            this._drawingManager.setMap(null);
```
GordonSmith: Should remove the listener at this point?

src/map/GMap.js
```diff
@@ -455,6 +649,66 @@
     GMap.prototype.zoomToFit = function () {
         return this.zoomTo(this.data());
     };
+    
+    GMap.prototype.drawingOptions = function(_) {
+        if(typeof(_) === "object") {
+            if (typeof(this._drawingManager) === "object") {
```
GordonSmith: This test is not needed.

src/map/GMap.js
```diff
@@ -455,6 +649,66 @@
     GMap.prototype.zoomToFit = function () {
         return this.zoomTo(this.data());
     };
+    
+    GMap.prototype.drawingOptions = function(_) {
+        if(typeof(_) === "object") {
```
GordonSmith: I don't think I would bother with this test (since we don't test params like this anywhere else).

src/map/GMap.js
```diff
@@ -455,6 +649,66 @@
...
+            if (typeof(this._drawingManager) === "object") {
+                this._drawingManager.setOptions(_);
+            }
+        }
+        return this._drawingManager;
+    };
```
GordonSmith: Typically the pattern we use is:
```
GMap.prototype.drawingOptions = function(_) {
    if (!arguments.length) return this._drawingManager.getOptions();
    this._drawingManager.setOptions(_);
    return this;
}
```
I am not sure "getOptions" exists though?

src/map/GMap.js
```diff
@@ -455,6 +649,66 @@
...
+    GMap.prototype.userShapeSelection = function (_) {
+        if (!arguments.length) return this._userShapeSelection;
+        if (this._userShapeSelection) {
+            this._userShapeSelection.setEditable(false);
+        }
+        this._userShapeSelection = _;
+        if (this._userShapeSelection) {
+            this._userShapeSelection.setEditable(true);
+        }
+    };
```
GordonSmith: should ```return this;``` for method chaining.

src/map/GMap.js
```diff
@@ -455,6 +649,66 @@
...
+                }
+            });
+            window.addEventListener('keyup', function(e) {
+                if (e.ctrlKey === false) {
+                    ctrl = false;
+                }
+            });
+            google.maps.event.addListener(newShape, 'click', function (ev) {
+                context.userShapeSelection(newShape);
+                if (ev && ctrl === true) {
```
GordonSmith: I thought "ev" would contain the ctrl key state at tthis point and there is no need for the key down/hey up handlers?

src/map/GMap.js
```diff
@@ -455,6 +649,66 @@
...
+                if (e.keyIdentifier === "Control" || e.ctrlKey === true) {
+                    ctrl = true;
+                }
+            });
+            window.addEventListener('keyup', function(e) {
+                if (e.ctrlKey === false) {
+                    ctrl = false;
+                }
+            });
+            google.maps.event.addListener(newShape, 'click', function (ev) {
```
GordonSmith: Is it ok to add this event handler repeatedly?

src/map/GMap.js
jesse-morris-ln: There is no getOptions method in on the drawing manager object in the google maps API.  Returning the drawing manager object itself though provides an object that encapsulates all of the drawing options currently specified.

I can re-factor this to do that if it's acceptable.

src/map/GMap.js
jesse-morris-ln: To my understanding there's no guarantee that the "ev" parameter will contain the ctrl key state, that's the reasoning behind having the keydown and keyup handlers.

src/map/GMap.js
jesse-morris-ln: OK, will do.

src/map/GMap.js
jesse-morris-ln: I can use the [clearInstanceListeners](https://developers.google.com/maps/documentation/javascript/reference#event) method to just clear any event listeners assigned to the drawing manager object wholesale if that's acceptable here.

src/map/GMap.js
jesse-morris-ln: For clarity's sake, the reason that the "ev" parameter having the ctrl key state is a problem is because it is a google maps event object, not a DOM event object.  Essentially, "ev" contains an instance of "MouseEvent" but that property is not documented, and its name may not be the same on each execution.  Furthermore there is no method exposed by the google maps event object for accessing the underlying DOM event.  The workaround that I've seen most frequently in searching for a solution is to register separate keyup and keydown handlers to manage ctrl key state.

jesse-morris-ln:
@GordonSmith I made some changes in line with what you requested.  Can you give it a look?  This is needed for AVCC.

src/map/GMap.js
```diff
@@ -456,5 +651,65 @@
         return this.zoomTo(this.data());
     };
 
+    GMap.prototype.drawingOptions = function(_) {
+        if (!arguments.length) {
+            return this._drawingManager;
```
GordonSmith: We decided ter

Issue #2156, opened 2017-01-27
Border Radius for Widgets


albertodesignz:
Ability to specify radius to specific corners of a widget:

surfaceBorderRadiusAll - All corners are affected

surfaceBorderRadiusTop -Top 2 corners are affected

surfaceBorderRadiusTopRight - Top right corner is affected

surfaceBorderRadiusTopLeft - Top left Corner is affected

surfaceBorderRadiusBottom - Bottom 2 corners are affected

surfaceBorderRadiusBottomLeft - Bottom left corner is affected

surfaceBorderRadiusBottomRight - Bottom right corner is affected

GordonSmith:
Surface as we know it is going away in 2.0

albertodesignz:
Great! do you have a link or place where i can see the update?

PR #2167, opened 2017-02-21
GH-2156 Border Radius

mounicarachakonda:
Fixed GH-2156

Signed-off by : Mounica Rachakonda mounica.rachakonda@lexisnexis.com

### DIFF
--- a/src/other/PropertyEditor.js
+++ b/src/other/PropertyEditor.js
@@ -40,6 +40,21 @@
     PropertyEditor.prototype.publish("excludeParams", [], "array", "Exclude this array of params (widget.param)",null, {});
 
     PropertyEditor.prototype.publish("widget", null, "widget", "Widget",null,{tags:["Basic"], render:false});
+    /*
+    * Start:- Border radius functionality for surface area
+     */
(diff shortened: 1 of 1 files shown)


GordonSmith:
@mounicarachakonda This only changes the cell style while its in the property editor - but won't change it when loaded in a customers application?

GordonSmith:
@mounicarachakonda bump

mounicarachakonda:
Sure, Will look into it.

src/other/PropertyEditor.js
```diff
@@ -78,6 +93,49 @@
...
+            radiusValExist = true;
+        }
+        if(this.borderRadiusBottomLeft_exists() && this.borderRadiusBottomLeft()){
+            radiusStyleObj = {};
+            radiusStyleObj['border-bottom-left-radius'] = this.borderRadiusBottomLeft()+'em';
+            radiusValExist = true;
+        }
+        return radiusStyleObj;
+    };
+
```
hamiltwp: Your JS probably handles it but it looks here like you might set the shorthand border-radius and then possibly expect to use individual inherited properties, which my understanding would be ignored by the browser? If it could be an issue perhaps some logic here to avoid that situation.

Issue #2164, opened 2017-02-14
Dashboard with Summary widgets loading EMPTY data in verision 1.14.8-rc3


Mohammed-Idris-K-J:
Came across this issue in version 1.14.8-rc3 where the data for the dashboard with summary widgets were loaded empty. But the same dashboard works in 1.14.6.

After doing code compare found the below piece of code in HipieDDL.js to fix the issue.
![dashboards loading empty data](https://cloud.githubusercontent.com/assets/20775282/22947509/48c0be0c-f2c9-11e6-8cfd-4b08f100568c.png)
 Please let us know whether its the appropriate fix.

GordonSmith:
Is your summary widget dependent on a selection from an upstream widget?

Mohammed-Idris-K-J:
Its not dependent on any selection from the other widget. When it loads, it loads with data. In 1.14.6 when it loads it loads data for all the summary widgets without any selection.

jchambers-ln:
Idris, if you have applied a filter then it does depend on upstream widgets (it just isn't obvious that is occurring some times)

PR #2168, opened 2017-02-22
GH-2164 PrimeData Logic Priorities

GordonSmith:
During initial load change logic to give preference over inputs with default values.
Then fallback to default call for roxie queries and widgets with no inputs.

Fixes GH-2164


GordonSmith:
@jbrundage @jchambers-ln @drealeed Please Review

jchambers-ln:
looks good

Issue #2254, opened 2017-07-28
DDL Sort Fails on calculated field


GordonSmith:

GordonSmith:
Resolved now.

PR #2256, opened 2017-07-28
GH-2254 Sorting by calculated fields fails.

GordonSmith:
Move sorting to after mappings.
Rollback previous stable nesting change.
Warn when WU is archived.


GordonSmith:
@drealeed, @jchambers-ln please review.
This essentially rolls back to the rc1 fix, but with the correct sort mappings.

PR #2259, opened 2017-07-28
Amcharts serial: Formatter getting clobbered with multiple Y Axis

clewisln:
There are two problems with how the custom formatting is done with the Y Axis in CommonSerial:

1. All formatters are being written to **this.valueFormatter**; 'this' being the chart object. Automatically, the X axis formatter will be clobbered by the Y axis formatter on a serial chart. If there are multiple Y axes, then the last one "wins".

2. While the labelFunction is an anonymous function accessing the yAxis via closure, the reference it is grabbing is always the **last Y axis**. For serial charts with more than one Y axis, earlier axes will be looking at the wrong data for formatting purposes. This was resolved by creating a wrapper that passes in the axis and formatter as parameters, to force a new closure.

### DIFF
--- a/src/amchart/CommonSerial.js
+++ b/src/amchart/CommonSerial.js
@@ -345,7 +345,8 @@
         };
 
         for (var i = 0; i < this.yAxes().length; i++) {
-            var yAxis = this.yAxes()[i];
+            var valueFormatter,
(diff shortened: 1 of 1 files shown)


clewisln:
Closing, as a new request against 1.16.x was created instead

Issue #2264, opened 2017-07-28
Viz deselect and nullable.


GordonSmith:
Selecting and then deselecting an item in a visualization will cause the filter condition to be ignored,

GordonSmith:
Resolved.

PR #2265, opened 2017-07-28
GH-2264 Select/deselect issue

GordonSmith:
On deselection we expect the changed field to be "undefined"

Fixes GH-2264

### DIFF
--- a/src/other/Comms.js
+++ b/src/other/Comms.js
@@ -927,7 +927,7 @@
         function filterResults(request) {
             var changedFilter = {};
             for (var key in request) {
-                if (request[key] !== undefined && request[key + "_changed"] !== undefined) {
+                if (request[key + "_changed"] !== undefined) {
                     changedFilter[key] = {
                         value: request[key]
                     };


GordonSmith:
@drealeed @jchambers-ln please review

src/other/Comms.js
```diff
@@ -927,7 +927,7 @@
         function filterResults(request) {
             var changedFilter = {};
             for (var key in request) {
-                if (request[key] !== undefined && request[key + "_changed"] !== undefined) {
+                if (request[key + "_changed"] !== undefined) {
                     changedFilter[key] = {
                         value: request[key]
```
jchambers-ln: you are looping through the request on the keys so I assume request[key] could never be a NPE... but I'm not sure what you are gaining by removing request[key] !== undefined since that should never be the case should it?  Would it also not be the case that the for (var key in request) would also contain the key that you are building from request[key +"_changed"]... maybe I'm missing something.

GordonSmith: This is the "deselect" issue, on first click you get a request like this:
{
paramA:  "007",
paramA_changed: true
}

When deselecting you would get:
{
paramA: undefined,
paramA_changed:true
}

Which was getting filtered out by the above code.

Issue #2309, opened 2017-11-03
Liquid Fill Gauge widget


mounicarachakonda:
Should return an object with an update method which allows the gauge value to be changed.

Features should include:

Changeable min/max values.
All colors.
Outer circle thickness.
Gap between the outer circle and inner fill area.
Wave height.
Wave speed.
Wave count.
Wave rise time.
Wave height scaling on/off. Reduces the wave height near the min/max values so that the wave won't make the fill area appear total full or totally empty.
Wave starting offset. Most useful when wave animation is turned off and you want the wave min or max at a specific horizontal position in the fill area.
Wave rising upon load on/off.
Wave animation on/off.
Text height.
Text vertical position.
Text increment from min value upon loading.
Display of % symbol on/off.

mounicarachakonda:
GH-2310

GordonSmith:
Expired - please reopen if needed.

PR #2310, opened 2017-11-03
GH-2309 Liquid Fill Gauge

mounicarachakonda:
http://rawgit.com/mounicarachakonda/Visualization/gauge/demos/dermatology.html?src/chart/Guage.simple&numerator=D&denominator=A&value=5.8

### DIFF
--- a/src/chart/LiquidFillGauge.css
+++ b/src/chart/LiquidFillGauge.css
@@ -1,0 +1,4 @@
+.chart_LiquidFillGauage path {
+    cursor: pointer;
+}
+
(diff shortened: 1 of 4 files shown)


GordonSmith:
There appears to be no indentation on the main file which makes it very hard to read.

Also how interchangeable is this gauge with the existing Summary widget?

mounicarachakonda:
@GordonSmith Corrected the indentations on the main file.
This gauge is filled with the water by calculating the numerator and denominator values that user selects from the drop down.

Issue #2316, opened 2017-11-20
Pins tooltip functionality should use the standard "tooltipHTML" implementation


jbrundage:

GordonSmith:
Merged

PR #2318, opened 2017-11-20
GH-2316 Adds d3-tip to map/Pins

jbrundage:
Fixes GH-2316

@GordonSmith please review

### DIFF
--- a/src/map/Pins.js
+++ b/src/map/Pins.js
@@ -1,11 +1,11 @@
 "use strict";
 (function (root, factory) {
     if (typeof define === "function" && define.amd) {
-        define(["d3", "./Layer", "./Utility", "../common/Palette", "../common/Utility", "css!./Pins"], factory);
+        define(["d3", "./Layer", "./Utility", "../common/Palette", "../common/Utility", "../api/ITooltip", "css!./Pins"], factory);
     } else {
-        root.map_Pins = factory(root.d3, root.map_Layer, root.map_Utility, root.common_Palette, root.common_Utility);
+        root.map_Pins = factory(root.d3, root.map_Layer, root.map_Utility, root.common_Palette, root.common_Utility, root.api_ITooltip);
     }
-}(this, function (d3, Layer, MapUtility, Palette, Utility) {
+}(this, function (d3, Layer, MapUtility, Palette, Utility, ITooltip) {
     function Pins() {
         Layer.call(this);
         Utility.SimpleSelectionMixin.call(this);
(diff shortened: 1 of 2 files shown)


src/map/Pins.js
```diff
@@ -74,17 +75,22 @@
...
+            })
+        ;
     };
 
     Pins.prototype.layerUpdate = function (base) {
         Layer.prototype.layerUpdate.apply(this, arguments);
 
         this._pinsTransform
-            .style("opacity", this.opacity())
+        .style("opacity", this.opacity())
```
jbrundage: @GordonSmith These indentation changes are somehow happening via VSCode. Have you experienced this issue before?

src/map/Pins.js
```diff
@@ -104,6 +110,16 @@
                 }
             })
         ;
+        gPinEnter
+            .each(function(d){
+                if(d.ext && d.ext.tooltip){
```
GordonSmith: While unlikely a data row may start off with no tooltip data, but may get some later on - so you should add the event to every pin regardless.  I may mean we need some more logic elsewhere.

src/map/Pins.js
```diff
@@ -74,17 +75,22 @@
...
     Pins.prototype.layerEnter = function (base, svgElement, domElement) {
         Layer.prototype.layerEnter.apply(this, arguments);
-
+        
         this._pinsTransform = svgElement;
         this._selection.widgetElement(this._pinsTransform);
         this.pinsPaths = d3.select(null);
+        this
+            .tooltipHTML(function (d) {
+                return d.ext.tooltip === undefined ? "" : d.ext.tooltip;
```
GordonSmith: Could show the label if there was no specific tooltip?

src/map/Pins.js
GordonSmith: FWIW I turn off "format on save" for non VS Code projects (File/settings -> workspace).  Once you get to 2.0, they have all been reformatted for vscode.

Also you can add the ?w=1 option to the address bar to ignore the whitespace changes.

src/map/Pins.js
jbrundage: I don't think the GMapPin data has a label?

src/map/Pins.js
jbrundage: Which address bar? The git repo I'm pointing to while pushing?

src/map/Pins.js
jbrundage: I believe the pin data points are considered "new" if any aspect of them changes (line ~96 in Pins.js), so "enter" status should apply to the pins after a tooltip is added.

jbrundage: Would this call to **.data()** key the data rows on index? or on the row's content?
```javascript
this.pinsPaths = this._pinsTransform.selectAll(".pin").data(this.visible() ? this.pinsData() : []);
```

src/map/Pins.js
GordonSmith: ...true...

src/map/Pins.js
GordonSmith: The one pointing to this page ^^^^

src/map/Pins.js
GordonSmith: When no "id" function is passed into the ```data``` function on line ~96 it uses the idx position of the data row as its identifier.

jbrundage:
@GordonSmith ready to be rereviewed

Issue #2322, opened 2017-11-22
Retrofit 1.14.x serialization versions to 1.18.x


jbrundage:

GordonSmith:
Merged

PR #2326, opened 2017-11-22
GH-2322 Adds 114 to 118 retrofit logic to loader

jbrundage:
Fixes GH-2322

@GordonSmith please review


src/loader.js
```diff
@@ -348,6 +348,9 @@
                             });
                             break;
                         default:
+                            if(parseInt(state.__version.split('.')[1]) === 14){
```
GordonSmith: Needs to check the major version as well:
In the future we will have a 2.14.x

src/loader.js
```diff
@@ -360,6 +363,27 @@
                     function requireErrorHandler(err) {
                         console.log("Loader 'create' Error:\n" + err.message);
                     }
+                    function retrofit_114_to_118(state){
+                        var replacement_version = '1.18.0';
```
GordonSmith: Should spit out a warning to the console that it has encountered an old persist and is upgrading it.

src/loader.js
jbrundage: I considered this but thought that this code likely wouldn't exist in 2.0+

jbrundage:
@GordonSmith requested changes added

GordonSmith:
@jbrundage you missed my comment at the top about a test or two.  You either grab some of our demo ones (we don't need the data to test this) or create some here:  http://rawgit.com/hpcc-systems/Visualization/v1.14.8/demos/dermatology.html (there is a "save" option in the menu.

src/loader.js
```diff
@@ -360,6 +364,28 @@
...
```
GordonSmith: You can get the current version from common/Platform

src/loader.js
```diff
@@ -360,6 +364,28 @@
...
+                        console.log("Upgrading old persist from "+state.__version+" to "+replacement_version);
+                        var _json_str = JSON.stringify(state);
+                        _json_str = _json_str.split('"'+state.__version).join('"'+replacement_version);
+                        var ret_obj = JSON.parse(_json_str);
+                        ret_obj.__properties.content.forEach(function(n){//FOR EACH top tier layout_Cell
+                            if(JSON.stringify(n).split('graph_Graph').length > 1 && parseInt(replacement_version.split('.')[1]) >= 16){
```
GordonSmith: Add a new function to common/Utility to parse a semVersion.

jbrundage:
@GordonSmith 

I've made your requested changes.

http://rawgit.com/jbrundage/Visualization/GH2322_b/demos/dermatology.html?src/marshaller/HTML.with_114_GRAPH

src/common/Utility.js
```diff
@@ -499,6 +499,46 @@
...
+        template: template,
+        parseVersionString: function(versionString){
+            var _sp = versionString.split('.');
+            return {
+                major: parseInt(_sp[0].replace('v','')),
+                minor: parseInt(_sp[1]),
+                patch: parseInt(_sp[2].split('-')[0]),
+            };
+        },
+        retrofit_114_serialization: function(state,replacement_version) {
```
GordonSmith: @jbrundage parseVersionString belongs in Utility - but the "retrofit" does not.

src/common/Utility.js
```diff
@@ -499,6 +499,46 @@
...
+            replacement_version = !replacement_version || replacement_version === '1.14.2-dev' ? '1.18.0' : replacement_version;
```
GordonSmith: replacementVersion should be the current version found in "common/Platform"

src/common/Utility.js
```diff
@@ -499,6 +499,46 @@
...
+            if(!state.__version)return state;
+            var state_version_obj = this.parseVersionString(state.__version);
+            var target_version_obj = this.parseVersionString(replacement_version);
+            if(state_version_obj.major === 1 && state_version_obj.minor === 14){
+                console.log("Upgrading old persist from "+state.__version+" to "+replacement_version);
+                var _json_str = JSON.stringify(state);
+                _json_str = _json_str.split('"'+state.__version).join('"'+replacement_version);
+    
+                var ret_obj = JSON.parse(_json_str);
```
GordonSmith: Not sure I would have done it that way, but rather walked (visited) the structure and changed the version per item (and then testing for "graph_Graph" at the same time.

You could also abort early at the first instance of a version > 1.14

src/loader.js
```diff
@@ -348,12 +348,16 @@
...
                         default:
-                            this.cdn("v" + state.__version, function (_req) {
-                                _req(["src/other/Persist"], function (Persist) {
-                                    Persist.create(state, function (widget) {
-                                        callback(widget, _req);
-                                    });
-                                }, requireErrorHandler);
+                            var context = this;
+                            require(['src/common/Utility','src/common/Platform'],function(Utility,Platform){
+                                state = Utility.retrofit_114_serialization(state,new Platform().version());
```
GordonSmith: The retrofit code, belongs in the loader.js file.

GordonSmith: This doesn't look correct what if the persist string is v1.14 and the referenced ddl is the correct 1.14 version?

src/other/Persist.js
```diff
@@ -280,6 +280,7 @@
             if (typeof state === "string") {
                 state = JSON.parse(state);
             }
+            state = Utility.retrofit_114_serialization(state,new Platform().version());
```
GordonSmith: The retrofit code belongs in this source file - and if so it doesn't need to use any Platform().version test at all (as we know it has to be > 1.14.

IOW If the viz framework in use is >= v1.20.0 and we are passed a v1.14 persist string, then we do the upgrade on the fly.

The responsibility of initializing the correct viz framework version resides externally.

src/common/Utility.js
jbrundage: I considered this but the string split approach seemed less error prone

src/other/Persist.js
jbrundage: The `Platform().version` test gives us the target version for the upgrade script. The only alternative that I can imagine is a hard-coded default target? (probably 1.18.0)

src/loader.js
jbrundage: Okay sounds good, I'll move it

Issue #2328, opened 2017-11-27
PropertyEditor optional "range" input for "number" publish parameter type


jbrundage:

GordonSmith:
Expired - please reopen if needed.

PR #2329, opened 2017-11-27
GH-2328 Adds optional 'range' input for 'number' params

jbrundage:
@GordonSmith please review

The 'range' input makes it quite a bit easier to test certain things.

### DIFF
--- a/src/other/PropertyEditor.js
+++ b/src/other/PropertyEditor.js
@@ -494,23 +494,46 @@
                 ;
                 break;
             default:
-
-                cell.append("input")
-                    .attr("id", this.id() + "_" + param.id)
-                    .classed("property-input", true)
-                    .on("change", function () {
-                        context.setProperty(widget, param.id, this.value);
-                    })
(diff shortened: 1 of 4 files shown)


src/other/PropertyEditor.js
```diff
@@ -494,23 +494,46 @@
...
+                    if (param.type === "html-color" && !this.isIE) {
+                        cell.append("input")
+                            .attr("id", this.id() + "_" + param.id + "_2")
+                            .classed("property-input", true)
+                            .attr("type", "color")
+                            .on("change", function () {
+                                context.setProperty(widget, param.id, this.value);
+                            })
+                        ;
+                    }
```
GordonSmith: @jbrundage  I thought we had color already - but I guess not?

jbrundage: Github's diff is a bit misleading. That color conditional was on line 505 in the old file.

I believe this seems new because of the indentation?

PR #2354, opened 2018-01-23
Adds publish param interfaces

jbrundage:
@GordonSmith What are we using for the type assignments for "array" and "widgetArray" publish param types?


jbrundage:
Closing in favor of taking a different approach to publish parameter interfaces

Issue #2342, opened 2018-01-18
Legend Hover + Chart Highlight


GordonSmith:
Dependent on:  https://github.com/hpcc-systems/Visualization/issues/2341

When the mouse hovers over an item in the legend AND the corresponding chart implements the IHighlightColumn interface (ask me about typescript "type guards"), then the legend should call the highlight method with the series ID.  (and call it with null when the mouse is no longer hovering over an item)

PR #2358, opened 2018-01-26
Adds series classes on chart package widgets

### LINKED ISSUE #2341: Chart Highlight
1.  Add a new interface (in API?):
```
interface IHighlightColumn {
  highlight(columnID: string);
}
```
2.  For each chart that is derived from XYAxis make sure its "data point" has a css class style "series" corresponding to the column idx:  class="series-1" for example.  Where a "data point" for the column chart is the rectangle and for the line chart its the line etc.)
3.  Implement that interface on "XYAxis" and when its set is can do a `selectAll(".highlight").classed(".highlight", false).select(`.series-${idx}`).classed(".highlight", true)` type of thing (no need to call render this way).

jbrundage:
Fixes #2341 aside from the "Interface" bit... could that be handled in #2342 ?

@GordonSmith please review

### DIFF
--- a/packages/chart/src/Bubble.ts
+++ b/packages/chart/src/Bubble.ts
@@ -70,7 +70,9 @@ export class Bubble extends SVGWidget {
 
         //  Enter  ---
         node.enter().append("g")
-            .attr("class", "node")
+            .attr("class", function (d, i) {
+                return "node series series-" + i;
+            })
             .attr("opacity", 0)
             .call(this._selection.enter.bind(this._selection))
             .on("click", function (d) {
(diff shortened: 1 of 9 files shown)


packages/chart/src/Bubble.ts
```diff
@@ -70,7 +70,9 @@ export class Bubble extends SVGWidget {
 
         //  Enter  ---
         node.enter().append("g")
-            .attr("class", "node")
+            .attr("class", function (d, i) {
+                return "node series series-" + i;
+            })
```
GordonSmith: Can you change these to the "shorthand" version:
```
    .attr("class", (d, i) => "node series series-" + i)
```

GordonSmith:
Closing in favor of https://github.com/hpcc-systems/Visualization/pull/2380

PR #2366, opened 2018-01-31
GH-2365 Add Phosphor TabPanel

GordonSmith:
Fixes GH-2365


GordonSmith:
@jbrundage Please Review

packages/phosphor/src/TabPanel.ts
```diff
@@ -0,0 +1,87 @@
...
+            if (wa.widget === widget) {
+                retVal = wa;
+                return true;
+            }
+            return false;
+        });
+        return retVal;
+    }
+
+    addWidget(widget: SVGWidget | HTMLWidget, title: string) {
```
jbrundage: Shouldn't this be `widget: Widget`?

jbrundage:
Seems like there should be phosphor/src/Panel.ts that DockPanel.ts and TabPanel.ts can extend.

packages/phosphor/src/TabPanel.ts
GordonSmith: Not sure why I went with that choice, I didn't test the CanvasWidget...  But will change in the future if needed.

Issue #2372, opened 2018-01-31
FontAwesome + rollup bundles


GordonSmith:
Need to replace the relative URLs in the css with a @hpcc-js/common reference.

jbrundage:
Maybe SASS integration should take place before this change?

GordonSmith:
I suspect this issue will still prevail (unless there are sass files in the font-awesome project.

A relatively simple search/replace with the rollup bundle will fix it.

jbrundage:
font-awesome does offer Less and Sass support

GordonSmith:
Expired - please reopen if needed.

PR #2373, opened 2018-02-01
GH-2372 Remove FA from common bundle

GordonSmith:
Since the fonts will get pulled relative to the host web page and not the
bundle, it can never work in a "prod" environment.


jbrundage:
Actually... here's a pretty interesting way of serving up a font "file" as a base64 data URI:
https://github.com/keithorbit/font-awesome-base64

GordonSmith:
That works - we do something similar for other assets already (via webpack) but I hadn't been able to get it to work with the font files because of the way they were imported).

PR #2380, opened 2018-02-04
GH-2342 Highlight series from legend

GordonSmith:
Adds series classes on chart package widgets
Highlight series when mouse over legend items

Fixes GH-2342

### DIFF
--- a/packages/common/src/Class.ts
+++ b/packages/common/src/Class.ts
@@ -62,5 +62,14 @@ export class Class {
         };
         return this;
     }
+
+    cssTag(id: string): string {
+        return id.replace(/[^a-z0-9]/g, (s) => {
+            const c = s.charCodeAt(0);
+            if (c === 32) return "-";
+            if (c >= 65 && c <= 90) return "_" + s.toLowerCase();
+            return "_0x" + c.toString(16);
+        });
+    }
 }
 Class.prototype._class = "common_Class";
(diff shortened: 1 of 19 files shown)


GordonSmith:
@jbrundage please review

packages/common/src/Class.ts
```diff
@@ -62,5 +62,14 @@ export class Class {
...
         return this;
     }
+
+    cssTag(id: string): string {
+        return id.replace(/[^a-z0-9]/g, (s) => {
+            const c = s.charCodeAt(0);
+            if (c === 32) return "-";
+            if (c >= 65 && c <= 90) return "_" + s.toLowerCase();
+            return "_0x" + c.toString(16);
+        });
```
jbrundage: Might be better to simply `return btoa(id);` to avoid duplicate return values for edge cases (identical column names with different capitalization).

GordonSmith: True - or I could do a hashsum (is btoa very large?)

But in this scheme `Hello` and `HeLLo` do not clash as the later becomes:  `He_l_lo` - but I guess that could still clash...

jbrundage: The length of `btoa` response is determined by the content, but there are some undesirable characters like "=" and "+".

So it would need to be something like... `return btoa(id).split('=').join('_').split('+').join('-')`

PR #2397, opened 2018-02-08
GH-2379 Comms use fetch instead of request

GordonSmith:
Fixes GH-2379-Fetch

### DIFF
--- a/packages/common/src/Class.ts
+++ b/packages/common/src/Class.ts
@@ -64,7 +64,7 @@ export class Class {
     }
 
     cssTag(id: string): string {
-        return id.replace(/[^a-z0-9]/g, (s) => {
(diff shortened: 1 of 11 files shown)


GordonSmith:
@jbrundage please review

GordonSmith:
@jbrundage bump

packages/common/src/Class.ts
```diff
@@ -64,7 +64,7 @@ export class Class {
     }
 
     cssTag(id: string): string {
-        return id.replace(/[^a-z0-9]/g, (s) => {
+        return ("" + id).replace(/[^a-z0-9]/g, (s) => {
```
jbrundage: Why was this necessary?

GordonSmith: I think its for the case where id is a number.

PR #2465, opened 2018-04-11
update git ignore to ignore intellij configuration files

jchambers-ln:
DSP development team is primary using intelliJ now for development so I've added to the .gitignore file .idea which will prevent any of the IDE project files from being committed


GordonSmith:
@jchambers-ln can you target "candidate-1.18.x" and I will up merge through the active branches.

jchambers-ln:
new pull issued

Issue #2475, opened 2018-04-17
ChartPanel + Legend should use Fields


GordonSmith:
...instead of columns

PR #2476, opened 2018-04-17
GH-2475 ChartPanel + Legend use Fields

GordonSmith:
Switch ChartPanel and Legend to use Fields instead of Columns.
Legend should not show fields that start with "__".

Fixes GH-2475


GordonSmith:
@jbrundage please review

PR #2504, opened 2018-04-27
GH-2503 Tests using old version of packages

GordonSmith:
Fixes GH-2503


GordonSmith:
@jbrundage please review

GordonSmith:
"*" - effectively yes.  The same is true if the versions match, but if they don't then the tests would get an older version...

PR #2549, opened 2018-05-10
feat: Prevent Render when hidden

GordonSmith:
Fixes GH-2547

### DIFF
--- a/packages/common/src/Widget.ts
+++ b/packages/common/src/Widget.ts
@@ -478,6 +478,12 @@ export abstract class Widget extends PropertyExt {
         return this;
     }
 
+    isDOMHidden(): boolean {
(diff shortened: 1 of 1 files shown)


jbrundage:
Calling `render` should render the widget. Adding exceptions within render to prevent that from happening seems unconventional. If a widget's target element is hidden and something tells it to render, then I think the fault lies with the logic that told it to render, no?

When rendering to a canvas element, for example, it may very well be hidden (or even outside the dom), but if the canvas has a height and width it can still be drawn on and used as image data. Thoughts?

GordonSmith:
The problem is when an ancestor DOM id Display:none or if the DOM isn't in the page as you point out, then any call get getBBox of GetClientRect will fail causing issues when the widget does get around to displaying.

Issue #2558, opened 2018-05-15
Table auto fit columns


GordonSmith:
Guesstimate suitable column widths on initial load.

PR #2560, opened 2018-05-15
feat: Auto size columns on table load.

GordonSmith:
Fixes GH-2558

### DIFF
--- a/packages/dgrid/src/Table.ts
+++ b/packages/dgrid/src/Table.ts
@@ -1,6 +1,20 @@
 ﻿import { hashSum } from "@hpcc-js/util";
 import { Common } from "./Common";
 
+function charW(w, c) {
+    if (c === "W" || c === "M") w += 15;
+    else if (c === "w" || c === "m") w += 12;
+    else if (c === "I" || c === "i" || c === "l" || c === "t" || c === "f") w += 4;
+    else if (c === "r") w += 8;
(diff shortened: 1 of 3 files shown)


packages/dgrid/src/Table.ts
```diff
@@ -54,10 +68,31 @@ export class Table extends Common {
...
+                const sampleData = [];
+                for (let i = 0; i < Math.min(3, data.length); ++i) {
+                    sampleData.push(data[i][column.idx]);
+                }
+                this.guessWidth(column.children, sampleData);
+            } else {
+                column.width = data.reduce((prevVal: number, row) => {
+                    const cell = ("" + row[column.idx]).trim();
+                    return Math.max(prevVal, textWidth(cell));
+                }, textWidth("" + column.label));
```
jbrundage: This should run faster:
`column.width = Math.max.apply(undefined, data.map(row=>textWidth(("" + row[column.idx]).trim())))`

jbrundage: Also, if we want a fast and precise way to measure the text... 
```javascript
measureText(font,text){
    this.text_ctx = this.text_ctx ? this.text_ctx : document.createElement('canvas').getContext('2d');
    this.text_ctx.font = font;
    return this.text_ctx.measureText(text).width;
}
measureText("12px Arial","some text to measure");
```

Issue #2567, opened 2018-05-16
Add better font size detection


GordonSmith:
As per your suggestion:
```
measureText(font,text){
    this.text_ctx = this.text_ctx ? this.text_ctx : document.createElement('canvas').getContext('2d');
    this.text_ctx.font = font;
    return this.text_ctx.measureText(text).width;
}
measureText("12px Arial","some text to measure");
```

PR #2592, opened 2018-05-24
fix: Optimize getBBox calls

GordonSmith:
Use calculated values in preference to DOM calls.

Fixes GH-2567

### DIFF
--- a/packages/common/src/Icon.ts
+++ b/packages/common/src/Icon.ts
@@ -19,6 +19,16 @@ export class Icon extends SVGWidget {
         this._faChar = new FAChar();
     }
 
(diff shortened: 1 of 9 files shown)


packages/common/src/Icon.ts
```diff
@@ -19,6 +19,16 @@ export class Icon extends SVGWidget {
         this._faChar = new FAChar();
     }
 
+    getBBox(refresh = false, round = false) {
+        const diameter = this.diameter();
+        return {
+            x: - diameter / 2,
+            y: - diameter / 2,
```
jbrundage: `x: this._pos.x` and  `y: this._pos.y`?

PR #2625, opened 2018-06-12
fix: Better bubble text clipping

GordonSmith:


GordonSmith:
In "Text" when you assign it a fixed width.

Issue #2618, opened 2018-06-08
GMapPins + click -> StreetView


GordonSmith:
Not working when used inside a marshaller and that pin click also drives another widget

jbrundage:
Adding Control+LeftClick to enter StreetView would resolve this collision of functionality. Thoughts?

PR #2648, opened 2018-06-19
feat(map): Ctrl+Click gives priority to streetview

jbrundage:
Fixes GH-2618

@GordonSmith please review

### DIFF
--- a/src/map/GMapPin.js
+++ b/src/map/GMapPin.js
@@ -13,7 +13,15 @@
         this._pins = new Pins()
             .columns(["lat", "long", "ext"])
             .on("click", function (row, col, sel) {
-                context.click(context.rowToObj(row.ext.origRow), "", sel);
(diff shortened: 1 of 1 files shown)


src/map/GMapPin.js
```diff
@@ -13,7 +13,17 @@
         this._pins = new Pins()
             .columns(["lat", "long", "ext"])
             .on("click", function (row, col, sel) {
-                context.click(context.rowToObj(row.ext.origRow), "", sel);
+                if (window.event.ctrlKey) {
+                    if (context.streetViewOnClick()) {
```
GordonSmith: Change to:  `if (window.event.ctrlKey && context.streetViewOnClick())`

Then the previous behavior will be the same when streetViewOnClick() is false.

GordonSmith:
@jbrundage commit message should probably be `fix(gmap)`?

PR #2671, opened 2018-06-29
chore(form): form package params as interface

jbrundage:


packages/form/src/Button.ts
```diff
@@ -36,16 +36,27 @@ export class Button extends HTMLWidget {
...
+    name_exists(): boolean;
+    label_exists(): boolean;
+    value_exists(): boolean;
+    validate_exists(): boolean;
+}
+
+Button.prototype.publish("name", null, "string");
+Button.prototype.publish("label", null, "string");
+Button.prototype.publish("value", null, "string");
+Button.prototype.publish("validate", null, "string");
```
GordonSmith: I think these PPs are already published in IInput (which gets mixed-in via line #51/41)?

GordonSmith:
Yes - its telling you that your publishing a property which has already been published for the item in question ("type" I think).

packages/form/src/InputRange.ts
```diff
@@ -62,25 +62,30 @@ export class InputRange extends HTMLWidget {
...
 
+export interface InputRange {
+    name(): string;
+    name(_: string): this;
+    type(): string;
+    type(_: string): this;
+    label(): string;
+    label(_: string): this;
+    value(): any;
+    value(_: any): this;
```
GordonSmith: Should be an array of two items

packages/form/src/InputRange.ts
```diff
@@ -62,25 +62,30 @@ export class InputRange extends HTMLWidget {
...
+    name_exists(): boolean;
+    label_exists(): boolean;
+    value_exists(): boolean;
+    validate_exists(): boolean;
+    inlineLabel_exists(): boolean;
+}
+
 InputRange.prototype.publish("type", "text", "set", "InputRange type", ["number", "date", "text", "time", "datetime", "hidden"]);
 InputRange.prototype.publish("inlineLabel", null, "string", "InputRange Label", null, { optional: true });
-InputRange.prototype.publish("value", ["", ""], "array", "Input Current Value", null, { override: true });
```
GordonSmith: Removing this is probably ok with the correct value prototype above.  But you will need to add a suitable default to value in the constructor.

jbrundage:
@GordonSmith I've made the requested changes to InputRange.ts

jbrundage:
@GordonSmith ready for re-review

packages/form/src/InputRange.ts
GordonSmith: Specifically:
```
   value(): [string|number:string|number];
    value(_: [string|number:string|number]): this;
```

jbrundage:
@GordonSmith I've made the requested change

PR #2700, opened 2018-07-18
feat(marshaller): Add json/TreeFormat.ts

jbrundage:
@GordonSmith this TreeFormat.ts might not necessarily belong in marshaller package, but it seems true to the definition of "marshalling". It's just a simple script I wrote that converts any/all json shapes into this format that can be consumed by any of the tree widgets:
`export interface TreeDataFormat {
    label: string;
    children?: TreeDataFormat[];
    size?: number;
    weight?: number;
}`

### DIFF
--- a/packages/marshaller/src/json/TreeFormat.ts
+++ b/packages/marshaller/src/json/TreeFormat.ts
@@ -1,0 +1,39 @@
+
+export function TreeFormat(json: string | object): TreeDataFormat { // Maybe write interfaces for all of our expected data formats?
+    return obj_to_tree({label: "root"}, typeof json === "string" ? JSON.parse(json) : json);
+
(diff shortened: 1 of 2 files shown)


packages/marshaller/src/json/TreeFormat.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    function obj_to_tree(visitor, _data) {
+        switch (typeof _data) {
+            case "object":
+                if (_data !== null) {
+                    if (_data instanceof Array) {
+                        visitor.children = _data.map((row, row_idx) => {
+                            return this.obj_to_tree({label: row_idx}, row);
+                        });
+                    } else {
```
GordonSmith: Move null test to here as an `else if` (is typeof null === "object")?

PR #2710, opened 2018-07-24
fix(phosphor): Remove "crypto" dependency

GordonSmith:


packages/phosphor-shim/rollup.config.js
```diff
@@ -34,15 +35,18 @@ export default {
         name: pkg.name
     }],
     plugins: [
+        replace({
+            "(typeof require !== 'undefined' && require('crypto')) ||": "/*---@hpcc-js (typeof require !== 'undefined' && require('crypto')) || @hpcc-js---*/",
```
jbrundage: Wow, what's going on here?

GordonSmith: In phosphor.js they are calling `require('crypto')` without declaring `crypto` in package.json (and it can only get to that code _if_ its being run in Node.js) - this code strips out that code.

PR #2711, opened 2018-07-24
feat: Add performance logging to ScopedLogging

jbrundage:
logger.perfStart("measurement_name") and logger.perfEnd("measurement_name") can be placed throughout code and logger.getPerfData() retrieves the performance measurements in a data format that timeline_MiniGantt can consume.

### DIFF
--- a/packages/util/src/logging.ts
+++ b/packages/util/src/logging.ts
@@ -147,6 +147,8 @@ export const logger = Logging.Instance();
 
 export class ScopedLogging {
     protected _scopeID: string;
+    protected _performance_start_times: {[key: string]: number[]} = {};
+    protected _performance_end_times: {[key: string]: number[]} = {};
 
     constructor(scopeID: string) {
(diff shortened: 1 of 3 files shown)


packages/util/src/logging.ts
```diff
@@ -193,6 +195,28 @@ export class ScopedLogging {
...
+        this._performance_start_times[name].push(Date.now());
+    }
+    perfEnd(name: string) {
+        if (typeof this._performance_end_times[name] === "undefined")this._performance_end_times[name] = [];
+        this._performance_end_times[name].push(Date.now());
+    }
+
+    getPerfData(ranges_only?: boolean) {
+        const ret: any[] = [];
+        Object.keys(this._performance_end_times).forEach(name => {
```
GordonSmith: Object.keys won't work in IE

GordonSmith:
Also you should add some unit tests for this (same is true for everything in util).

PR #2723, opened 2018-07-31
feat(dashy): Add Thumbnails inplace of chartType

jbrundage:


jbrundage:
image

GordonSmith:
This looks great.  Two comments:
1. They might be better in a drop list?
2. In-lining the thumbs in the CSS will really bloat all apps that includes a dashy generated dashboards (given the current package dependencies) - I suspect the solution is to remove the dependency of "other" from the marshaller package (which probably means moving the "dashy" sources into "other").

jbrundage:
I've added the HSCROLL:

image

With a bit of the next row showing (a visual indicator that there's content that can scroll)

PR #2725, opened 2018-08-01
fix(d3-event): d3-event is a global instance.

GordonSmith:
d3-event is a global instance and needs to be shared between packages.


GordonSmith:
Some of it anyway and move it into "d3-shim".

PR #2728, opened 2018-08-01
fix(graph): Remove inline status messages.

GordonSmith:

### DIFF
--- a/packages/graph/src/Graph.ts
+++ b/packages/graph/src/Graph.ts
@@ -45,7 +45,6 @@ export class Graph extends SVGZoomWidget {
     protected svgC;
     protected svgE;
     protected svgV;
-    protected svgStatus;
 
     constructor() {
         super();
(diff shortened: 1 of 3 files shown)


packages/graph/src/Graph.ts
```diff
@@ -467,9 +445,9 @@ export class Graph extends SVGZoomWidget {
             .attr("class", "graphEdge")
             .style("opacity", 1e-6)
             .on("click.selectionBag", function (d) {
-                context._selection.click(d, d3Event);
+                context._selection.click(d, event);
```
jbrundage: Is this going to work the same way d3Event did?

packages/graph/src/Graph.ts
```diff
@@ -950,6 +930,10 @@ export class Graph extends SVGZoomWidget {
     tooltip;
     tooltipHTML: (_) => string;
     tooltipFormat: (_) => string;
+
+    //  Progess Events ---
+    progress(what: "start" | "end" | "layout-start" | "layout-tick" | "layout-end") {
+    }
```
jbrundage: Is the idea to override this method when you'd want to gauge performance?

packages/graph/src/Graph.ts
GordonSmith: This should not have made it into this PR - good catch.

packages/graph/src/Graph.ts
GordonSmith: You would add an `on("progess", what=>{})` method to the hosting app.

PR #2737, opened 2018-08-06
feat(form): Add node build_hpccjs.js script

jbrundage:
Running 'node build_hpccjs.js' in the root directory concatenates all files specified within build_hpccjs.js and creates one all inclusive 'hpccjs.js' in the root directory.

'init_hpccjs.js' is appended to the end of the output of build_hpccjs.js and creates a global 'hpccjs' object with references to each loaded package.

### DIFF
--- a/build_hpccjs.js
+++ b/build_hpccjs.js
@@ -1,0 +1,33 @@
+const fs = require('fs');
+const os = require('os');
+const files = [
+    "packages/util/dist/index.js",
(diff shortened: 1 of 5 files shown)


GordonSmith:
Rather than getting into the JS Bundler game, you could just re-use a rollupjs?  See:  https://github.com/GordonSmith/Visualization/tree/ROLL_TEST/apps/tmp2

Issue #2745, opened 2018-08-07
Timeline render is a performance bottleneck


jbrundage:

GordonSmith:
Some perf improvements included in:  https://github.com/hpcc-systems/Visualization/pull/2783 please reopen if needed.

PR #2748, opened 2018-08-07
feat(timeline): Add useCachedHeight to EntityPin

jbrundage:
Fixes GH-2745

Using the cached height in EntityPin's calcHeight method significantly improves MiniGantt's render performance.


GordonSmith:
For a lot of the core widgets you can call "getBBox" without rendering the widget in question, see:
https://github.com/hpcc-systems/Visualization/blob/master/packages/common/src/Text.ts#L13
https://github.com/hpcc-systems/Visualization/blob/master/packages/common/src/TextBox.ts#L54
https://github.com/hpcc-systems/Visualization/blob/master/packages/common/src/Icon.ts#L22

So if you add a method called "getBBox" here and do something similar you will get a nice perf boost without worrying about stale cache issues...

PR #2750, opened 2018-08-09
fix(comms): Query.attach should not return a promise.

GordonSmith:


GordonSmith:
No - its was just inconsistent with other Attach methods (and I picked up on it when updating the roxie tutorial / test.

PR #2777, opened 2018-08-15
feat(comms): Export hookSend from marshaller

GordonSmith:
If a consumer is using the marshaller and a different version of comms, then they had no way of hooking the marshaller comms layer.


GordonSmith:
If you have marshaller@2.0.0 dependent on comms@2.0.0 but manually include a comms@1.0.0 elsewhere in your project, then they have different instances of hookSend.  The issue was you had no easy way to get the appropriate hookSend that the marshaller uses.

This is only ever an issue with "global" things, of which we don't have many.

PR #2793, opened 2018-08-22
feat(gallery): Add 1 Gauge sample

jbrundage:


GordonSmith:
Can you move this into the "Widget" sample folder (and rename) - there is already a Gauge example in there...

jbrundage:
@GordonSmith this doesn't belong in the Gauge folder? I thought we were going to have multiple examples per widget in their respective folders?

GordonSmith:
Yes...
But it should be in "widget/gauge" not "chart/gauge".

PR #2794, opened 2018-08-22
feat(bubble): Add minRadius and maxRadius

jbrundage:
d3-interpolate allows the min and max radius to be applied


packages/chart/src/Bubble.ts
```diff
@@ -65,13 +66,21 @@ export class Bubble extends SVGWidget {
         if (this.useClonedPalette()) {
             this._palette = this._palette.cloneNotExists(this.paletteID() + "_" + this.id());
         }
-
+        console.log('this.cloneData()',this.cloneData());
```
GordonSmith: Remove console.log

packages/chart/src/Bubble.ts
```diff
@@ -65,13 +66,21 @@ export class Bubble extends SVGWidget {
...
         const root = d3Hierarchy({ children: this.cloneData() })
             .sum(function (d) { return d[1]; })
             .sort(function (b, a) {
                 return a[1] < b[1] ? -1 : a[1] > b[1] ? 1 : 0;
             })
             ;
+        if(this.minRadius() !== null && this.maxRadius() !== null){
```
GordonSmith: Should be `this.minRadius_exists()` and `this.maxRadius_exists()`

packages/chart/src/Bubble.ts
```diff
@@ -65,13 +66,21 @@ export class Bubble extends SVGWidget {
...
+            const interp = d3Interpolate(this.minRadius(),this.maxRadius());
+            const minVal = Math.min.apply(undefined, root.children.map(n=>n.value));
+            const maxVal = Math.max.apply(undefined, root.children.map(n=>n.value));
+            this.d3Pack.radius(function(n){
+                return n.value === minVal ? context.minRadius() : interp((n.value - minVal)/(maxVal - minVal));
+            });
+        }
```
GordonSmith: Should be able to specify a min on its own (without a max) and similarly for max?  This would make it the same as yAxisDomainLow and yAxisDomainHigh in the XYAxis charts?

packages/chart/src/Bubble.ts
```diff
@@ -175,3 +186,5 @@ Bubble.prototype.mixin(Utility.SimpleSelectionMixin);
 
 Bubble.prototype.publish("paletteID", "default", "set", "Palette ID", Bubble.prototype._palette.switch(), { tags: ["Basic", "Shared"] });
 Bubble.prototype.publish("useClonedPalette", false, "boolean", "Enable or disable using a cloned palette", null, { tags: ["Intermediate", "Shared"] });
+Bubble.prototype.publish("minRadius", null, "number", "Minimum radius of bubble (px)");
+Bubble.prototype.publish("maxRadius", null, "number", "Maximum radius of bubble (px)");
```
GordonSmith: Should have {optional:true} in the ext settings.

packages/chart/src/Bubble.ts
jbrundage: If the minimum radius would have been 50 for example, but the user specifies the maxRadius as 40 (but doesn't specify minRadius), should all of the circles have a radius of 40 (essentially saying "Don't allow them to get any bigger than 40")? Or should it set the largest value's radius at 40 and relatively scale the rest of them (essentially saying "Use 40 for the largest and scale the rest")?

jbrundage: ...because if I'm remembering correctly the yAxisDomainLow/yAxisDomainHigh simply cap the upper and lower ticks, but the data still displays beyond the Low and/or High (they just get clipped off at the edge of the widget).

GordonSmith: Ahh ok - Its the difference between domain low/high (I know all exam results must be between 0 and 100) and physical pixel min and max...
But when you set the constrained min/max doesn't that prevent the bubble chart from using up all the available space?

jbrundage:
@GordonSmith requested changes have been made (except for the ability to specify one and not the other). I couldn't find a clean way to implement that feature. Can this be added later?

packages/chart/src/Bubble.ts
```diff
@@ -152,6 +160,10 @@ export class Bubble extends SVGWidget {
 
     paletteID: { (): string; (_: string): Bubble; };
     useClonedPalette: { (): boolean; (_: boolean): Bubble; };
+    minRadius: { (): number; (_: number): Bubble; };
+    maxRadius: { (): number; (_: number): Bubble; };
+    minRadius_exists: { (): boolean; (_: boolean): Bubble; };
+    maxRadius_exists: { (): boolean; (_: boolean): Bubble; };
```
GordonSmith: should be `maxRadius_exists() => boolean;` (in this "old" stub style)

PR #2800, opened 2018-08-23
chore: Update gallery package.json hpcc-js deps

jbrundage:
This fixes the tree samples.


GordonSmith:
@jbrundage If you run the build the command it will generate the new config file - then if you include that in this commit I can pull and it will start working...

PR #2805, opened 2018-08-24
feat(DockPanel): Optionally show close [x] button

GordonSmith:

### DIFF
--- a/packages/phosphor/src/WidgetAdapter.ts
+++ b/packages/phosphor/src/WidgetAdapter.ts
@@ -24,6 +24,10 @@ export namespace Msg {
     }
 }
 
+export interface IClosable {
+    canClose(e: Widget, wa: WidgetAdapter): boolean;
+}
+
 export class WidgetAdapter extends PWidget {
     protected _owner;
     protected _element;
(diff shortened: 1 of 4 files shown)


packages/phosphor/src/WidgetAdapter.ts
```diff
@@ -34,16 +38,18 @@ export class WidgetAdapter extends PWidget {
     padding: number = 0;
     _width: number = 0;
     _height: number = 0;
+    private _closable: IClosable;
 
-    constructor(owner?: Widget, widget?: Widget | object, lparam: object = {}) {
+    constructor(owner?: Widget, widget?: Widget | object, lparam: object = {}, closable?: IClosable) {
```
jbrundage: IMO `closable` should be a boolean and the `canClose` logic should take place as the WidgetAdapter is instantiated.

GordonSmith: Can you mock and example (not sure I follow 100%)

jbrundage: I mean...
```javascript

```javascript
addWidget(widget: Widget, title: string, location: PhosphorDockPanel.InsertMode = "split-right", refWidget?: Widget, closable?: boolean) {
...and...
constructor(owner?: Widget, widget?: Widget | object, lparam: object = {}, closable?: boolean) {
```
...then the `closable` option can be toggled whenever addWidget is used. Rather than passing a callback.

GordonSmith: Need the callback to "optionally" prevent it from closing (in my use case, I ask the user if its ok to close).  But it could be a "boolean | ICallback" that would make sense.

GordonSmith:
Changes made as requested.

PR #2806, opened 2018-08-24
feat: Add RadialBar.ts

jbrundage:

### DIFF
--- a/packages/chart/src/RadialBar.ts
+++ b/packages/chart/src/RadialBar.ts
@@ -1,0 +1,167 @@
+import { INDChart, ITooltip } from "@hpcc-js/api";
+import { SVGWidget, Utility } from "@hpcc-js/common";
+import { interpolate as d3interpolate } from "d3-interpolate";
+import { scaleLinear as d3scaleLinear } from "d3-scale";
+
+import { arc as d3arc } from "d3-shape";
+import "../src/RadialBar.css";
+export class RadialBar extends SVGWidget {
+    protected svg;
+    protected arc;
+    protected arcs;
+    protected arcWidth;
+    protected axialAxis;
+    protected radialAxis;
+    constructor() {
+        super();
+        INDChart.call(this);
+        ITooltip.call(this);
+        Utility.SimpleSelectionMixin.call(this);
+    }
+    enter(domNode, element) {
+        super.enter(domNode, element);
+        this._selection.widgetElement(element);
+        this
+            .tooltipHTML(d => {
+                return this.tooltipFormat({ label: d[0], value: d[1] });
+            })
+            ;
+        this.axialAxis = element.append("g")
+            .attr("class", "a axis")
+            ;
+        this.radialAxis = element.append("g")
+            .attr("class", "r axis")
+            ;
+        this.arcs = element.append("g")
+            .attr("class", "data")
+            ;
+    }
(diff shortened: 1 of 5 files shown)


jbrundage:
image

packages/chart/src/RadialBar.ts
```diff
@@ -0,0 +1,167 @@
...
+    }
+    enter(domNode, element) {
+        super.enter(domNode, element);
+        this._selection.widgetElement(element);
+        this
+            .tooltipHTML(d => {
+                return this.tooltipFormat({ label: d[0], value: d[1] });
+            })
+            ;
+        this.svg = this.element().append('g');
```
GordonSmith: this.svg is probably not needed - you can just use "element" as the root node?

packages/chart/src/RadialBar.ts
```diff
@@ -0,0 +1,167 @@
...
+    tweenDuration(_: number): this;
+    tickCount(): number;
+    tickCount(_: number): this;
+    arcPadding(): number;
+    arcPadding(_: number): this;
+    arcMinRadius(): number;
+    arcMinRadius(_: number): this;
+}
+RadialBar.prototype.publish("paletteID", "default", "set", "paletteID", RadialBar.prototype._palette.switch());
+RadialBar.prototype.publish("radiusPadding", 40, "number", "Padding between chart edge and container edge (px)", RadialBar.prototype._palette.switch());
```
GordonSmith: Should not have:  RadialBar.prototype._palette.switch()

packages/chart/src/RadialBar.ts
```diff
@@ -0,0 +1,167 @@
...
+RadialBar.prototype.publish("rowDelay", 100, "number", "Delay between the start of each row animation (ms)");
+RadialBar.prototype.publish("tweenDuration", 1000, "number", "Duration of a row's animation (ms)");
+RadialBar.prototype.publish("maxRange", 2 * Math.PI, "number", "Maximum angular length of a bar (radians)");
+RadialBar.prototype.publish("maxDomainMultiplier", 1.62, "number", "Multiplied by the highest value to determine the maximum arc length (>= 1.0)");
```
GordonSmith: Not sure what this is doing?  Looks like its multiplying the max "value" not domain by 1.62?  I suspect its to prevent the bar from going a full 360 degrees?

jbrundage: Right, it provides a buffer for the upper limit of the domain. `.maxDomainMultiplier(1)` would display the largest value as a 360 arc.

packages/chart/src/RadialBar.ts
jbrundage: I'm not sure what you mean? This is declared just like the other chart widgets?

jbrundage: *smack forehead* now I see what you mean...

GordonSmith:
@jbrundage When I tested this I found a lot of issues:
* Corrupted view when resized or rendered more than once (none of the d3 "selectAll" patterns looked correct)
* Selection not working
* Animation always started from 0 even when data was just changed
 
I ended up refactoring here:
https://github.com/hpcc-systems/Visualization/pull/2807 it might be worthwhile reviewing my changes in depth.

PR #2809, opened 2018-08-26
fix(ddl): Merge Dreas changes

GordonSmith:
Enable strictNullChecks to highlight omittable fields.
Fix DDL1 schema.


packages/ddl-shim/src/upgrade.ts
```diff
@@ -433,6 +439,16 @@ class DDLUpgrade {
         }
     }
 
+    rule2condition(_: DDL1.IFilterRule): DDL2.IMappingConditionType {
```
jbrundage: Maybe this should be more sensibly named?

GordonSmith: "filterRule2MappingConditionType"?

Issue #2842, opened 2018-09-06
chartPanel + Custom colours


GordonSmith:
Need to test and ensure its working as expected with a single:
```
overriderMethod('fillColor', ...)
```
on the chartPanel.

PR #2847, opened 2018-09-07
fix(Legend): Legend should use "fillColor" when available

GordonSmith:
Fixes GH-2842

### DIFF
--- a/packages/layout/src/Legend.ts
+++ b/packages/layout/src/Legend.ts
@@ -138,6 +138,25 @@ export class Legend extends SVGWidget {
         return Palette.ordinal("default");
     }
 
+    getPaletteType() {
+        return this.getPalette().type();
+    }
+
+    fillColorFunc() {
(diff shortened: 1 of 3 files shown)


packages/layout/src/Legend.ts
```diff
@@ -138,6 +138,25 @@ export class Legend extends SVGWidget {
...
+    getPaletteType() {
+        return this.getPalette().type();
+    }
+
+    fillColorFunc() {
+        const widget = this.getWidget();
+        if (widget && widget.fillColor) {
+            return widget.fillColor;
+        }
+        const palette = Palette.ordinal("default");
```
jbrundage: Why not use the widget's paletteID if it exists?
`const palette = Palette.ordinal(widget && widget.paletteID ? widget.paletteID() : "default");`

GordonSmith: Good point!

PR #2852, opened 2018-09-11
fix(dgrid): Maintain scroll pos and pagination on refresh.

GordonSmith:
Add watch script to dgrid-shim
Add new table samples.


GordonSmith:
@jbrundage please review

GordonSmith:
As it stands I don't think we can do that for publish properties (without breaking the PropertyEditor)?

PR #2856, opened 2018-09-12
feat: Widget parameters as constructor arg

jbrundage:
Auto-render if any parameters are specified in this way

### DIFF
--- a/packages/common/src/Widget.ts
+++ b/packages/common/src/Widget.ts
@@ -64,8 +64,11 @@ export abstract class Widget extends PropertyExt {
 
     protected _overlayElement;
 
-    constructor() {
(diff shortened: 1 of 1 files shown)


GordonSmith:
What is the motivation for this?

GordonSmith:
Targeting wrong branch for PR

jbrundage:
What do you think about this concept? We aren't currently using the constructor arguments for anything so I threw this PR out there as a conversation starter. Passing the publish parameter values as options in this way would actually simplify the code generation process. Thoughts?

GordonSmith:
I am going to close this for now as would need to think about it (for the 3.x time frame) - open an issue for the discussion.

PR #2857, opened 2018-09-17
fix: Appropriate Tooltip direction now determined

### LINKED ISSUE #2833: Tooltip and Edge clipping
Investigate improved schemes to avoid edge clipping.

jbrundage:
Fixes #2833


jbrundage:
image

packages/api/src/ITooltip.css
```diff
@@ -6,6 +6,7 @@
   color: #fff;
   border-radius: 2px;
   pointer-events: none !important;
+  white-space: nowrap;
```
GordonSmith: This is causing text to spill out of the "bubble":
image

jbrundage: I've removed the nowrap line. Its only purpose was to prevent the "north" direction from being chosen when the "east" or "west" would have looked nicer.

jbrundage: I've got a version that always renders it nicely, gotta get it into PR shape though

jbrundage:
I've added those features, will PR shortly

jbrundage:
@GordonSmith I've updated the PR with your requested changes. The tooltips now display nicely in every situation (assuming the widget has a larger width and height than the tooltip itself).

GordonSmith:
And:
tooltip2

jbrundage:
It reverts to the direction with the least overlap... then determines how much to add or subtract from the tooltip's `top` and `left` values to fit within the containing element.

jbrundage:
@GordonSmith I believe this is working as expected now

GordonSmith:
Still has issues:
tooltip3

GordonSmith:
tooltip4

PR #2877, opened 2018-09-21
feat(tooltip): Added key value tooltip feature

jbrundage:
Also added a use-case example to Graph.les_miz


jbrundage:
image

packages/api/src/ITooltip.ts
```diff
@@ -123,6 +123,20 @@ export abstract class ITooltip extends Widget {
         }
     }
 
+    tooltipKeyValueFormat(obj) {
+        const _key_color = this.tooltipLabelColor() ? this.tooltipLabelColor() : this.tooltipSeriesColor();
```
GordonSmith: Should use `tooltipLabelColor_exists` or it will fail on color "0"

packages/graph/src/Graph.ts
```diff
@@ -448,10 +448,42 @@ export class Graph extends SVGZoomWidget {
...
+                return "";
+            });
+        }
+    }
+
     update(domNode, element) {
         super.update(domNode, element);
         this.tooltip.hide();
 
+        this.setTooltipType();
```
GordonSmith: Can't do this here as the user may have already provided their own tooltip function.

packages/graph/src/Graph.ts
```diff
@@ -448,10 +448,42 @@ export class Graph extends SVGZoomWidget {
...
+        } else {
+            this.tooltipHTML(function (d: any) {
+                if (d instanceof Subgraph) {
+                    return `<p style="text-align:center">${d.title().replace("\n", "<br>")}</p>`;
+                } else if (d instanceof Vertex) {
+                    return `<p style="text-align:center">${d.text().replace("\n", "<br>")}</p>`;
+                } else if (d instanceof Edge) {
+                    return `<p style="text-align:center">${d.text().replace("\n", "<br>")}</p>`;
+                }
+                return "";
```
GordonSmith: This is duplicate code...

packages/graph/src/Graph.ts
```diff
@@ -448,10 +448,42 @@ export class Graph extends SVGZoomWidget {
...
+                    return `<p style="text-align:center">${d.title().replace("\n", "<br>")}</p>`;
+                } else if (d instanceof Vertex) {
+                    return context.tooltipKeyValueFormat({
+                        label: d.text().replace("\n", "<br>"),
+                        connections: context.data().edges.filter(e => e.targetVertex().id() === d.id() || e.sourceVertex().id() === d.id()).length
+                    });
+                } else if (d instanceof Edge) {
+                    return `<p style="text-align:center">${d.text().replace("\n", "<br>")}</p>`;
+                }
+                return "";
```
GordonSmith: You could make this the default option in the constructor?

packages/graph/src/Graph.ts
```diff
@@ -944,6 +977,11 @@ export class Graph extends SVGZoomWidget {
 Graph.prototype._class += " graph_Graph";
 Graph.prototype.implements(IGraph.prototype);
 Graph.prototype.implements(ITooltip.prototype);
+export interface Graph {
+    useDetailedTooltip(): boolean;
+    useDetailedTooltip(_: boolean): this;
+}
+Graph.prototype.publish("useDetailedTooltip", false, "boolean");
```
GordonSmith: This paradigm doesn't work - as the whole purpose of the tooltipHtml function is to allow the user to provide their own.

packages/graph/src/Graph.ts
```diff
@@ -448,9 +448,29 @@ export class Graph extends SVGZoomWidget {
...
+            return "";
+        });
+    }
+
     update(domNode, element) {
         super.update(domNode, element);
         this.tooltip.hide();
+        if (this.useDetailedTooltip()) {
+            this.overrideTooltipHTML();
+        }
```
jbrundage: @GordonSmith I've updated it to be optional

packages/graph/src/Graph.ts
GordonSmith: Can you add a description and move it so its below "snapToGrid".

Also will you rebase onto latest candidate-2.x.x as I think it won't merge correctly as it stands (I am bit surprised that GitHub isn't warning about it...

jbrundage:
@GordonSmith updated with a sample
image

jbrundage:
BTW this sample exposed some odd BubbleXY behavior. I'll add an issue to track it.

packages/api/src/ITooltip.ts
```diff
@@ -123,10 +123,25 @@ export abstract class ITooltip extends Widget {
...
+        switch (this.tooltipStyle()) {
+            case "none":
+                break;
+            default:
+                return Object.keys(obj).map(key => _keyValueSpans(key, obj[key])).join("<br/>");
+        }
+        function _keyValueSpans(_key, _val) {
+            return `<span style='color:${_key_color}'>${_key}</span>:  <span style='color:${_value_color}'>${_val}</span>`;
+        }
+    }
```
GordonSmith: There is a similar implementation to this here:  https://github.com/hpcc-systems/Visualization/blob/master/demos/imdb/src/IMDBVertex.ts#L70-L86 
With better alignment - I am not sure you can support the "title" feature in the general case though?

jbrundage: @GordonSmith I've updated the PR with your alternative function. It does look nicer.

PR #2884, opened 2018-09-25
fix: Add scale to move within SVGWidget.ts

jbrundage:


jbrundage:
This would allow us to easily scale specific vertices (or tooltips etc) via publish parameter.

image

jbrundage:
@GordonSmith you're right, I fixed it and modified the commit message

PR #2895, opened 2018-09-26
feat: Add value display options for XYAxis etc

### LINKED ISSUE #2707: AVCC: Need ability to show values per column with chart_Line

### LINKED ISSUE #2890: Column/Bar label/value alignment and styling improvements
This may be partially resolved by #2708

jbrundage:
Fixes #2890
Fixes #2707

### DIFF
--- a/packages/chart/src/Column.ts
+++ b/packages/chart/src/Column.ts
@@ -141,6 +141,7 @@ export class Column extends XYAxis {
                     const element = d3Select(this);
                     const domainPos = host.dataPos(dataRow[0]) + (host.yAxisStacked() ? 0 : columnScale(d.column)) + offset;
                     const upperValue = d.value instanceof Array ? d.value[1] : d.value;
+                    const valueText = d.origRow[d.idx];
                     const upperValuePos = host.valuePos(upperValue);
                     const lowerValuePos = host.valuePos(d.value instanceof Array ? d.value[0] : 0);
                     const valuePos = Math.min(lowerValuePos, upperValuePos);
@@ -153,6 +154,7 @@ export class Column extends XYAxis {
                         .style("fill", (d: any) => context.fillColor(d.row, d.column, d.value))
                         ;
 
+                    const valueCentered = context.valueCentered();
                     const dataText = element.selectAll(".dataText").data(context.showValue() ? [`${upperValue}`] : []);
                     const dataTextEnter = dataText.enter().append("g")
                         .attr("class", "dataText")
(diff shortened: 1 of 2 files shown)


packages/chart/src/Column.ts
```diff
@@ -217,11 +232,17 @@ export interface Column {
...
+    centeredValues(_: boolean): this;
+    hideCenteredValueByHeight(): number;
+    hideCenteredValueByHeight(_: number): this;
 }
 
 Column.prototype.publish("paletteID", "default", "set", "Palette ID", () => Column.prototype._palette.switch(), { tags: ["Basic", "Shared"] });
 Column.prototype.publish("useClonedPalette", false, "boolean", "Enable or disable using a cloned palette", null, { tags: ["Intermediate", "Shared"] });
 Column.prototype.publish("showValue", false, "boolean", "Show Value in column");
+Column.prototype.publish("centeredValues", false, "boolean", "Show Value in center of column");
+Column.prototype.publish("hideCenteredValueByHeight", null, "number", "Hide centered value if the rect height is less than this value");
```
GordonSmith: Remove:
* If `valueCentered` is `false` and there is not enough room to display the text at the top / right of the column / bar, then it should be displayed inside the column / bar
* If `valueCentered` is `true` and there is not enough room to display the text inside the column / bar, then it should be displayed as if `valueCentered` === `false` (IOW at the top / right of the column / bar)

packages/chart/src/XYAxis.ts
```diff
@@ -602,6 +602,13 @@ export class XYAxis extends SVGWidget {
     @publish(26, "number", "maxPointSize")
     maxPointSize: publish<this, number>;
 
+    @publish([], "array", "Array of row indices to display values at their points")
+    showValuesFor: publish<this, number[]>;
```
GordonSmith: Why is this not just `showValues` similar to Columns?

packages/chart/src/Column.ts
jbrundage: This could cause confusion when bars are stacked.

packages/chart/src/XYAxis.ts
jbrundage: I was thinking `showValues` implies a boolean. I added the "For" as an indicator that the input expects an array of column indices. Maybe `showValuesByIndex`?

packages/chart/src/Column.ts
jbrundage: ...and when there are several stacked bar/column rectangles with a small height the labels will overlap with that approach. This implementation works as expected and provides the more flexibility for this situation to be controlled at the application layer.

GordonSmith: Yes it might:
* If stacked AND valueCentered === true AND value doesn't fit, then hide it.
* If stacked AND valueCentered === false then the chart will be confusing (IMO)

GordonSmith:
@jbrundage should be targeting candidate 2.x.x?

jbrundage:
@GordonSmith this is targeting 2.x.x now

Issue #2892, opened 2018-09-26
Add total for all series to Legend widget


jbrundage:
This is a "nice-to-have" for the CC risk product

PR #2905, opened 2018-10-02
feat: Add option to show legend values

### LINKED ISSUE #2887: Add counts to each series in Legend widget

jbrundage:
Fixes #2887

### DIFF
--- a/packages/layout/src/Legend.ts
+++ b/packages/layout/src/Legend.ts
@@ -222,11 +222,19 @@ export class Legend extends SVGWidget {
         const ordinal = d3ScaleOrdinal()
             .domain(dataArr.map(row => row[1]))
             .range(dataArr.map(row => row[0]));
-
+        let total = 0;
         this._legendOrdinal
             .orient(this.orientation())
             .title(this.title())
             .scale(ordinal)
+            .labels(d => {
+                const val = this.data()[d.i].slice(1).reduce((acc, n) => acc + n, 0);
+                const disabled = this.isDisabled(d.domain[d.i]);
+                if (!disabled) {
+                    total += val;
+                }
+                return d.domain[d.i] + (!disabled && this.showSeriesTotal() ? ` (${val})` : "");
+            })
             ;
         this._g.call(this._legendOrdinal);
 
@@ -234,8 +242,20 @@ export class Legend extends SVGWidget {
 
         const bbox = this.getBBox(true, true);
         this._g.attr("transform", `translate(${this.width() / 2 - bbox.width / 2 + 5},${this.height() / 2 - bbox.height / 2})`);
+        const legendCellsBbox = this._g.select(".legendCells").node().getBBox();
+        const legendCellHeight = legendCellsBbox.height / (dataArr.length || 1);
+        const legendTotal = this._g.selectAll(".legendTotal").data(dataArr.length && this.showLegendTotal() ? [total] : []);
+        const firstLabel = this._g.select(".label");
+        legendTotal
+            .enter()
+            .append("text")
+            .classed("legendTotal", true)
+            .merge(legendTotal)
+            .attr("transform", `${dataArr.length ? firstLabel.attr("transform").split(",")[0] : "translate(0"}, ${legendCellsBbox.height + (legendCellHeight / (3 / 2))})`)
+            .text(`Total: ${total}`)
+            ;
+        legendTotal.exit().remove();
     }
-
     updateDisabled(element, dataArr) {
         element
             .style("cursor", "pointer")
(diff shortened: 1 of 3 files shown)


packages/layout/src/Legend.ts
```diff
@@ -340,9 +345,12 @@ export interface Legend {
     rainbowBins(): number;
     rainbowBins(_: number): this;
     rainbowBins_exists: () => boolean;
+    showSeriesValue(): boolean;
+    showSeriesValue(_: boolean): this;
```
GordonSmith: Should this be showSeriesTotal?

jbrundage:
image

packages/layout/src/Legend.ts
```diff
@@ -222,20 +222,51 @@ export class Legend extends SVGWidget {
...
+                    return `${d.domain[d.i]} (${val})`;
+                }
+                return d.domain[d.i];
+            })
             ;
         this._g.call(this._legendOrdinal);
 
         this.updateDisabled(element, dataArr);
 
+        this.removeTotal();
```
GordonSmith: Bleh

packages/layout/src/Legend.ts
```diff
@@ -222,20 +222,51 @@ export class Legend extends SVGWidget {
...
         const bbox = this.getBBox(true, true);
         this._g.attr("transform", `translate(${this.width() / 2 - bbox.width / 2 + 5},${this.height() / 2 - bbox.height / 2})`);
+        if (this.showLegendTotal()) {
+            this.insertTotal(total, bbox);
+        }
```
GordonSmith: bleh 2

GordonSmith: How about:
```
this._g.selectAll(".legend-total").data(this.showLegendTotal() ? [total] : [])
```

GordonSmith:
Also can you slide the total over so it lines up with the other text?

packages/layout/src/Legend.ts
jbrundage: I'll try adding this, but the remove/insert paradigm allows us to use bbox for the placement. I'll try inserting the total as if it was a typical legend "cell" with your suggested selectAll/data pattern.

packages/layout/src/Legend.ts
```diff
@@ -222,20 +222,59 @@ export class Legend extends SVGWidget {
...
         this._g.attr("transform", `translate(${this.width() / 2 - bbox.width / 2 + 5},${this.height() / 2 - bbox.height / 2})`);
+        const legendCellsBbox = this._g.select(".legendCells").node().getBBox();
+        const legendCellHeight = legendCellsBbox.height / (dataArr.length || 1);
+        const legendTotal = this._g.selectAll(".legendTotal").data(this.showLegendTotal() ? [total] : []);
+        legendTotal
+            .enter()
+            .append("text")
+            .classed("legendTotal", true)
+            .merge(legendTotal)
+            .attr("transform", `${this._g.select(".label").attr("transform").split(",")[0]}, ${legendCellsBbox.height + (legendCellHeight / (3 / 2))})`)
```
GordonSmith: Can you test this when there is not data loaded?

jbrundage:
@GordonSmith should be good now. Is this the correct way to check if a d3 selection didn't find any elements?

GordonSmith:
@jbrundage You can add a fixes #2892 to the commit message as well.

GordonSmith:
I think selections have an "empty" method to test if they don't have any references to dom nodes?

packages/layout/src/Legend.ts
```diff
@@ -222,20 +222,42 @@ export class Legend extends SVGWidget {
...
-
+        let total = 0;
         this._legendOrdinal
             .orient(this.orientation())
             .title(this.title())
             .scale(ordinal)
+            .labels(d => {
+                if (this.showSeriesTotal()) {
+                    const val = this.data()[d.i].slice(1).reduce((acc, n) => acc + n, 0);
+                    total += val;
```
GordonSmith: This calculation is needed when either `showLegendTotal` or `showSeriesTotal` is `true`.

packages/layout/src/Legend.ts
```diff
@@ -222,20 +222,42 @@ export class Legend extends SVGWidget {
...
 
         this.updateDisabled(element, dataArr);
 
         const bbox = this.getBBox(true, true);
         this._g.attr("transform", `translate(${this.width() / 2 - bbox.width / 2 + 5},${this.height() / 2 - bbox.height / 2})`);
+        const legendCellsBbox = this._g.select(".legendCells").node().getBBox();
+        const legendCellHeight = legendCellsBbox.height / (dataArr.length || 1);
+        const legendTotal = this._g.selectAll(".legendTotal").data(this.showLegendTotal() ? [total] : []);
+        const firstLabel = this._g.select(".label");
+        if (firstLabel.node()) {
```
GordonSmith: Should be `!firstLabel.empty()` (I think).

GordonSmith: If you moved lines 230 -> 237 into a local variable "legendLabels" then you can use "legendLabels.length" here to test if there are any labels and then change your selection logic to:
```
this._g.selectAll(".legendTotal").data(this.showLegendTotal() && legendLabels.length ? [total] : []);
```

Then you won't need any of this "checking" to see if the legend exists AND it will clean itself up when the data goes from 10 rows to 0 rows...

packages/layout/src/Legend.ts
jbrundage: The pie percentage does change as legend series items are deselected (some sort of previous total value cache would have to be added to support the unchanging % concept). This is something I've been meaning to talk to you about. It doesn't make sense to me that someone would want the pie percentage to remain the same after they've filtered certain things out of the pie widget. By filtering they're essentially saying... "I'm not interested in seeing how this series effects the rest of the data."

GordonSmith: Ok - I can buy that - lets leave this as is now so and see if any users have different preferences.

PR #2906, opened 2018-10-03
feat: Add getPartialCircleDimensions to Utility

jbrundage:
Add degreesToRadians to Utility
Add radiansToDegrees to Utility


jbrundage:
@GordonSmith `Utility.getPartialCircleDimensions` can be used to optimize the placement of the `HalfDonut` and `QuarterDonut` widgets without needing to get the bbox from the DOM.

It returns the bounding height and width of a partial circle:
image

packages/common/src/Utility.ts
```diff
@@ -587,3 +587,36 @@ export function removeHTMLFromString(str: string, div?: HTMLDivElement) {
...
+    if (firstAngle <= Math.PI * 1.0 && lastAngle >= Math.PI * 1.0) point_arr.push([-radius,       0]);
+    if (firstAngle <= Math.PI * 1.5 && lastAngle >= Math.PI * 1.5) point_arr.push([      0, -radius]);
+    if (firstAngle <= Math.PI * 2.0 && lastAngle >= Math.PI * 2.0) point_arr.push([ radius,       0]);
+    point_arr.push([x1, y1]);
+    point_arr.push([x2, y2]);
+    point_arr.push([0, 0]);
+    const minX = Math.min.apply(undefined, point_arr.map(n => n[0]));
+    const minY = Math.min.apply(undefined, point_arr.map(n => n[1]));
+    const maxX = Math.max.apply(undefined, point_arr.map(n => n[0]));
+    const maxY = Math.max.apply(undefined, point_arr.map(n => n[1]));
```
GordonSmith: Would be a little cleaner to use https://github.com/d3/d3-array#extent

packages/common/src/Utility.ts
```diff
@@ -587,3 +587,36 @@ export function removeHTMLFromString(str: string, div?: HTMLDivElement) {
...
+    return {
+        width: maxX - minX,
+        height: maxY - minY
+    };
+}
```
GordonSmith: I think these should be in @hpcc-js/util in a file called geometry?
Also in util, you will see everything has unit tests and a lot of stuff has some API documentation - which is probably a good idea for these types of utility?

GordonSmith:
Quick question - the center point (shown in the image) is not used as part of the bounding box equation correct?

jbrundage:
The center point is used as part of the bounding box equation. Example: 
If start angle is 30 and end angle is 60 we need the center point to represent the top and left bounds.
image

GordonSmith:
I think I would rename it to:  getPartialPieDimensions so?

GordonSmith:
Or Partial Disc ?

packages/util/src/geometry.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (firstAngle <= Math.PI * 0.5 && lastAngle >= Math.PI * 0.5) point_arr.push([      0,  radius]);
+    if (firstAngle <= Math.PI * 1.0 && lastAngle >= Math.PI * 1.0) point_arr.push([-radius,       0]);
+    if (firstAngle <= Math.PI * 1.5 && lastAngle >= Math.PI * 1.5) point_arr.push([      0, -radius]);
+    if (firstAngle <= Math.PI * 2.0 && lastAngle >= Math.PI * 2.0) point_arr.push([ radius,       0]);
+    point_arr.push([x1, y1]);
+    point_arr.push([x2, y2]);
+    point_arr.push([0, 0]);
+    const x_arr = point_arr.map(n => n[0]);
+    const y_arr = point_arr.map(n => n[1]);
+    const minX = Math.min.apply(undefined, x_arr);
```
GordonSmith: `const minX = Math.min(...x_arr);` would be cleaner

jbrundage:
I'm not sure what happened to it... I don't see it in my stash list... investigating now

GordonSmith:
@jbrundage bump

GordonSmith:
@jbrundage can you rebase and I will merge.

GordonSmith:
@jbrundage  bump

PR #2920, opened 2018-10-12
feat: Add change callback support for Slider.ts

jbrundage:

### DIFF
--- a/packages/form/src/Slider.ts
+++ b/packages/form/src/Slider.ts
@@ -14,6 +14,8 @@ export class Slider extends SVGWidget {
     moveMode: "both" | "left" | "right";
     moveStartPos: number;
 
+    prevValue;
+
     slider;
 
     handleLeft;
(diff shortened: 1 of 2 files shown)


packages/form/src/Slider.ts
```diff
@@ -260,6 +271,7 @@ export class Slider extends SVGWidget {
     };
 
     name: (_?: string) => string | this;
+    change: (_: Slider) => void;
```
GordonSmith: Can you change this signature to:  `change(row, col, sel)` (sel will always be true)

And double check the params match as closly as possible to a Chart click event?

jbrundage: I agree that implementation makes the most sense, but this `change` is defined in `IInput.ts` and specifies the widget as the first input argument. Can I make this change here as well? https://github.com/hpcc-systems/Visualization/blob/b6817413f1d7a7aef2ee9115c43d05b54f81fdb9/packages/api/src/IInput.ts#L52

GordonSmith: Ahh ok - yes and input change and chart click are different things...

PR #2921, opened 2018-10-12
feat: Layered.ts can give placement to layers

jbrundage:
Add DonutSummary.ts
Add samples/chart/Pie/DonutSummary.js
Added ability to center value and labels to SummaryC.ts

### DIFF
--- a/demos/dermatology/tests/layoutFactory.js
+++ b/demos/dermatology/tests/layoutFactory.js
@@ -343,7 +343,46 @@
                         ;
                     callback(retVal);
                 });
-            }
+            },
+            placement: function (callback) {
+                legacyRequire(["test/DataFactory", "src/layout/Layered", "src/form/Slider", "src/graph/Graph", "src/graph/Vertex", "src/graph/Edge"], function (DataFactory, Layered, Slider, Graph, Vertex, Edge) {
+                    var graph = new Graph();
+                    var vertices = [];
+                    var edges = [];
+
+                    var rawData = DataFactory.Graph.simple;
+                    rawData.nodes.forEach(function (node, i) {
+                        let dateStr = `${Math.floor(Math.random() * 27) + 1991}-01-01`;
+                        vertices.push(new Vertex().text(node.name + ` (${dateStr})`).faChar(node.icon));
+                        vertices[i].date = new Date(dateStr).getTime();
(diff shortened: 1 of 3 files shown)


jbrundage:
image

jbrundage:
@GordonSmith this is a mashup of widgets (like we talked about the other day)

I would say that this is nearly production ready... though I haven't done a thorough testing as I'm not sure what your take is on my approach to the layer placement in `Layered.ts`

packages/chart/src/DonutSummary.ts
```diff
@@ -0,0 +1,79 @@
+import { Pie, SummaryC } from "@hpcc-js/chart";
```
GordonSmith: Should be:
`import { Pie} from "./Pie";`

packages/chart/src/DonutSummary.ts
```diff
@@ -0,0 +1,79 @@
...
+import { HTMLWidget } from "@hpcc-js/common";
+import { Layered } from "@hpcc-js/layout";
```
GordonSmith: I don't think chart should have a dependency on Layout...

packages/chart/src/DonutSummary.ts
```diff
@@ -0,0 +1,79 @@
...
+    summaryValue(): number;
+    summaryValue(_: number): this;
+    summaryValue_exists(): boolean;
+}
+DonutSummary.prototype.publish("summaryText", "", "string", "Subtext in centered summary widget");
+DonutSummary.prototype.publish("summaryValue", null, "number", "Value of the centered summary widget (defaults to sum of pie values when empty)");
+DonutSummary.prototype.publishProxy("paletteID", "_donut", "paletteID");
+DonutSummary.prototype.publishProxy("tooltipValueFormat", "_donut", "tooltipValueFormat");
+
+// TODO: Is there a simple way to expose all of the _donut publish properties here?
```
GordonSmith: I like the concept - but adding layout / layered and a Canvas widget is very heavy approach, when you could simply add two common/Text widgets.
Also with the Text widgets you can set their sizes and they will clip nicely with "..."s when needed.

jbrundage:
@GordonSmith it looks like comms tests are causing the unit tests to fail

PR #2935, opened 2018-10-22
chore(gallery): Add option to "test" single sample easily

GordonSmith:


GordonSmith:
Yes when launched it will use all the local "compile-umd" sources, so you can use this to test your widget development.

PR #2936, opened 2018-10-22
fix(graph): Use Vertex.pos as starting position

GordonSmith:
If the user has set a pos for a Vertex, use it as its starting position.

### DIFF
--- a/packages/graph/src/Vertex.ts
+++ b/packages/graph/src/Vertex.ts
@@ -23,6 +23,7 @@ export class Vertex extends SVGWidget {
         this._icon = new Icon();
         this._textBox = new TextBox();
         this._annotationWidgets = {};
+        this.pos({ x: undefined, y: undefined });
     }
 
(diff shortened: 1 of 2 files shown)


packages/graph/src/Vertex.ts
```diff
@@ -23,6 +23,7 @@ export class Vertex extends SVGWidget {
         this._icon = new Icon();
         this._textBox = new TextBox();
         this._annotationWidgets = {};
+        this.pos({ x: undefined, y: undefined });
```
jbrundage: Why is this necessary?

GordonSmith: As it defaults to {x: 0, y: 0} which may be a legitimate initial position...

GordonSmith: @jbrundage bump

jbrundage:
@GordonSmith looks good

PR #2937, opened 2018-10-22
feat: Value/label centering in SummaryC.ts

jbrundage:

### DIFF
--- a/packages/chart/src/SummaryC.ts
+++ b/packages/chart/src/SummaryC.ts
@@ -24,53 +24,82 @@ export class SummaryC extends CanvasWidget {
 
     update(domNode, element) {
         super.update.apply(this, arguments);
+        const context = this;
         if (this._playIntervalIdx >= this.data().length) {
             this._playIntervalIdx = 0;
         }
-        const _size = this.size();
-        const _mult = this.mult();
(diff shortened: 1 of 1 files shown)


packages/chart/src/SummaryC.ts
```diff
@@ -162,8 +174,16 @@ export interface SummaryC {
...
+
+    centeredLabel(): boolean;
+    centeredLabel(_: boolean): this;
+    centeredValue(): boolean;
+    centeredValue(_: boolean): this;
+
 }
 
+SummaryC.prototype.publish("centeredValue", false, "boolean", "Horizontally centers the value if true");
+SummaryC.prototype.publish("centeredLabel", false, "boolean", "Horizontally centers the label if true");
```
GordonSmith: How about:
```
...publish("valueAnchor", "start", "set", "Anchor Position", ["start", "middle", "end"], { tags: ["Intermediate"] });
...publish("labelAnchor", "start", "set", "Anchor Position", ["start", "middle", "end"], { tags: ["Intermediate"] });
```
To be consistent with regular "Text" widget?

jbrundage: I should have thought of this

Issue #2933, opened 2018-10-22
Widget: classed - add matching d3 v4 call signature


GordonSmith:
Classed should also support:
("string", true/false)

PR #2945, opened 2018-10-24
fix(Marshaller): Open new DDL should auto-select first element

GordonSmith:
Fixes GH-2933

### DIFF
--- a/packages/marshaller/src/dashy.ts
+++ b/packages/marshaller/src/dashy.ts
@@ -146,6 +146,10 @@ export class Dashy extends SplitPanel {
         this._elementContainer.clear();
         this._dashboard.restore(json);
         return this._dashboard.renderPromise().then(() => {
+            const elements = this._elementContainer.elements();
+            if (elements.length) {
+                this._dashboard.activate(elements[0]);
+            }
             return this._elementContainer.refresh();
         }).then(() => {
             for (const error of this._elementContainer.validate()) {


packages/marshaller/src/dashy.ts
```diff
@@ -146,6 +146,10 @@ export class Dashy extends SplitPanel {
         this._elementContainer.clear();
         this._dashboard.restore(json);
         return this._dashboard.renderPromise().then(() => {
+            const elements = this._elementContainer.elements();
+            if (elements.length) {
+                this._dashboard.activate(elements[0]);
+            }
```
jbrundage: I'm not exactly sure what's going on here, but it doesn't seem to be related to "Enhanced classed call to match d3 v4 call signature"

GordonSmith: In d3 v4 the classed method can only be called like this:
Setting:
```
element.classed("someClassID", true);
```
Getting:
```
element.classed("someClassID");
```
In v3 it also supported:
Setting:
```
element.classed({
    "someClassID": true,
    "someOtherClassID": false
});
```
Which is the variant we supported.

This PR simply adds the ability to set / get both ways.

jbrundage: I understood that about this original PR, but it seems like you may have overwritten that PR with this commit? "fix(Marshaller): Open new DDL should auto-select first element"

GordonSmith: No idea what happened there - but the title is now corrected (and I removed the other confusing comments!)

Issue #2917, opened 2018-10-08
ddl1-to-ddl2 upgrade issue with faChar


drealeed:
A ddl1 that has "faChar":"0xf007" will produce a ddl2 that has "faChar":"".  Looks like the issue is in the faCharFix() code in upgrade.ts.

Here's an example ddl1 experiencing the issue:
[facharddl.txt](https://github.com/hpcc-systems/Visualization/files/2457276/facharddl.txt)

GordonSmith:
FWIW There is no diff between the two - but it may be better to keep the Hex version of the char.

PR #2946, opened 2018-10-24
fix(ddl): Don't convert FAChar hex values to chars

GordonSmith:
Fixes GH-2917


packages/ddl-shim/src/upgrade.ts
```diff
@@ -10,8 +10,9 @@ interface IDatasourceOutputFilter extends IDatasourceOutput {
     filter: DDL1.IFilter;
 }
 
+const UPGRADE_HEX_CHAR: boolean = false;
```
jbrundage: Is the idea here to allow `UPGRADE_HEX_CHAR` to be globally overwritten if/when hex values *should* be upgraded to char?

PR #2952, opened 2018-10-26
fix(Graph): Calling raise during dragstart breaks click event

GordonSmith:

### DIFF
--- a/packages/graph/src/Graph.ts
+++ b/packages/graph/src/Graph.ts
@@ -210,7 +210,14 @@ export class Graph extends SVGZoomWidget {
                     };
                 });
             }
-            d3Select(d.target()).raise();
+
+            //  Safe Raise - does not interfere with current click event  ---
+            const target = d.target();
+            let nextSibling = target.nextSibling;
+            while (nextSibling) {
+                target.parentNode.insertBefore(nextSibling, target);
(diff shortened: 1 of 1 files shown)


packages/graph/src/Graph.ts
```diff
@@ -210,7 +210,14 @@ export class Graph extends SVGZoomWidget {
                     };
                 });
             }
-            d3Select(d.target()).raise();
+
+            //  Safe Raise - does not interfere with current click event  ---
+            const target = d.target();
+            let nextSibling = target.nextSibling;
+            while (nextSibling) {
+                target.parentNode.insertBefore(nextSibling, target);
```
jbrundage:
```suggestion
                parentNode.insertBefore(nextSibling, target);
```

packages/graph/src/Graph.ts
```diff
@@ -210,7 +210,14 @@ export class Graph extends SVGZoomWidget {
                     };
                 });
             }
-            d3Select(d.target()).raise();
+
+            //  Safe Raise - does not interfere with current click event  ---
+            const target = d.target();
+            let nextSibling = target.nextSibling;
```
jbrundage:
```suggestion
            let nextSibling = target.nextSibling;
            const parentNode = target.parentNode;
```

PR #2953, opened 2018-10-26
feat(Graph): Add PP to change centroid colour

GordonSmith:
Common up selection and centroid glow colour code.

### DIFF
--- a/demos/gallery/samples/graph/Les Miserables.js
+++ b/demos/gallery/samples/graph/Les Miserables.js
@@ -4,14 +4,11 @@ import { Graph, Vertex, Edge } from "@hpcc-js/graph";
 new Graph()
     .target("target")
     .data(createData())
-    .layout("Circle")
+    .layout("ForceDirected")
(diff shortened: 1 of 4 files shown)


demos/gallery/samples/graph/Les Miserables.js
```diff
@@ -356,10 +353,10 @@ function createData() {
             { "source": 76, "target": 58, "value": 1 }
         ]
     };
-    rawData.nodes.forEach(function (node, idx) {
+    rawData.nodes.forEach(function (node) {
         vertices.push(
             new Vertex()
-                .centroid(idx === 0)
+                .centroid(node.name === "Valjean")
```
jbrundage: As a side note: maybe the `centroid` method should belong to Graph and accept a function as its argument?

GordonSmith: Maybe - centroid v selection seems like they should be more similar.

PR #2954, opened 2018-10-26
feat(Graph): Add option to clear selection on background click

GordonSmith:

### DIFF
--- a/packages/graph/src/Graph.ts
+++ b/packages/graph/src/Graph.ts
@@ -285,6 +285,12 @@ export class Graph extends SVGZoomWidget {
     enter(domNode, element) {
         super.enter(domNode, element);
 
+        this._zoomGrab.on("click.clear", () => {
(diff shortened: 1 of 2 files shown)


packages/graph/src/Graph.ts
```diff
@@ -285,6 +285,12 @@ export class Graph extends SVGZoomWidget {
     enter(domNode, element) {
         super.enter(domNode, element);
 
+        this._zoomGrab.on("click.clear", () => {
+            if (this.selectionClearOnBackgroundClick()) {
```
jbrundage: Clicking on the background seems like a clear intent to deselect. I don't think it needs to be an option to behave otherwise... unless that background click serves some other functionality?

jbrundage: Ah okay that makes sense. Should I open an issue to deselect on background click for all widgets that support selection?

PR #2976, opened 2018-11-06
feat(pie): Add evenlySpacedLabels property to Pie

### LINKED ISSUE #2956: Pie chart tick labels overlap for smaller values

jbrundage:
Fixes #2956


GordonSmith:
Actually a good starting points:
* Are you using the height of the text to calculate the total height actually needed?
* For the text processing, divide it into 4 quadrants, rather than 2 half's: 
  * For the top left position from the middle up - only re-positioning if needed
  * For the bottom left position from the middle down - only re-positioning if needed
  * Same for the right hand side

GordonSmith:
One other thought is to spot at which point they start to overlap - and then group all those into a single slice under a "other" name.
The user would then be able to expand the "other" part, by selectively hiding some of the larger pieces via the legend?

jbrundage:
Closing in favor of #3054

PR #2996, opened 2018-11-26
fix(PropertyExt): Improved changed hashing

GordonSmith:

### DIFF
--- a/packages/common/src/PropertyExt.ts
+++ b/packages/common/src/PropertyExt.ts
@@ -682,11 +682,22 @@ export class PropertyExt extends Class {
     }
 
     hashSum(ignore: string[] = [], more = {}): string {
-        ignore = [...ignore, "fields", "classed"];
+        ignore = [...ignore, "classed"];
         const props: { [key: string]: any } = more;
         this.publishedProperties(false).filter(meta => ignore.indexOf(meta.id) < 0).forEach(meta => {
             if (this[meta.id + "_exists"]()) {
-                props[meta.id] = this[meta.id]();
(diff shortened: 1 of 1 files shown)


packages/common/src/PropertyExt.ts
```diff
@@ -682,10 +682,21 @@ export class PropertyExt extends Class {
...
         this.publishedProperties(false).filter(meta => ignore.indexOf(meta.id) < 0).forEach(meta => {
             if (this[meta.id + "_exists"]()) {
+                let value = this[meta.id]();
+                switch (meta.type) {
+                    case "widget":
+                        value = value.hashSum();
+                        break;
+                    case "widgetArray":
+                    case "PropertyArray":
+                        value = hashSum(value.map(v => v.hashSum()));
```
jbrundage: It might be a good idea to rename the imported `hashSum` to `utilityHashSum` to avoid confusion?

jbrundage: `value` seems to be unused in this context? Does assigning the response from `hashSum` to it somehow affect `this[meta.id]` as a pass by reference?

GordonSmith: It is used further down the source code (just not in the changed code).

GordonSmith: Hmmm it does look iffy.

GordonSmith:
The idea is it should improve performance for the array of object "widgets" as currently it was calculating the hash on the entire "class" instance (including functions etc.).

GordonSmith:
(...fixed now uses new value)

PR #3000, opened 2018-11-26
chore(gallery): Increase test timeout

GordonSmith:


GordonSmith:
Its really only needed for the first item as far as I can tell, after that it seems fine...

PR #3021, opened 2018-12-03
chore(typescript):  Bump version

GordonSmith:


packages/util/src/stateful.ts
```diff
@@ -39,7 +39,7 @@ export class StateObject<U, I> {
     protected set<K extends keyof U>(key: K, newVal: U[K], batchMode?: boolean): IEvent;
     protected set<K extends keyof U>(keyOrNewVals: K | U, newVal?: U[K], batchMode: boolean = false): IEvent[] | IEvent | null {
         if (typeof keyOrNewVals === "string") {
-            return this.setSingle(keyOrNewVals, newVal as U[K], batchMode);
+            return this.setSingle(keyOrNewVals as any, newVal as U[K], batchMode);  //  TODO:  "as any" should not be needed (TS >= 3.1.x)
```
jbrundage: Why is "as any" needed here if the version is being bumped to `3.1.4`?

GordonSmith: The version bump caused the new syntax error - and that works around it (for now).

PR #3050, opened 2018-12-13
feat(dashy): Enable XYAxis range selection

GordonSmith:
image

image

### DIFF
--- a/packages/chart/src/XYAxis.ts
+++ b/packages/chart/src/XYAxis.ts
@@ -2,7 +2,7 @@ import { d3SelectionType, publish, publishProxy, SVGWidget, Utility } from "@hpc
 import { max as d3Max, min as d3Min } from "d3-array";
 import { brush as d3Brush, brushSelection as d3BrushSelection, brushX as d3BrushX, brushY as d3BrushY } from "d3-brush";
 import { hsl as d3Hsl } from "d3-color";
-import { event as d3Event } from "d3-selection";
+import { event as d3Event, select as d3Select } from "d3-selection";
 import { Axis } from "./Axis";
 
 import "../src/XYAxis.css";
@@ -194,7 +194,7 @@ export class XYAxis extends SVGWidget {
         const currSel: any = d3BrushSelection(this.svgBrush.node());
         if (currSel) {
             selected = this.data().filter(function (d) {
-                const pos = context.dataPos(d[0]) + context.domainAxis.bandwidth() / 2;
+                const pos = context.dataPos(d[0]);
                 if (context.use2dSelection()) {
                     const pos2 = context.valuePos(d[1]) + context.valueAxis.bandwidth() / 2;
                     return pos >= currSel[0][0] && pos <= currSel[1][0] && pos2 >= currSel[0][1] && pos2 <= currSel[1][1];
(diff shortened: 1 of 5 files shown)


GordonSmith:
@jbrundage please review

packages/chart/src/XYAxis.ts
```diff
@@ -194,7 +194,7 @@ export class XYAxis extends SVGWidget {
         const currSel: any = d3BrushSelection(this.svgBrush.node());
         if (currSel) {
             selected = this.data().filter(function (d) {
-                const pos = context.dataPos(d[0]) + context.domainAxis.bandwidth() / 2;
```
jbrundage: Why was ` + context.domainAxis.bandwidth() / 2` removed? Should the same thing happen to ` + context.valueAxis.bandwidth() / 2` on line 199?

GordonSmith: It was not needed (once I started highlighting which ones were selected it was pretty obvious).  and yes probably...

GordonSmith: Which may mean that when the X-Axis is linear it will be needed

GordonSmith: Just tested seems to work as expected - the 2D selection is still a bit wonky though (the handles don't move with it and there are no handles on the vertical...

packages/marshaller/src/ddl2/activities/filter.ts
```diff
@@ -100,16 +100,27 @@ export class ColumnMapping extends PropertyExt {
...
-        if (isString) {
-            fs = fs.trim();
+        let fs = filterSelection.length ? filterSelection.map(sel => {
+            let retVal = sel[rf];
+            const isString = typeof retVal === "string";
+            if (isString) {
+                retVal = retVal.trim();
+            }
+            return retVal;
+        }).join(", ") : undefined;
```
jbrundage: The `.join(", ")` will cause everything to return as a string (unless it's undefined)

jbrundage: ...which will cause issues in the switch statement within `createFilter`

packages/marshaller/src/ddl2/model/element.ts
```diff
@@ -79,8 +79,13 @@ export class Element extends PropertyExt {
...
         this._elementContainer = ec;
-        vizID++;
-        this._id = `e_${vizID}`;
+        while (true) {
+            vizID++;
+            this._id = `e_${vizID}`;
+            if (!this._elementContainer.elementExists(this._id)) {
+                break;
+            }
+        }
```
jbrundage: What could cause duplicate `this._id`? It seems like this should never be necessary?

packages/marshaller/src/ddl2/activities/filter.ts
GordonSmith: Good catch - this is a potential issue:  https://github.com/hpcc-systems/Visualization/issues/3051

packages/marshaller/src/ddl2/model/element.ts
GordonSmith: Here is the scenario - create a dashboard with 4 widgets and then delete the first 2 and save it.

Now when you open it the max ID is e_4 - but there are only "2" widgets so the internal counter is now at 2

At some point there is the potential that there will be a collision when the user adds another 2 widgets...

PR #3055, opened 2018-12-17
fix(tooltip): Fixed tooltip styling issues

### LINKED ISSUE #2880: Tooltip issues
Some issues with the tooltip:
* When forced to the bottom, the arrow is not taken into consideration:
image
* The left and right arrows are a lot bigger then the top/bottom ones:
image
* Text is making it outside of the balloon:
image

jbrundage:
Fixes #2880

### DIFF
--- a/packages/api/src/ITooltip.css
+++ b/packages/api/src/ITooltip.css
@@ -15,65 +15,62 @@
 
 /* Creates a small triangle extender for the tooltip */
 .d3-tip:after {
+  content: " ";
   box-sizing: border-box;
   display: inline-block;
-  font-size: 10px;
-  width: 100%;
-  line-height: 1;
-  color: rgba(0, 0, 0, 0.66);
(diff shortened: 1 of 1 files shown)


jbrundage:
tooltips

jbrundage:
I looked at how facebook's tooltip+arrow worked and it's the same trick that's been done for many years... three transparent borders and one border the color of the arrow.

packages/api/src/ITooltip.css
```diff
@@ -15,65 +15,62 @@
 
 /* Creates a small triangle extender for the tooltip */
 .d3-tip:after {
+  content: " ";
```
GordonSmith: Is this needed (its not in the other ones?)

jbrundage: Yes, the browser won't display pseudo elements with empty content.

PR #3056, opened 2018-12-22
fix(PropertyExt): Hashsum fails with proxied properties

GordonSmith:

### DIFF
--- a/packages/common/src/PropertyExt.ts
+++ b/packages/common/src/PropertyExt.ts
@@ -296,6 +296,15 @@ export class PropertyExt extends Class {
         }
     }
 
+    resolvePublishedProxy(meta: Meta | MetaProxy) {
+        let item = this;
+        while (meta instanceof MetaProxy) {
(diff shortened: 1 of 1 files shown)


packages/common/src/PropertyExt.ts
```diff
@@ -296,6 +296,15 @@ export class PropertyExt extends Class {
         }
     }
 
+    resolvePublishedProxy(meta) {
+        let item = this;
+        while (meta.type === "proxy") {
+            item = item[(meta as any).proxy];
```
jbrundage: Isn't meta already `any`?

GordonSmith: No it has a real type (Meta) - that line should really be:
```
item = item[(meta as MetaProxy).proxy]
```
IOW its really:
```
        while (meta instanceof MetaProxy) {
            item = item[meta.proxy];
```

GordonSmith: Changes made

PR #3103, opened 2019-01-24
feat: Added series-table tooltipStyle to ITooltip

jbrundage:
Added tooltipFollowMouse to ITooltip
Added support for tooltipStyle series-table to Bubble, Column, Pie and Scatter

## Checklist:
  - [x] The commit is signed.

## Testing:
Dermatology tested in IE10 and Chrome

### DIFF
--- a/src/api/ITooltip.js
+++ b/src/api/ITooltip.js
@@ -47,7 +47,8 @@
     }
     ITooltip.prototype = Object.create(Widget.prototype);
 
-    ITooltip.prototype.publish("tooltipStyle", "default", "set", "Style", ["default", "none"], {});
+    ITooltip.prototype.publish("tooltipStyle", "default", "set", "Style", ["default", "series-table", "none"], {});
+    ITooltip.prototype.publish("tooltipFollowMouse", false, "boolean", "If true, the tooltip will follow mouse movement", null, {});
     ITooltip.prototype.publish("tooltipValueFormat", ",.2f", "string", "Value Format", null, {});
     ITooltip.prototype.publish("tooltipSeriesColor", "#EAFFFF", "html-color", "Series Color", null, {});
     ITooltip.prototype.publish("tooltipLabelColor", "#CCFFFF", "html-color", "Label Color", null, {});
@@ -60,6 +61,13 @@
         this.tooltip = d3Tip()
             .attr("class", "d3-tip")
             .offset(function (d) {
+                if(event && context.tooltipFollowMouse()){
+                    var d3tipElement = document.querySelector(".d3-tip"); // d3Tip offers no reference to the '.d3-tip' element...?
+                    d3tipElement.style.display = "block";
+                    d3tipElement.style.left = context.tooltipOffset() + (event.clientX) + "px";
+                    d3tipElement.style.top = event.clientY + "px";
+                    return [];
+                }
                 switch (context.tooltip.direction()()) {
                     case "e":
                         return [0, context.tooltipOffset()];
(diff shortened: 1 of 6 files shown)


jbrundage:
series_table_style

jbrundage:
@jchambers-ln this PR allows healthcare to use `BAR` rather than `C3_BAR` they'll just need to set these two publish properties:
`tooltipStyle: "series-table"`
`tooltipFollowMouse: true`

src/api/ITooltip.js
```diff
@@ -60,6 +61,13 @@
         this.tooltip = d3Tip()
             .attr("class", "d3-tip")
             .offset(function (d) {
+                if(event && context.tooltipFollowMouse()){
+                    var d3tipElement = document.querySelector(".d3-tip"); // d3Tip offers no reference to the '.d3-tip' element...?
+                    d3tipElement.style.display = "block";
+                    d3tipElement.style.left = context.tooltipOffset() + (event.clientX) + "px";
+                    d3tipElement.style.top = event.clientY + "px";
+                    return [];
+                }
```
GordonSmith: Is it ok to use event this way (I am guessing so)?

src/api/ITooltip.js
```diff
@@ -60,6 +61,13 @@
         this.tooltip = d3Tip()
             .attr("class", "d3-tip")
             .offset(function (d) {
+                if(event && context.tooltipFollowMouse()){
+                    var d3tipElement = document.querySelector(".d3-tip"); // d3Tip offers no reference to the '.d3-tip' element...?
+                    d3tipElement.style.display = "block";
+                    d3tipElement.style.left = context.tooltipOffset() + (event.clientX) + "px";
+                    d3tipElement.style.top = event.clientY + "px";
+                    return [];
```
GordonSmith: Why set the element top + left instead of returning the offset in the array?

GordonSmith: Should you take the "direction" into consideration?

src/api/ITooltip.js
```diff
@@ -74,6 +82,13 @@
     ITooltip.prototype.tooltipUpdate = function () {
         var classed = this.tooltip.attr("class");
         classed = classed.split(" notick").join("") + (this.tooltipTick() ? "" : " notick") + (this.tooltipStyle() === "none" ? " hidden" : "");
+        classed = classed.split(" ")
+            .filter(function(_class){
+                return _class.indexOf("ITooltip-tooltipStyle-")!==0
+            })
+            .join(" ")
+            ;
+        classed += " ITooltip-tooltipStyle-"+this.tooltipStyle();
```
GordonSmith:
```
.this.tooltip.classed("ITooltip-tooltipStyle-default", this.tooltipstyle()==="default")
.this.tooltip.classed("ITooltip-tooltipStyle-series-table", this.tooltipstyle()==="series-table")
```
Would be cleaner?

jchambers-ln:
I think this would work for them, though they would need to change the widget and do a qa cycle.

jchambers-ln:
Senthil mentioned that the reason they went to the c3 bar was due to the length of the lines and labels.

src/api/ITooltip.js
jbrundage: The offset is based on `d3-tip`'s determined placement. Setting the `left` and `top` allow us to position it based on its relative parent element.

jbrundage: In order to take direction into consideration we would need to getBoundingClientRect on the resulting tooltip html. The placement happens on mouseover tick so this would be very expensive... that said... we could cache the resulting rect based on a hash of the tooltip html and only have to call getBoundingClientRect once for each series. Thoughts?

src/api/ITooltip.js
jbrundage: Unfortunately `this.tooltip` is an object returned from `d3-tip` (not an actual d3 selection). It doesn't offer the `classed` method.

src/api/ITooltip.js
jbrundage: It seems okay to me.

jbrundage:
@jchambers-ln if they went to c3 for the length of the lines and labels then could we schedule some time to discuss these differences? This is valuable information that would allow us to learn how our native d3 charts can be enhanced to offer the type of flexibility they're looking for.

PR #3114, opened 2019-02-05
feat(map): Add GMap styling to leaflet/GMap.ts

jbrundage:

## Checklist:
  - [x] The commit is signed.

## Testing:
I tested this using `map/dist/index.js` and some styles from snazzy maps. It only works on the initial render call, but it's better than not working at all.

### DIFF
--- a/packages/map/src/leaflet/GMap.ts
+++ b/packages/map/src/leaflet/GMap.ts
@@ -26,8 +26,15 @@ export class GMapLayer extends TileLayer {
     layerEnter(map: Map) {
         super.layerEnter(map);
         this.add(new (GoogleMutant as any)({
-            type: "roadmap"
+            type: "roadmap",
+            styles: this.googleMapStyles()
         }));
     }
 }
 GMapLayer.prototype._class += " map_MapBoxLayer";
+export interface GMapLayer {
+    googleMapStyles(): object;
+    googleMapStyles(_: object): this;
+}
+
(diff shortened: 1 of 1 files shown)


packages/map/src/leaflet/GMap.ts
```diff
@@ -26,8 +26,15 @@ export class GMapLayer extends TileLayer {
...
         }));
     }
 }
 GMapLayer.prototype._class += " map_MapBoxLayer";
+export interface GMapLayer {
+    googleMapStyles(): object;
+    googleMapStyles(_: object): this;
+}
+
+GMapLayer.prototype.publish("googleMapStyles", {}, "object", "Styling for map colors etc", null, { tags: ["Basic"] });
```
GordonSmith: This doesn't allow the user to change the styles on the fly in the property editor (only works before first render)?

jbrundage: Right, the change to add this feature "on-enter-only" is very minimal and provides a great deal of customizability. I don't think GoogleMutant exposes the capability of modifying the `styles` parameter post-render

GordonSmith: We could tweak (improve) the following code to force a reload of GMap:  https://github.com/hpcc-systems/Visualization/blob/master/packages/map/src/leaflet/Leaflet.ts#L146-L148

jbrundage: I'm not sure what those lines are doing. Does that conditional essentially mean "if the map needs to be reloaded"?

PR #3119, opened 2019-02-08
fix(marshaller): Debug "v1" + "v1->v2" not refreshing correctly

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/phosphor/src/TabPanel.ts
+++ b/packages/phosphor/src/TabPanel.ts
@@ -44,6 +44,19 @@ export class TabPanel extends HTMLWidget {
         return this;
     }
 
+    removeWidget(widget: SVGWidget | HTMLWidget) {
+        const wa = this.getWidgetAdapter(widget);
+        if (wa) {
+            const found = this.content.indexOf(wa);
+            if (found >= 0) {
+                this.content.splice(found, 1);
+            }
+            widget.target(null);
+            wa.dispose();
+        }
+        return this;
+    }
+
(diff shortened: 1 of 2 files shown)


packages/phosphor/src/TabPanel.ts
```diff
@@ -44,6 +44,19 @@ export class TabPanel extends HTMLWidget {
         return this;
     }
 
+    removeWidget(widget: SVGWidget | HTMLWidget) {
+        const wa = this.getWidgetAdapter(widget);
+        if (wa) {
+            const found = this.content.indexOf(wa);
+            if (found >= 0) {
+                this.content.splice(found, 1);
+            }
```
jbrundage: Seems like lines 50 to 53 should be in `util_array` as `removeFromArray(needle: any, haystack: array)`

jbrundage: Also, is `this.content` an array of `WidgetAdapters`? Because `indexOf` doesn't work well when looking for the index of objects:

Ex: `[{"name":"A"}].indexOf({"name":"A"}) === -1`

GordonSmith: Your example uses two different objects while what I am doing is akin to:
```
var obj = {"name":"A"};
[obj].indexOf(obj) === 0
```

jbrundage: Okay cool

PR #3125, opened 2019-02-13
feat(html): Add SimpleTable.ts

jbrundage:

## Checklist:
  - [x] The commit is signed.

## Testing:
I'm using it in the DataPatterns output, but evidently something semi-recently changed RE package source loading in the gallery?


packages/html/src/SimpleTable.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    enter(domNode, element) {
+        super.enter(domNode, element);
+
+        this._table = element.append("table");
+        this._tbody = this._table.append("tbody");
+    }
+    update(domNode, element) {
+        super.update(domNode, element);
+        const trSelection = this._tbody.selectAll("tr").data(this.data());
```
GordonSmith: Move to below table.style (IOW keep with trSelection.enter())

packages/html/src/SimpleTable.ts
```diff
@@ -0,0 +1,42 @@
...
+            .each(function(d) {
+                const tr = d3Select(this);
+                const tdSelection = tr.selectAll("td").data(d);
+                tdSelection.enter()
+                    .append("td")
+                    .text(_d => (_d).toString())
+                    ;
+                tdSelection.exit().remove();
+            })
+            ;
```
GordonSmith: Missing `update` and `exit` sections - won't work when data changes...

packages/html/src/SimpleTable.ts
```diff
@@ -0,0 +1,51 @@
...
+            .each(function(d) {
+                const tr = d3Select(this);
+                const tdSelection = tr.selectAll("td").data(d);
+                tdSelection.enter()
+                    .append("td")
+                    .each(function(n, i) {
+                        d3Select(this)
+                            .classed("col-" + i, true)
+                            ;
+                    })
```
GordonSmith: Does not need to be inside an "each"

packages/html/src/SimpleTable.ts
```diff
@@ -0,0 +1,51 @@
+import { HTMLWidget } from "@hpcc-js/common";
+import { select as d3Select } from "d3-selection";
+
+export class SimpleTable extends HTMLWidget {
+    protected _table;
+    protected _tbody;
+    _transformedData;
```
GordonSmith: Looks dubious - remove.

jbrundage: It's used to allow extending classes to transform `this.data()` before calling update...

GordonSmith: Where (and why)?

jbrundage: In the `BreakdownTable.ts` (it wasn't part of the PR until I amended it this morning)

GordonSmith: This is not referenced anywhere in this PR, it should be removed (if needed in dependent code, we can review it then).

GordonSmith:
@jbrundage - two comments above.

packages/html/src/SimpleTable.ts
jbrundage: I've replaced it with this method. Does that work? That will allow extending widgets to overwrite that method with their own transformation logic.
```javascript
    transformData(data) {
        return data;
    }
```

GordonSmith: Add that method in your other PR, when its needed - otherwise your just adding a "Nothing" method.

GordonSmith: As far as I can tell - even if an extending method overrode it, it has no use in this class?

packages/html/src/SimpleTable.ts
GordonSmith:
```
                tdSelection.enter()
                    .append("td")
                    .attr("class", (d, i) ->"col-" + i)
...
```
Should do the same thing.

packages/html/src/SimpleTable.ts
```diff
@@ -0,0 +1,60 @@
...
+            ;
+        const theadTrSelection = this._thead.selectAll("tr > th").data(this.columns());
+        theadTrSelection.enter()
+            .append("th")
+            .attr("class", (n, i) => `th-${i}`)
+            .merge(theadTrSelection)
+            .text(_d => (_d).toString())
+            ;
+        theadTrSelection.exit().remove();
+        const trSelection = this._tbody.selectAll("tr").data(this.transformData(this.data()));
```
jbrundage: @GordonSmith it's used here

jbrundage: Nevermind, it's removed

packages/html/src/SimpleTable.ts
jbrundage: Done, it's removed

GordonSmith: `_transformedData` is still there...

PR #3132, opened 2019-02-19
fix(marshaller): Flyout getting clipped by parent

### LINKED ISSUE #3115: form should either float outside the parent container or internally scroll but not be longer than the containing div...
While testing HCSearchFacilityByIDV5 dashboard in dsp-dev we found that when the form has too many search fields the submit button is hidden and the user was requesting if we can make the form float outside the parent container or internally scroll. When the user tabs to the submit the form the first few rows are hidden in the result table widget.

jbrundage:
Fixes #3115

## Checklist:
  - [x] The commit is signed.

## Testing:
I've tested that the `position: fixed` style resolves the issue, but I was unable to test the actual code change on a DDL-rendered dashboard. I'll work towards doing that now.

### DIFF
--- a/src/marshaller/FlyoutButton.js
+++ b/src/marshaller/FlyoutButton.js
@@ -24,21 +24,21 @@
                     font: "FontAwesome",
                 }
             ])
-            .on("click", function(ann) {
+            .on("click", function (ann) {
                 if (ann.class === "close") {
                     context._popup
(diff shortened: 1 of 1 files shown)


jbrundage:
@GordonSmith I confirmed that this does fix the issue in DSP.

PR #3143, opened 2019-02-25
feat(html): Add StyledTable.ts

jbrundage:
## Checklist:
  - [x] The commit is signed.

## Testing:
Tested in gallery

### DIFF
--- a/packages/html/src/StyledTable.ts
+++ b/packages/html/src/StyledTable.ts
@@ -1,0 +1,58 @@
+import { SimpleTable } from "./SimpleTable";
+
+export class StyledTable extends SimpleTable {
+    constructor() {
+        super();
+    }
+
+    protected applyStyleObject(selection, styleObject) {
+        Object.keys(styleObject).forEach(styleName => {
+            selection.style(styleName, styleObject[styleName]);
+        });
+    }
+
+    update(domNode, element) {
+        super.update(domNode, element);
+
+        element.selectAll("tr,th,td").attr("style", "");
+
+        this.theadColumnStyles().forEach((styleObj, i) => {
+            this.applyStyleObject(element.select(`.th-${i}`), styleObj);
+        });
+        this.tbodyColumnStyles().forEach((styleObj, i) => {
+            this.applyStyleObject(element.selectAll(`.col-${i}`), styleObj);
+        });
+        const evenRowStylesExist = Object.keys(this.evenRowStyles()).length > 0;
+        const lastRowStylesExist = Object.keys(this.lastRowStyles()).length > 0;
+        const tbodyRows = element.selectAll("tbody > tr");
+        if (evenRowStylesExist) {
+            const tbodyEvenRows = tbodyRows.select(function(d, i) { return i % 2 ? this : null; });
+            this.applyStyleObject(tbodyEvenRows, this.evenRowStyles());
+        }
+        if (lastRowStylesExist) {
+            const tbodyLastRow = tbodyRows.select(function(d, i, arr) { return i === arr.length - 1 ? this : null; });
+            this.applyStyleObject(tbodyLastRow, this.lastRowStyles());
+        }
+    }
+}
+StyledTable.prototype._class += " html_StyledTable";
+
+export interface StyledTable {
+    tbodyColumnStyles(): Array<{[styleID: string]: any}>;
+    tbodyColumnStyles(_: Array<{[styleID: string]: any}>): this;
+    tbodyColumnStyles_default(_: Array<{[styleID: string]: any}>): this;
+    theadColumnStyles(): Array<{[styleID: string]: any}>;
+    theadColumnStyles(_: Array<{[styleID: string]: any}>): this;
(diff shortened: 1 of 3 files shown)


packages/html/src/StyledTable.ts
```diff
@@ -0,0 +1,60 @@
...
+    theadColumnStyles_default(_: any): this;
+    theadColumnStyles_exists(): boolean;
+    lastRowStyles(): any;
+    lastRowStyles(_: any): this;
+    lastRowStyles_default(_: any): this;
+    lastRowStyles_exists(): boolean;
+    evenRowStyles(): any;
+    evenRowStyles(_: any): this;
+    evenRowStyles_default(_: any): this;
+    evenRowStyles_exists(): boolean;
```
GordonSmith: All styles should have better typing

packages/html/src/StyledTable.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export class StyledTable extends SimpleTable {
+    constructor() {
+        super();
+    }
+    update(domNode, element) {
+        super.update(domNode, element);
+
+        this.theadColumnStyles().forEach((styleObj, i) => {
+            const columnSelection = element.selectAll(".th-" + i);
```
GordonSmith: `select` would be better than `selectAll`?

packages/html/src/StyledTable.ts
```diff
@@ -0,0 +1,60 @@
...
+        if (this.lastRowStyles_exists()) {
+            this.applyStyleObject("tbody > tr:last-child", this.lastRowStyles());
+        }
+    }
+    applyStyleObject(selector, styleObject) {
+        const lastRowSelection = this.element().selectAll(selector);
+        Object.keys(styleObject).forEach(styleName => {
+            lastRowSelection.style(styleName, styleObject[styleName]);
+        });
+    }
```
GordonSmith: Could do a better job commoning up some of the loops?

packages/html/src/StyledTable.ts
jbrundage: How can these be typed? `any[]` and `any{}`?

jbrundage:
@GordonSmith I've commoned up the loops and changed two of them to `select`

packages/html/src/StyledTable.ts
```diff
@@ -0,0 +1,53 @@
...
+    update(domNode, element) {
+        super.update(domNode, element);
+
+        this.theadColumnStyles().forEach((styleObj, i) => {
+            this.applyStyleObject(element.select(`.th-${i}`), styleObj);
+        });
+        this.tbodyColumnStyles().forEach((styleObj, i) => {
+            this.applyStyleObject(element.selectAll(`.col-${i}`), styleObj);
+        });
+        if (this.evenRowStyles_exists()) {
```
GordonSmith: Why the exists for this case (and not the others)?

packages/html/src/StyledTable.ts
GordonSmith: `{[styke:string]: any}`?

packages/html/src/StyledTable.ts
GordonSmith: Question:  How to styles get "unset"?

packages/html/src/StyledTable.ts
```diff
@@ -0,0 +1,53 @@
...
+            this.applyStyleObject(element.selectAll("tbody > tr:nth-child(even)"), this.evenRowStyles());
+        }
+        if (this.lastRowStyles_exists()) {
+            this.applyStyleObject(element.select("tbody > tr:last-child"), this.lastRowStyles());
+        }
+    }
+    applyStyleObject(selection, styleObject) {
```
GordonSmith: Should be `protected`

packages/html/src/StyledTable.ts
```diff
@@ -0,0 +1,53 @@
...
+        super.update(domNode, element);
+
+        this.theadColumnStyles().forEach((styleObj, i) => {
+            this.applyStyleObject(element.select(`.th-${i}`), styleObj);
+        });
+        this.tbodyColumnStyles().forEach((styleObj, i) => {
+            this.applyStyleObject(element.selectAll(`.col-${i}`), styleObj);
+        });
+        if (this.evenRowStyles_exists()) {
+            this.applyStyleObject(element.selectAll("tbody > tr:nth-child(even)"), this.evenRowStyles());
```
GordonSmith: This will break when the data changes - specifically if a single row is removed at the top?

packages/html/src/StyledTable.ts
jbrundage: > Why the exists for this case (and not the others)?

The two exist conditionals are meant to avoid the expensive dom element selection, but they always evaluate to true because they're not optional. 

> Question: How to styles get "unset"?

Styles can't currently be unset. Though I'll add that feature when I amend this PR to remove the `_exist` conditions.

jbrundage:
@GordonSmith this is revised with your suggested changes

PR #3147, opened 2019-02-26
DRAFT: Add logging to all method calls

jbrundage:
@GordonSmith Take a look at this concept. This would allow us to track the full life cycle of a widget by hooking into every method call. Writing to the console is just one example of how this could be used... but imagine the start and end log statements as the timestamp ranges. The method calls could be drawn to the MiniGantt... and their ranges could be colored based on which outer method they were called from (enter/update/exit).

### DIFF
--- a/packages/common/src/Widget.ts
+++ b/packages/common/src/Widget.ts
@@ -1,3 +1,4 @@
+import { scopedLogger } from "@hpcc-js/util";
 import { BaseType as d3BaseType, select as d3Select, Selection as d3Selection } from "d3-selection";
 import "d3-transition";
 import { Field, Grid } from "./Database";
@@ -66,10 +67,13 @@ export abstract class Widget extends PropertyExt {
 
     protected _overlayElement;
 
+    protected _logger;
+
     constructor() {
         super();
         this._class = Object.getPrototypeOf(this)._class;
         this._id = this._idSeed + widgetID++;
+        this._logger = scopedLogger(this._id);
 
         this._db = new Grid();
         this._pos = { x: 0, y: 0 };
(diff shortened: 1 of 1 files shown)


packages/common/src/Widget.ts
```diff
@@ -66,10 +67,13 @@ export abstract class Widget extends PropertyExt {
 
     protected _overlayElement;
 
+    protected _logger;
+
     constructor() {
         super();
         this._class = Object.getPrototypeOf(this)._class;
         this._id = this._idSeed + widgetID++;
+        this._logger = scopedLogger(this._id);
```
GordonSmith: Interesting - I only envisioned using the scoping for "per file" not "per widget instance"

packages/common/src/Widget.ts
```diff
@@ -657,7 +661,21 @@ export abstract class Widget extends PropertyExt {
         return this;
     }
 
-    enter(_domNode: HTMLElement, _element: d3SelectionType) { }
+    enter(_domNode: HTMLElement, _element: d3SelectionType) {
+        if ((window as any).__hpcc_debug) {
+            const context = this;
```
GordonSmith: No need for context

packages/common/src/Widget.ts
```diff
@@ -657,7 +661,21 @@ export abstract class Widget extends PropertyExt {
         return this;
     }
 
-    enter(_domNode: HTMLElement, _element: d3SelectionType) { }
+    enter(_domNode: HTMLElement, _element: d3SelectionType) {
+        if ((window as any).__hpcc_debug) {
```
GordonSmith: I think this code would need its own "switch" as its going to be very very noisy. 
Also it should be done on the constructor and probably in the `Class` class?

GordonSmith: And maybe only for specific methods?

jbrundage: Hmm... maybe another global variable could be used to control which methods are logged?
For example:
```javascript
window.__hpcc_debug_methods = ["enter","update","exit","render","resize","target","columns","data"];
```
...and how they're logged:
```javascript
window.__hpcc_debug_callbacks = {
    "args": methodName=>{
        this._logger.args(methodName, arguments);
    },
    "timestamps": methodName=>{
        this._logger.timestamps(methodName, Date.getTime());
    },
    "trace": methodName=>{
        console.groupCollapsed(this._logger._id+" "+methodName);
        console.trace();
        console.groupEnd();
    }
};
```

jbrundage:
@GordonSmith what's the take-away from this? It would be very handy to be able to modify a global debug config to tell it to log specific methods (particularly when debugging issues in production)

packages/common/src/Widget.ts
```diff
@@ -657,7 +661,21 @@ export abstract class Widget extends PropertyExt {
...
+                const orig = context[methodName];
+                context[methodName] = function() {
+                    context._logger.debug(`start ${methodName}`);
+                    const ret = orig.apply(this, arguments);
+                    context._logger.debug(`end ${methodName}`);
+                    return ret;
+                };
+            });
+        }
+    }
```
GordonSmith: This code should really be in Class.ts

PR #3157, opened 2019-02-28
feat(html): Add BreakdownTable.ts

jbrundage:
## Checklist:
  - [x] The commit is signed.

## Testing:
Tested gallery sample


jbrundage:
image

packages/html/src/BreakdownTable.ts
```diff
@@ -0,0 +1,214 @@
...
+            ;
+    }
+    update(domNode, element) {
+        const rowCount = this.useCalculatedRowCount() ? this.calculateRowCount() : this.rowCount();
+        this.transformData = function() {
+            return this.breakdownData(rowCount);
+        };
+
+        this._tooltip
+            .tooltipHTML(data => {
```
GordonSmith: Does this function have to be reset every update?

packages/html/src/BreakdownTable.ts
```diff
@@ -0,0 +1,214 @@
...
+                        .direction("n")
+                        .data(context.data())
+                        .visible(true)
+                        .render()
+                        ;
+                })
+                ;
+        }
+    }
+    breakdownData(limit: number) {
```
GordonSmith: Mark as `protected` - move to above enter

packages/html/src/BreakdownTable.ts
```diff
@@ -0,0 +1,214 @@
...
+                data.push([row[0], perc + "%"]);
+            });
+        if (showOther) {
+            const otherLabel = `${this.otherLabel()} (${len - limit + 1})`;
+            const otherPercentage = "~" + (100 - percSum) + "%";
+            data.push([otherLabel, otherPercentage]);
+        }
+        return data;
+    }
+    calculateRowCount() {
```
GordonSmith: Mark as protected and move to before enter.

packages/html/src/BreakdownTable.ts
```diff
@@ -0,0 +1,214 @@
...
+        const theadRowHeight = this.columns().length > 0 ? this.thFontSize() + 5 : 0;
+        const tbodyRowHeight = this.fontSize() + 5;
+        const tbodyAvailableHeight = this.height() - theadRowHeight;
+        const rowCount = Math.floor(tbodyAvailableHeight / tbodyRowHeight);
```
GordonSmith: No need for `rowCount`

packages/html/src/BreakdownTable.ts
```diff
@@ -0,0 +1,214 @@
...
+        return rowCount;
+    }
+    tooltip;
+    tooltipHTML: (_) => string;
+    tooltipFormat: (_) => string;
```
GordonSmith: What are these three items for?

packages/html/src/BreakdownTable.ts
jbrundage: It's reset on every update because the data and/or parameters may change.

GordonSmith: But its not accessing the data at this point in time (or any of the values as far as I can tell)?

GordonSmith: (It is using RowCount, but that is the only local value, which could be moved into the function?

jbrundage: Valid, I'll move this to enter.

PR #3162, opened 2019-03-06
feat(Gauge): Add "click" event for entire Gauge

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/chart/src/Gauge.ts
+++ b/packages/chart/src/Gauge.ts
@@ -180,6 +180,11 @@ export class Gauge extends SVGWidget {
 
     enter(domNode: HTMLElement, element: d3SelectionType) {
         super.enter(domNode, element);
+
+        element.on("click", (d: Gauge) => {
+            this.click(d);
+        });
+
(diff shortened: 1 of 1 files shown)


packages/chart/src/Gauge.ts
```diff
@@ -249,6 +254,8 @@ export class Gauge extends SVGWidget {
 
     update(domNode: HTMLElement, element: d3SelectionType) {
         super.update(domNode, element);
+        element.style("cursor", this.click !== Gauge.prototype.click ? "pointer" : null);
```
jbrundage: This is an interesting idea... maybe this same concept could be applied to other widgets with optional click events. Something like: 
```javascript
showPointer() {
    return this.click && this.click !== super.click; // I'm not sure if this exact condition would work...
}
```

GordonSmith: Yea - was pleased it worked - I suspect we can worry about generalizing when we find a second place that needs it.

PR #3167, opened 2019-03-07
feat(SVGWidget): Allow SVG Widgets to Scroll

GordonSmith:
Ensure Pie does not sort the underlying data

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/layout/src/Legend.ts
+++ b/packages/layout/src/Legend.ts
@@ -251,7 +251,11 @@ export class Legend extends SVGWidget {
         this.updateDisabled(element, dataArr);
 
         const bbox = this.getBBox(true, true);
-        this._g.attr("transform", `translate(${this.width() / 2 - bbox.width / 2 + 5},${this.height() / 2 - bbox.height / 2})`);
+        this._g.attr("transform", "translate(5,8)");
+        this.pos({
+            x: this.width() / 2 - bbox.width / 2,
+            y: this.height() / 2 - bbox.height / 2
+        });
         const legendCellsBbox = this._g.select(".legendCells").node().getBBox();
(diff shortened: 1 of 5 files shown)


jbrundage:
This is fantastic! So it looks like we'll need to do a bit of custom handling on the widgets that could significantly benefit from it because I believe the `this._boundingBox` calculation within SVGWidget currently calls `getBBox` on the `svg` element (which just fits the target element's dimensions).

jbrundage:
...but the widgets that immediately come to mind are:
- layout_Legend
- graph_Graph
- tree_Indented
- tree_Dendrogram
- timeline_MiniGantt

packages/layout/src/Legend.ts
```diff
@@ -251,7 +251,11 @@ export class Legend extends SVGWidget {
         this.updateDisabled(element, dataArr);
 
         const bbox = this.getBBox(true, true);
-        this._g.attr("transform", `translate(${this.width() / 2 - bbox.width / 2 + 5},${this.height() / 2 - bbox.height / 2})`);
+        this._g.attr("transform", "translate(5,8)");
+        this.pos({
+            x: this.width() / 2 - bbox.width / 2,
+            y: this.height() / 2 - bbox.height / 2
+        });
```
jbrundage: This new scroll bar feature works on the Legend with just this code change? I would have expected that some custom logic would be needed to read the y-offset of the last ".cell" because the bbox would only reflect the dimensions of the outer svg element?

GordonSmith: The getBBox reflects this size of the inner items and is used to manually center them (as in this case) - but rather than doing the centering inside the widget, if you let the SVG element do it for you, then it also has the scope to handle the scrolling also.

PR #3185, opened 2019-03-17
feat(build): Support dynamic versioned loading

GordonSmith:
Added name/version stamping

## Checklist:
  - [x] The commit is signed.


packages/ddl-shim/src/ddl2.ts
```diff
@@ -288,9 +291,14 @@ export interface IView {
 //  DDL  ======================================================================
 export interface Schema {
     version: "0.0.22";
+    createdBy: {
+        name: string;
+        version: string;
+    };
```
GordonSmith: @drealeed (not 100% if this is a good idea), but whichever app writes the DDL should fill these details in?

packages/ddl-shim/src/ddl2.ts
```diff
@@ -288,9 +291,14 @@ export interface IView {
...
     datasources: DatasourceType[];
     dataviews: IView[];
-    properties: IWidgetProperties;
+    properties?: IProperties;
+    hipieProperties?: IProperties;
```
GordonSmith: @drealeed here is a placeholder you can put "any" info in and the marshaller will persist (and ignore).

GordonSmith:
@jbrundage please review

packages/ddl-shim/src/ddl2.ts
jbrundage: Might be a good idea to capture each package's version here as an encoded json object. Something like:
```javascript
version_meta = JSON.stringify(btoa({"common":"2.2.4","chart":"2.4.0","marshaller":"2.6.4"}))
```

packages/composite/src/__package__.ts
```diff
@@ -0,0 +1,2 @@
+export const PKG_NAME = "@hpcc-js/composite";
+export const PKG_VERSION = "2.0.21";
```
jbrundage: Can't we just parse this info from each package's `package.json` file at build time and store it globally? Or within `common_Widget`?

packages/ddl-shim/src/upgrade.ts
```diff
@@ -702,13 +703,18 @@ class DDLUpgrade {
         return retVal;
     }
 
-    writeProperties(): DDL2.IWidgetProperties {
-        return {};
+    writeProperties(): DDL2.IProperties | undefined {
+        //  TODO:  Upgrade v1.x.x dermatologies  ---
+        return undefined;
```
jbrundage: What will be the arguments for this `writeProperties`? IOW, will it take in one widget's properties and output v2 of that one widget's properties? Or will it take in a v1 marshaller serialization and output a v2 marshaller serialization?

packages/loader/src/requirejs.ts
```diff
@@ -109,41 +92,122 @@ export function cdn(url: string, min: boolean = true, additionalPaths: { [key: s
...
 
-function local(additionalPaths: { [key: string]: string }, min: boolean = false): any {
-    const config = parseScriptUrl(true);
-    window.__hpcc_topoJsonFolder = `${config.libUrl}/map/TopoJSON`;
+export function dev(additionalPaths: { [key: string]: string } = {}): any {
+    window.__hpcc_topoJsonFolder = `${hostUrl}/map/TopoJSON`;
     const thirdPartyPaths: { [key: string]: string } = {};
     for (const key in npmPackages) {
-        thirdPartyPaths[key] = `${config.node_modulesUrl}/${npmPackages[key]}`;
+        thirdPartyPaths[key] = `${hostUrl}/../node_modules/${npmPackages[key]}`;
```
jbrundage: Can this path assumption be made here? Seems like this belongs at the application layer.

packages/ddl-shim/src/ddl2.ts
GordonSmith: That info or something similar will got into the props - this createdBy will also be used by HIPIE.
FYI - Just catching the @hpcc-js/loader version will give us a definitive version for all the packages (via the loaders package.json).

packages/composite/src/__package__.ts
GordonSmith: But you might have a newer version of "map" using an older version of "common" and it would screw things up (unlikely I know).
But I do think storing the loader version here will be a good idea (the loader version implicitly defines all the modules versions as well)

packages/ddl-shim/src/upgrade.ts
GordonSmith: It will take the v1 dermatology "layout" and upgrade it into v2 format.

packages/loader/src/requirejs.ts
GordonSmith: Its only within the context of the "dev" mode, so should be fine...

packages/loader/src/requirejs.ts
```diff
@@ -206,7 +206,7 @@ export function amd(altUrl: string = "", min: boolean = true, additionalPaths: {
     const _loaderUrl = altUrl ? altUrl + "/loader" : loaderUrl;
     if (_loaderUrl.indexOf("unpkg.com/@hpcc-js/loader") > 0) {
         return unpkgPath(_loaderUrl + "/package.json", min, additionalPaths);
-    } else if (!altUrl && _loaderUrl.indexOf("file://") === 0) {
+    } else if (!altUrl && (_loaderUrl.indexOf("file://") === 0 || _loaderUrl.indexOf("http://localhost") === 0)) {
```
GordonSmith: These changes tend to no get highlighted by GitHub very well - it would be better if you suggested the change in the comments as part of the review?

GordonSmith:
Travis failed due to:  https://github.com/hpcc-systems/Visualization/pull/3190

PR #3201, opened 2019-03-24
feat(codemirror): Add markText to Editor.ts

jbrundage:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/codemirror/src/Editor.ts
+++ b/packages/codemirror/src/Editor.ts
@@ -3,8 +3,14 @@ import { HTMLWidget } from "@hpcc-js/common";
 
 import "../src/Editor.css";
 
+export interface IPosition {
+    line: number;
+    ch: number;
+}
+
 export class Editor extends HTMLWidget {
     private _codemirror: CodeMirror.EditorFromTextArea;
+    private _markedText = [];
     protected _initialText: string = "";
 
     options(): any {
@@ -30,6 +36,38 @@ export class Editor extends HTMLWidget {
         if (!arguments.length) return this._initialText;
         this._initialText = _;
         return this;
+    }
+
+    highlight(start: IPosition, end: IPosition): this {
+        if (this._codemirror) {
+            this._markedText.push(this._codemirror.markText(start, end, {className: "cm-marked-text"}));
+        }
+        return this;
+    }
+
+    removeAllHighlight(): this {
+        for (const marked of this._markedText) {
+            marked.clear();
+        }
+        this._markedText = [];
+        return this;
+    }
+
+    positionAt(i: number): IPosition {
+        return this._codemirror.posFromIndex(i);
+    }
+
+    highlightSubstring(str: string) {
+        const splitTextArr = this.text().split(str).slice(0, -1);
+        let idx = 0;
+        splitTextArr.forEach(splitText => {
+            idx += splitText.length;
+            this.highlight(
+                this.positionAt(idx),
+                this.positionAt(idx + str.length)
+            );
+            idx += str.length;
+        });
     }
 
     enter(domNode, element) {
--- a/demos/gallery/samples/codemirror/JSEditor.js
+++ b/demos/gallery/samples/codemirror/JSEditor.js
@@ -16,8 +16,14 @@ new Pie()
     ;
 `;
 
+let count = 0;
 new JSEditor()
     .target("target")
     .javascript(code)
-    .render()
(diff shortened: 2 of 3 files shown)


packages/codemirror/src/Editor.ts
```diff
@@ -32,6 +38,22 @@ export class Editor extends HTMLWidget {
         return this;
     }
 
+    markText(start: MarkTextPosition, end: MarkTextPosition): this {
+        if (this._codemirror) {
+            start.line--;
+            end.line--;
```
GordonSmith: Don't change the callers data.

packages/codemirror/src/Editor.ts
```diff
@@ -32,6 +38,22 @@ export class Editor extends HTMLWidget {
         return this;
     }
 
+    markText(start: MarkTextPosition, end: MarkTextPosition): this {
```
GordonSmith: rename to highlightText

packages/codemirror/src/Editor.ts
```diff
@@ -32,6 +38,22 @@ export class Editor extends HTMLWidget {
...
+        if (this._codemirror) {
+            start.line--;
+            end.line--;
+            this._codemirror.markText(start, end, {className: "cm-marked-text"});
+        }
+        return this;
+    }
+
+    clearMarkedText(): this {
```
GordonSmith: rename to "clearAllHighlight"

packages/codemirror/src/Editor.ts
```diff
@@ -32,6 +38,22 @@ export class Editor extends HTMLWidget {
...
+        for (const marked of this._markedText) {
+            marked.clear();
+        }
```
GordonSmith: I doubt this works....

jbrundage: It does work. I'll add an "app" to showcase it when make the requested rename changes.

GordonSmith: FWIW I don't see any code which adds new rows to this._markedText...

jbrundage: *smack forehead* you're right, it was working on my test page because I was modifying the `markText` method on Editor's prototype

packages/codemirror/src/Editor.ts
jbrundage: It would be unnatural to pass in `{line: 1, ch: 4}, {line: 1, ch: 5}` resulting in line 2 character 4 being highlighted. This is why I subtract one from the line input.

jbrundage:
@GordonSmith done

packages/codemirror/src/Editor.ts
```diff
@@ -3,8 +3,14 @@ import { HTMLWidget } from "@hpcc-js/common";
 
 import "../src/Editor.css";
 
+export interface MarkTextPosition {
```
GordonSmith: Rename to IPosition (as it will get used in other places I suspect)

packages/codemirror/src/Editor.ts
```diff
@@ -32,6 +38,30 @@ export class Editor extends HTMLWidget {
...
+    }
+
+    positionAt(i: number): IPosition {
+        const substr = this.text().slice(0, i);
+        const lineSplit = substr.split("\n");
+        return {
+            line: lineSplit.length - 1,
+            ch: lineSplit.pop().length
+        };
+    }
```
jbrundage: @GordonSmith added

GordonSmith: I was thinking more:
```
return this._codemirror.posFromIndex(i);
```

jbrundage: Well that's handy 👍

jbrundage:
@GordonSmith I think I see where you were going with the `positionAt` idea. I added `highlightSubstrings(str: string)`

packages/codemirror/src/Editor.ts
```diff
@@ -32,6 +38,38 @@ export class Editor extends HTMLWidget {
...
+        this._markedText = [];
+        return this;
+    }
+
+    positionAt(i: number): IPosition {
+        return this._codemirror.posFromIndex(i);
+    }
+
+    highlightSubstring(str: string) {
+        const splitTextArr = this.text().split(str).slice(0, -1);
```
GordonSmith: Change to a loop of indexOf for efficiency

PR #3213, opened 2019-03-29
feat(xyaxis): Add xAxisTitle support to XYAxis.ts

### LINKED ISSUE #3191: Expose xAxisTitle similarly to yAxisTitle
Would it be possible to expose the xAxisTitle so a consumer can override as needed similarly to how the yAxisTitle is exposed via a property of the chart?  In some cases, we may want a different value than the first value of our 'columns' in our data, or we may wish to not display it even though we have a value in the 'columns' in our data.

jbrundage:
Fixes #3191

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/chart/src/Axis.ts
+++ b/packages/chart/src/Axis.ts
@@ -587,53 +587,50 @@ export class Axis extends SVGWidget {
             .attr("class", "axisTitle")
             .merge(svgText)
             ;
+        const svgTitleTransition = svgTextUpdate.transition()
+            .attr("dx", null)
+            .style("text-anchor", "end")
+            ;
         switch (this.orientation()) {
             case "left":
-                svgTextUpdate.transition()
+                svgTitleTransition
                     .attr("transform", "rotate(-90)")
                     .attr("x", -2)
                     .attr("y", 2)
-                    .attr("dx", null)
                     .attr("dy", ".71em")
-                    .style("text-anchor", "end")
-                    .text(this.title())
                     ;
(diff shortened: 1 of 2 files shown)


packages/chart/src/XYAxis.ts
```diff
@@ -333,7 +333,7 @@ export class XYAxis extends SVGWidget {
 
         this.domainAxis
             .orientation(isHorizontal ? "bottom" : "left")
-            .title(this.columns()[0])
+            .title(this.xAxisTitle() ? this.xAxisTitle() : this.columns()[0])
```
GordonSmith: Should be:  `this.xAxisTitle_exists() ? ...`

packages/chart/src/Axis.ts
```diff
@@ -587,53 +587,50 @@ export class Axis extends SVGWidget {
...
-                    .attr("dx", null)
                     .attr("dy", null)
-                    .style("text-anchor", "end")
-                    .text(this.title())
                     ;
                 break;
             default:
         }
+        svgTitleTransition
+            .text(this.title_exists() ? this.title() : "")
```
GordonSmith: `this.title()` is fine here as it will either be "" or what the user has set.
(the same is not true in XYAxis)

jbrundage:
@GordonSmith the default `title` value in Axis.ts has to change from `""` to `null` to allow XYAxis to use `xAxisTitle_exists`

PR #3220, opened 2019-04-03
feat: template can use parens to removeHTMLFromString

jbrundage:

## Checklist:
  - [x] The commit is signed.

## Testing:
Tested a few template strings directly

### DIFF
--- a/src/common/Utility.js
+++ b/src/common/Utility.js
@@ -218,12 +218,18 @@
         return Object.prototype.toString.call(obj) === "[object Array]";
     }
 
+    function removeHTMLFromString(str, div){
+        div = div ? div : document.createElement("div");
+        div.innerHTML = str;
+        return div.textContent || div.innerText || "";
+    }
+
     //  Template   ---
     //  https://github.com/Matt-Esch/string-template (MIT)
     var nargs = /\{([0-9a-zA-Z_\s\[\]]+)\}/g;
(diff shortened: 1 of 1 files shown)


jbrundage:
image

GordonSmith:
I think those regular brackets would be too common to use in this scenario?

GordonSmith:
Is the same true for `{}`?

So does that mean:  `Some str ((a))` => `Some str (123)`

Did we ask if there was ever a situation where they wanted to retain the HTML formatting?

src/common/Utility.js
```diff
@@ -218,50 +218,63 @@
...
+        div.innerHTML = str;
+        return div.textContent || div.innerText || "";
+    }
 
-        if (!args || !args.hasOwnProperty) {
-            args = {};
+//  Template   ---
+//  https://github.com/Matt-Esch/string-template (MIT)
+var nargs = /\{([0-9a-zA-Z_\s\[\]]+)\}/g;
+var nargs2 = /\(([0-9a-zA-Z_\s\[\]]+)\)/g;
```
GordonSmith: How about:  `/\{\{([0-9a-zA-Z_\s\[\]]+)\}\}/g;`
So if {a} then its verbatim, but if its:  {{a}} then it strips the HTML?

jbrundage: That sounds good to me. I'll test it out

jbrundage:
@GordonSmith I've amended the PR to add the `{{var}}` concept. It seems to be working as expected.

PR #3226, opened 2019-04-05
fix(chart): Corrected pointShapeColumn use condition

jbrundage:
## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/chart/src/Scatter.ts
+++ b/packages/chart/src/Scatter.ts
@@ -150,7 +150,7 @@ export class Scatter extends XYAxis {
                 default:
             }
         }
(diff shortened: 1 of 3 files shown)


packages/chart/src/Scatter.ts
```diff
@@ -358,6 +358,7 @@ export interface Scatter {
     pointDarken(_: boolean): this;
     interpolateDarken(): boolean;
     interpolateDarken(_: boolean): this;
+    pointSizeColumn_exists(): boolean;
```
GordonSmith: This should be in XYAxis.ts

jbrundage: Done

PR #3230, opened 2019-04-11
feat(serialization): Alternative serialization approach

GordonSmith:

## Checklist:
  - [ ] The commit is signed.


GordonSmith:
For discussion - proposed replacement for PropertyExt

tests/test-util/src/serialize.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+        expect(instance2.backColor()).to.equal("red");
+        expect(instance2.backColor.exists()).to.equal(true);
+        expect(instance2.backColor.modified()).to.equal(false);
+    });
+
+    it("functor value", function () {
+        const instance = new Derived2();
+
+        //  Note "w" is typed correctly  ---
+        instance.backColor(w => w.foreColor());
```
GordonSmith: Note:  property values can now be functions...

packages/util/src/serialize.ts
```diff
@@ -0,0 +1,139 @@
...
+                    const metaProp = to_meta(prop);
+                    if ((this as any)[prop]) {
+                        (this as any)[prop] = (this as any)[prop].bind(this, __proto__[METAProp].func);
+                    } else {
+                        (this as any)[prop] = __proto__[METAProp].func.bind(this);
+                    }
+                    (this as any)[prop].reset = __proto__[METAProp].reset.bind(this, metaProp, METAProp);
+                    (this as any)[prop].modified = __proto__[METAProp].modified.bind(this, metaProp, METAProp);
+                    (this as any)[prop].exists = __proto__[METAProp].exists.bind(this, metaProp, METAProp);
+                });
```
GordonSmith: This loop is my main objection to this code - but I believe it is still better then the current implementations (I am trying to convince myself its ok...).

jbrundage: The use of lower and upper case "META" is confusing and makes this difficult to read.

packages/core/src/serialize.ts
```diff
@@ -0,0 +1,191 @@
...
+                const retVal = {
+                    module: moduleName,
+                    class: className,
+                    properties: {} as any
+                };
+                for (const key in __proto__) {
+                    if (is_META(key)) {
+                        const property = from_META(key);
+                        if ((this as any).modified(property)) {
+                            retVal.properties[property] = (this as any)[property]();
```
jbrundage: Maybe this line should be:
```javascript
let val = (this as any)[property]();
if(val && val.serialize){
    val = val.serialize();
}
retVal.properties[property] = val;
```

packages/core/src/serialize.ts
```diff
@@ -0,0 +1,191 @@
...
+                }
+                return retVal;
+            }
+
+            deserialize(_: SerializationObject) {
+                if (moduleName !== _.module || className !== _.class) throw new Error("Module / Class does not match");
+                for (const key in __proto__) {
+                    if (is_META(key)) {
+                        const property = from_META(key);
+                        (this as any)[property](_.properties[property]);
```
jbrundage: This is goes along with my suggestion for `serialize`. This line should become something like:
```javascript
let val = _.properties[property];
if(val && val.module){
    val = new Widget().deserialize(val);
}
(this as any)[property](val);
```

PR #3271, opened 2019-05-25
feat(dgrid): Add custom sort property

### LINKED ISSUE #3192: Allow pre-sort with dgrid Table
Could we have the functionality to be able to throw data into a table and tell the table to sort (via code) on a column?  It should appear as if the user had clicked on it to sort.  Basically a pre-sort using the table's mechanics and so we get the little arrow after the sort is set and the table is rendered.

GordonSmith:
Fixes #3192

## Checklist:
  - [x] The commit is signed.


packages/dgrid/src/Common.ts
```diff
@@ -24,6 +25,19 @@ export class Common extends HTMLWidget {
     pagination: publish<this, boolean>;
     @publish(false, "boolean", "Enable sorting by column")
     sortable: publish<this, boolean>;
+    @publish("", "string", "Default sort by (Use '-' for descending sort)")
+    sortBy: publish<this, string>;
+
+    protected formatSortBy(): [{ property: string, descending: boolean }] | undefined {
+        let sortBy = this.sortBy();
+        if (!sortBy) return undefined;
+        const descending = sortBy[0] === "-";
```
jbrundage: It would be easier to work with if this was a boolean publish parameter `.descending(true)`

GordonSmith: Changes made as suggested

PR #3280, opened 2019-06-13
feat(CodeMirror): Add support for DOT Language

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/codemirror/src/DOTEditor.ts
+++ b/packages/codemirror/src/DOTEditor.ts
@@ -1,0 +1,20 @@
+import { Editor } from "./Editor";
+
+export class DOTEditor extends Editor {
+    options(): any {
+        return {
+            ...super.options(),
+            mode: "text/x-dot"
+        };
+    }
+
+    dot(): string;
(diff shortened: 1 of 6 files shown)


GordonSmith:
@jbrundage please review

packages/codemirror/src/DOTEditor.ts
```diff
@@ -0,0 +1,20 @@
...
+    options(): any {
+        return {
+            ...super.options(),
+            mode: "text/x-dot"
+        };
+    }
+
+    dot(): string;
+    dot(_: string): this;
+    dot(_?: string): string | this {
```
jbrundage: Maybe we should solely rely on `.text()` when adding support for a new language? I realize the other languages do this same sort of thing, but is it really necessary?

GordonSmith: I did have the same thought, but went with the flow.
We could add the "text" method to the base class if it doesn't exist already?

Issue #3176, opened 2019-03-12
Title Bar + Icon not displaying correctly


GordonSmith:
https://raw.githack.com/hpcc-systems/Visualization/master/demos/gallery/playground.html?./samples/common/TitleBar.js

The Icon and Title should be on the same line.

GordonSmith:
@jbrundage - this is still an issue...

PR #3282, opened 2019-06-13
feat: Add styling params to common_TitleBar

jbrundage:
Add publish proxy params for the ChartPanel TitleBar

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/TitleBar.ts
+++ b/packages/common/src/TitleBar.ts
@@ -8,6 +8,8 @@ import "../src/TitleBar.css";
 export class Button extends HTMLWidget {
     private _enabled = true;
 
+    _iconElement;
+
     constructor() {
         super();
         this._tag = "a";
@@ -16,7 +18,7 @@ export class Button extends HTMLWidget {
     enter(domNode: HTMLElement, element) {
         super.enter(domNode, element);
         const context = this;
-        element
+        this._iconElement = element
             .attr("href", "#")
             .on("click", function () {
                 context.click();
@@ -34,6 +36,14 @@ export class Button extends HTMLWidget {
         element
             .classed("disabled", !this.enabled())
             .attr("title", this.tooltip())
+            .style("display", "inline-block")
+            .style("height", this.fixedHeight() + "px")
+            .style("line-height", this.fixedHeight() + "px")
+            .style("min-width", this.fixedHeight() + "px")
+            ;
+        this._iconElement
+            .style("font-size", this.fontSizeStyle())
+            .style("line-height", this.lineHeightStyle())
             ;
     }
 
@@ -54,6 +64,11 @@ export class Button extends HTMLWidget {
         this._enabled = _;
         return this;
     }
+
+    getBBox(refresh = false, round = false) {
+        return this.calcBBox(this._element.node() ? this._element.node() : null, refresh, round);
+    }
+
 }
 Button.prototype._class += " common_Button";
 export interface Button {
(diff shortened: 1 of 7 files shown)


GordonSmith:
@jbrundage - is this related to:  https://github.com/hpcc-systems/Visualization/issues/3176?

packages/common/src/TitleBar.ts
```diff
@@ -54,16 +64,55 @@ export class Button extends HTMLWidget {
...
+            };
+        }
+        return {
+            x: (round ? Math.round(this._boundingBox.x) : this._boundingBox.x) * this._widgetScale,
+            y: (round ? Math.round(this._boundingBox.y) : this._boundingBox.y) * this._widgetScale,
+            width: (round ? Math.round(this._boundingBox.width) : this._boundingBox.width) * this._widgetScale,
+            height: (round ? Math.round(this._boundingBox.height) : this._boundingBox.height) * this._widgetScale
+        };
+    }
+
```
GordonSmith: How is this different to the default version in HTML Widget?

packages/common/src/TitleBar.ts
```diff
@@ -34,6 +36,14 @@ export class Button extends HTMLWidget {
         element
             .classed("disabled", !this.enabled())
             .attr("title", this.tooltip())
+            .style("display", "inline-block")
+            .style("height", this.fixedHeight() + "px")
+            .style("line-height", this.fixedHeight() + "px")
+            .style("min-width", this.fixedHeight() + "px")
```
GordonSmith: Is this deliberate?

jbrundage: Yes, it ensures the buttons are square (or wider if the text content needs it to be)

packages/common/src/TitleBar.ts
jbrundage: The default version in HTML Widget returns the bbox of `this._element.node().firstElementChild`. That was causing problems with TitleBar.

GordonSmith: I saw that, but was wondering why?  (probably related to title bar not wanting to 100% high like most widgets?)

That being said, you could add a new method to the HTMLWIdget:
```
calcBBox(domNode: HTMLElement, refresh:boolean = false, round:boolean = false) {
...
}
```
And have both your variants call it?

packages/common/src/TitleBar.ts
GordonSmith: Ok (just checking)

jbrundage:
> @jbrundage - is this related to: #3176?

It's not related to that issue, but I'll work on that one after the CL IE issue that just came in

packages/common/src/TitleBar.ts
jbrundage: Sounds good

packages/common/src/TitleBar.ts
```diff
@@ -8,6 +8,8 @@ import "../src/TitleBar.css";
 export class Button extends HTMLWidget {
     private _enabled = true;
 
+    _iconElement;
+
```
GordonSmith: Should be private or public and have a type

packages/common/src/TitleBar.ts
```diff
@@ -148,6 +191,18 @@ export class IconBar extends HTMLWidget {
...
+            }
+            if ((item as any).fontSizeStyle) {
+                (item as any).fontSizeStyle(this.buttonFontSizeStyle());
+            }
+            if ((item as any).lineHeightStyle) {
+                (item as any).lineHeightStyle(this.buttonLineHeightStyle());
+            }
+            if ((item as any).margin) {
+                (item as any).margin(this.buttonMargin());
+            }
```
GordonSmith: These should be inside a: 
```
else if (item instanceof Button) {
...
}
```
Then you won't need to test for each one individually...

GordonSmith:
I think I would like to take a look at the issue that this is solving.  I am not sure the title bar should be mandating some of the button sizes the way it is.

PR #3284, opened 2019-06-13
feat: Add applyLayout to Grid

jbrundage:
Add adaptiveSnapping to Grid
Add altText to Surface

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/demos/gallery/samples/layout/Grid.js
+++ b/demos/gallery/samples/layout/Grid.js
@@ -1,0 +1,21 @@
+import { Bar, Column, Step } from "@hpcc-js/chart";
+import { Grid } from "@hpcc-js/layout";
+
+const columns = ["Category", "Value-1", "Value-2"];
+const data = [
+    ["A", 34, 21],
+    ["B", 55, 34],
+    ["C", 54, 90],
+    ["D", 80, 153],
+    ["E", 86, 92],
+    ["F", 144, 233]
+];
+
+const grid = new Grid()
+    .target("target")
+    .setContent(0, 0, new Bar().columns(columns).data(data))
+    .setContent(1, 0, new Column().columns(columns).data(data))
+    .setContent(0, 1, new Step().columns(columns).data(data))
+    ;
+
+grid.render();
--- a/packages/layout/src/Grid.ts
+++ b/packages/layout/src/Grid.ts
@@ -8,6 +8,8 @@ import "../src/Grid.css";
 
 const GridList = (_GridList && _GridList.default) || _GridList;
 
+export type ICellPosition = [number, number, number, number];
+
 export class Grid extends HTMLWidget {
     divItems;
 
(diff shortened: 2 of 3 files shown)


demos/gallery/samples/layout/Grid.js
```diff
@@ -0,0 +1,24 @@
...
+const grid = new Grid()
+    .target("target")
+    .render()
+    ;
+
+grid
+    .setContent(0, 0, new Bar().columns(columns).data(data))
+    .setContent(1, 0, new Column().columns(columns).data(data))
+    .setContent(0, 1, new Step().columns(columns).data(data))
+["render"]()
```
GordonSmith: Dubious looking code - needs to be fixed at the "setContent" typing.

packages/layout/src/Grid.ts
```diff
@@ -497,11 +504,32 @@ export class Grid extends HTMLWidget {
...
 
+    applyLayout(layoutArr: ICellPosition[]) {
+        this.divItems.each((d, i) => {
+            if (layoutArr[i]) {
+                const [x, y, w, h] = layoutArr[i];
+                d
+                    .gridCol(x)
+                    .gridRow(y)
+                    .gridColSpan(w)
+                    .gridRowSpan(h)
```
GordonSmith: w + h can be undefined as per your typing above?

demos/gallery/samples/layout/Grid.js
jbrundage: I added that to avoid the playground logic that removes some method calls (in `playground.ts -> syncJavaScript`)

packages/layout/src/Grid.ts
jbrundage: Hmm... I guess they shouldn't be optional

demos/gallery/samples/layout/Grid.js
```diff
@@ -0,0 +1,21 @@
...
+];
+
+const grid = new Grid()
+    .target("target")
+    .setContent(0, 0, new Bar().columns(columns).data(data))
+    .setContent(1, 0, new Column().columns(columns).data(data))
+    .setContent(0, 1, new Step().columns(columns).data(data))
+    ;
+
+grid.render();
```
GordonSmith: Why does the following not work:
```
const grid = new Grid()
    .target("target")
    .setContent(0, 0, new Bar().columns(columns).data(data))
    .setContent(1, 0, new Column().columns(columns).data(data))
    .setContent(0, 1, new Step().columns(columns).data(data))
    .render()
    ;
```

packages/layout/src/Grid.ts
```diff
@@ -488,7 +490,12 @@ export class Grid extends HTMLWidget {
                     this.postSelectionChange();
                 }
             } else {
-                this._selectionBag.set([selectionObj]);
+                const selected = this._selectionBag.get();
+                if (selected[0] && selected[0]._id === selectionObj._id) {
```
GordonSmith: Don't you need to check all the currently selected items to see if the click is already selected?

jbrundage: No, because this selection clearing feature only applies if a single element is selected. If two cells were selected and a user clicks on one of them, the expected behavior would be to select the one they clicked rather than deselecting both of them. Though now I'm realizing that first condition should be `selected.length === 1` rather than `selected[0]`

demos/gallery/samples/layout/Grid.js
jbrundage: It does work initially on page load, but then `playground.ts -> syncJavaScript` removes the lines between `.target` and `.render` if they contain a method call that isn't a publish parameter.

GordonSmith: You need to move the setContent to above the `target` call as they are not PP and that should fix also.

PR #3325, opened 2019-07-02
feat(v1): Add filter by Legend click

### LINKED ISSUE #3324: Need ability to filter PIE chart by clicking on Legend labels (v1)

jbrundage:
Fixes #3324

## Checklist:
  - [x] The commit is signed.

## Testing:
I've tested this in the dermatology editor and there are two edge case issues that can be experienced, but they may not be worth addressing in this pull request as this is an opt-in feature that will be primarily used on `PIE`. 
First issue: 
1) enable the `hideRowOnLegendClick` feature
2) hide a row
3) switch chart types
4) the previously hidden row is now missing from the dataset
Second issue:
1) change `chartType` to an ND chart
2) enable the `hideRowOnLegendClick` feature
3) click a legend label
4) a "row" is now missing (corresponding to the clicked legend index)... legend labels are based on "column"s for ND charts

### DIFF
--- a/src/other/Legend.js
+++ b/src/other/Legend.js
@@ -9,6 +9,8 @@
     function Legend() {
         Table.call(this);
         
+        this._resetCache = true;
+
         this.showHeader(false);
         this.pagination(false);
     }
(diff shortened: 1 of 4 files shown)


GordonSmith:
@jbrundage travis failing

src/other/Legend.js
```diff
@@ -29,16 +29,25 @@
...
         this._targetWidget = _;
+        var oldDataMethod = _.data;
+        var context = this;
+        _.data = function(){
+            if(arguments.length > 0 && !context._dataFilterFlag){
+                context._hiddenRows = [];
+                context._cachedData = context._targetWidget.data();
+            } else {
+                context._dataFilterFlag = false;
+            }
```
GordonSmith: This looks like the _cachedData won't get updated when the data changes?

jbrundage:
The cached data gets updated when `this._targetWidget.data` is called (unless the dataFilterFlag is true). If the dataFilterFlag is true then that means the Legend is applying a filter to the target widget.

GordonSmith:
So if the data() gets called while a filter is applied, it won't update?

jbrundage:
I guess w

> So if the data() gets called while a filter is applied, it won't update?

That's right, but it sets the flag to true and then immediately calls that function and unsets it... it was to prevent an infinite loop at the time of writing it but now I see that it's not necessary anymore. I'll change it and retest.

GordonSmith:
Failing tests...

PR #3326, opened 2019-07-03
feat(map-deck): Add showBuildings to Common.ts

jbrundage:

## Checklist:
  - [x] The commit is signed.

## Testing:
This is an "at-runtime" feature. It seems that mapgl doesn't support the ability to add/remove layers after initialization?

### DIFF
--- a/packages/map-deck/src/Common.ts
+++ b/packages/map-deck/src/Common.ts
@@ -23,6 +23,8 @@ export class Common extends HTMLWidget {
     protected _canvasDeck: any;
     protected _layers: any[] = [];
 
+    private _prevShowBuildings: boolean;
+
     constructor() {
         super();
     }
@@ -61,6 +63,7 @@ export class Common extends HTMLWidget {
 
     enter(domNode, element) {
         super.enter(domNode, element);
+        const context = this;
 
         this._div = element.append("div")
             .attr("id", "container")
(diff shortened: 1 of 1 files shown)


packages/map-deck/src/Common.ts
```diff
@@ -110,6 +111,41 @@ export class Common extends HTMLWidget {
...
+                        ],
+                        "fill-extrusion-base": [
+                            "interpolate", ["linear"], ["zoom"],
+                            15, 0,
+                            15.05, ["get", "min_height"]
+                        ],
+                        "fill-extrusion-opacity": .6
+                    }
+                }, labelLayerId);
+            }
```
GordonSmith: Code won't work if user disables / enables at runtime?

GordonSmith:
@jbrundage bump

packages/map-deck/src/Common.ts
jbrundage: Right, once the widget has rendered, this feature can't be toggled. Should we avoid features like this? Or recreate the widget when this feature is toggled?

GordonSmith: Recreate should be fine, its really only a design or user interaction type of thing.

jbrundage:
@GordonSmith it now supports toggling without having to re-render.

jbrundage:
5340C696-82A1-418B-92ED-C85E2450E31F

PR #3358, opened 2019-08-09
fix(v1): normalize undefined or null sort values

### LINKED ISSUE #3338: Table sorting
If a table has a sparsely populated column with single letter values, empty strings, and undefined values the sorting does not work correctly when paging is involved.  At first I thought it was sorting maybe blank, undefined, values but the values are randomly sorted.

jbrundage:
Fixes #3338

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/src/common/Utility.js
+++ b/src/common/Utility.js
@@ -280,6 +280,8 @@
                         var lVal = l[sortBy[i].idx];
                         var rVal = r[sortBy[i].idx];
                         if (lVal !== rVal) {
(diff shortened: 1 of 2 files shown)


jbrundage:
I think it's related to some ES6 syntax I used in the otherFactory... forgetting that v1 doesn't like ES6... *investigating now*

jbrundage:
@GordonSmith it's passing now

PR #3367, opened 2019-08-15
feat: Add example of ddl1 with modalIfData

jbrundage:
ddl1.html looks like this and the button at the bottom toggles the `w` widget between modal/non-modal:
image

And the `ddl1_modalIfData` dermatology entry displays the same databomb+ddl1 as seen in `ddl1.html`


GordonSmith:
Closing in favor of https://github.com/hpcc-systems/Visualization/pull/3369 (thx for the sample)

PR #3389, opened 2019-09-04
feat(docs): Add Pie md files

jbrundage:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/chart/docs/HalfPie.md
+++ b/packages/chart/docs/HalfPie.md
@@ -1,0 +1,48 @@
+# HalfPie
+
+```meta
+{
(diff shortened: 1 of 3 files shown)


GordonSmith:
I think we need more "documentation" in these files - or are these just placeholders?

jbrundage:
The Pie is pretty much self explanatory as it doesn't have any real transformative publish properties. I think the two samples in each of these Pie md files are enough to convey their capabilities. Maybe we should think about adding a `ChartPanel` example that's added to each of the appropriate widgets' md? (And maybe do the same with tooltips?) Someone might wonder how to enable the Legend and/or tooltips when they read these.

GordonSmith:
@jbrundage can this be closed?

PR #3459, opened 2019-12-19
chore(FAChar): Refactor faChar logic

GordonSmith:

## Checklist:
  - [x] The commit is signed.


packages/common/src/Utility.ts
```diff
@@ -268,9 +268,20 @@ SimpleSelectionMixin.prototype.deserializeState = function (state) {
...
+export function faKeys(): string[] {
+    return Object.keys(faCharMap);
+}
+
+export function faChar(key: string, defChar: string = "fa-question"): string {
+    if (key.indexOf("fa-") === 0) {
+        return faCharMap[key] || faCharMap[defChar];
+    }
+    return key;
```
jbrundage: Maybe throw an error if `key` doesn't start with `"fa-"`?

GordonSmith: No as it suports copy / paste of a single char from the cheat sheet:  `` type of thing.

PR #3460, opened 2019-12-19
chore(textSize): Refactor textSize out of Widget

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/Utility.ts
+++ b/packages/common/src/Utility.ts
@@ -603,3 +603,30 @@ export function parseVersionString(versionString) {
         patch: parseInt(_sp[2].split("-")[0])
     };
 }
+
+type TextSize = { width: number; height: number; };
+let g_fontSizeContext: CanvasRenderingContext2D;
+const g_fontSizeContextCache: { [key: string]: TextSize } = {};
+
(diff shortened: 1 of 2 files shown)


packages/common/src/Utility.ts
```diff
@@ -603,3 +603,30 @@ export function parseVersionString(versionString) {
...
+export function textSize(_text: string | string[], fontName: string = "Verdana", fontSize: number = 12, bold: boolean = false): TextSize {
+    if (!g_fontSizeContext) {
+        let fontSizeCalc = d3Select("body > #hpcc_js_font_size");
+        if (fontSizeCalc.empty()) {
+            fontSizeCalc = d3Select("body").append("canvas")
+                .attr("id", "hpcc_js_font_size")
+                ;
+        }
+        g_fontSizeContext = (fontSizeCalc.node() as HTMLCanvasElement).getContext("2d");
+    }
```
jbrundage: I know this is beyond the scope of this change, but I think this if statement would be less invasive and more performant if it was:
```javascript
    if (!g_fontSizeContext) {
        g_fontSizeContext = document.createElement("canvas").getContext("2d");
    }
```

GordonSmith: I agree it looks a little bit "to be extra extra sure" - but performance won't be an issue as its only ever created once...

PR #3462, opened 2019-12-19
chore(d3Event): Refactor d3Event to ensure global/common instance is …

GordonSmith:
…used

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -50,3 +50,6 @@ export * from "d3-selection";
 export * from "d3-time-format";
 export * from "d3-transition";
 export * from "d3-zoom";
+
+import { event as _d3Event } from "d3-selection";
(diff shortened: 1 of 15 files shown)


packages/common/src/index.ts
```diff
@@ -50,3 +50,6 @@ export * from "d3-selection";
 export * from "d3-time-format";
 export * from "d3-transition";
 export * from "d3-zoom";
+
+import { event as _d3Event } from "d3-selection";
+export const d3Event = () => _d3Event;
```
jbrundage: How does wrapping this in a function affect the usage? (other than needing to call it as a function)

GordonSmith: There is a recurring issue where there ends up being two instances of "d3.event" in different scopes and only one is valid - this ensures that "there is only one"...

PR #3466, opened 2019-12-19
feat(compare): New compare function for Object Arrays

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/util/src/array.ts
+++ b/packages/util/src/array.ts
@@ -38,20 +38,43 @@ export function find<T>(o: ReadonlyArray<T>, predicate: (value: T, index: number
     return undefined;
 }
 
-export interface IDifferences {
-    unchanged: any[];
-    removed: any[];
-    added: any[];
+export interface IDifferences<T> {
+    unchanged: T[];
+    removed: T[];
+    added: T[];
 }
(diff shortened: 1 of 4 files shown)


packages/util/src/array.ts
```diff
@@ -61,3 +61,26 @@ export function compare(before: any[], after: any[]): IDifferences {
...
+        removed: [],
+        added: []
+    };
+    const addedMap: { [key: string]: T } = {};
+    after.forEach(item => {
+        addedMap[idFunc(item)] = item;
+        retVal.added.push(item);
+    });
+    for (const row of before) {
+        const otherIdx = retVal.added.indexOf(addedMap[idFunc(row)]);
```
jbrundage: Shouldn't this be comparing the `idFunc` hashed representations of `item` instead of comparing item directly?

GordonSmith: retVal.added and addedMap are both initialized with the same objects at the start, so a straight `indexOf` should be fine.

jbrundage: Nevermind, I see the purpose of compare2 now.

PR #3468, opened 2019-12-19
feat(PReact): Bump preact version and add some widgets

GordonSmith:
## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/react/docs/render.md
+++ b/packages/react/docs/render.md
@@ -1,0 +1,39 @@
+# render.tsx
+
+Renderer contains several functions / adapters to simplify the rendering of react components within a HPCC Widget.
+
+The common example is a `SVGAdapter` which creates an SVG Element and renders the SVG React Component.
+
+```sample-code
+import { Icon, SVGAdapter } from "@hpcc-js/react";
+
+new SVGAdapter(Icon)
+    .target("target")
+    .prop("shape", "square")
+    .prop("height", 64)
+    .prop("fill", "lightgrey")
+    .prop("stroke", "black")
+    .prop("faCharFill", "black")
+    .prop("faChar", "fa-user")
+    .render()
+;
+```
+
+Alternative use:
+
+```sample-code
+import { Icon, SVGAdapter } from "@hpcc-js/react";
+
+new SVGAdapter(Icon)
+    .target("target")
+    .props({
+        shape: "square",
+        height: 64,
+        fill: "pink",
+        stroke: "black",
+        faCharFill: "black",
+        faChar: "fa-plus"
+    })
+    .render()
+    ;
+```
--- a/packages/react/src/faChar.tsx
+++ b/packages/react/src/faChar.tsx
@@ -1,0 +1,20 @@
+import { Utility } from "@hpcc-js/common";
+import * as React from "@hpcc-js/preact-shim";
+
+interface FAChar {
+    x?: number;
+    y?: number;
+    height?: number;
+    fill?: string;
+    stroke?: string;
+    faChar?: string;
+}
+
+export const FAChar: React.FunctionComponent<FAChar> = ({
+    x,
+    y = 0,
+    height = 12,
+    fill,
+    stroke,
+    faChar = ""
+}) => <text x={x} y={y - height * 3.5 / 12} fill={fill} stroke={stroke} font-family="FontAwesome" font-size={`${height}px`} style="text-anchor: middle;" >{Utility.faChar(faChar)}</text>;
--- a/packages/react/src/shape.tsx
+++ b/packages/react/src/shape.tsx
@@ -1,0 +1,60 @@
+import * as React from "@hpcc-js/preact-shim";
+
+interface Circle {
+    radius?: number;
+    fill?: string;
+    stroke?: string;
+}
+
+export const Circle: React.FunctionComponent<Circle> = ({
+    radius = 32,
+    fill = "navy",
+    stroke = fill
+}) => <circle r={radius} fill={fill} stroke={stroke} />;
+
+interface Square {
+    radius?: number;
+    cornerRadius?: number;
+    fill?: string;
+    stroke?: string;
+}
+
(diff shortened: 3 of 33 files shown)


GordonSmith:
Dependent on:  
* https://github.com/hpcc-systems/Visualization/pull/3459
* https://github.com/hpcc-systems/Visualization/pull/3460

packages/common/src/Utility.ts
```diff
@@ -268,9 +268,20 @@ SimpleSelectionMixin.prototype.deserializeState = function (state) {
...
+export function faKeys(): string[] {
+    return Object.keys(faCharMap);
+}
+
+export function faChar(key: string, defChar: string = "fa-question"): string {
+    if (key.indexOf("fa-") === 0) {
+        return faCharMap[key] || faCharMap[defChar];
+    }
+    return key;
```
jbrundage: If the key doesn't start with "fa-" maybe we should throw an error?

packages/react/docs/render.md
```diff
@@ -0,0 +1,23 @@
+# render.tsx
+
+#render
+
+<!--meta:render
+Unsupported: ":render"
+-->
+
+Bar and [Column](./Column.md) are effectively the same class, but have one different default value - their _orientation_. They support all of the same properties.
```
jbrundage: Did you mean to use the chart_Bar text here?

packages/react/src/render.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+export class SVGAdapter<P> extends SVGWidget {
+
+    @publish({}, "object", "Properties")
+    protected _props: P = {} as P;
+    props(): P;
+    props(_: Partial<P>): this;
+    props(_?: Partial<P>): P | this {
```
jbrundage: Will typescript recognize the first `prop` string argument and suggest valid values for the second?

packages/react/src/render.tsx
```diff
@@ -0,0 +1,71 @@
...
+        if (!arguments.length) return this._props;
+        this._props = { ...this._props, ..._ };
+        return this;
+    }
+
+    prop<K extends keyof P>(_: K): P[K];
+    prop<K extends keyof P>(_: K, value: P[K]): this;
+    prop<K extends keyof P>(_: K, value?: P[K]): this | P[K] {
```
jbrundage: Will typescript recognize the first prop string argument and suggest valid values for the second?

packages/react/src/shape.tsx
```diff
@@ -0,0 +1,66 @@
...
+    stroke?: string;
+}
+
+export const Circle: React.FunctionComponent<Circle> = ({
+    cx,
+    cy,
+    radius = 32,
+    fill = "navy",
+    stroke = fill
+}) => <circle cx={cx} cy={cy} r={radius} fill={fill} stroke={stroke} />;
```
jbrundage: If we position the shapes by setting their `top` and `left` bounds it should make it easier to position them relative to each other.

packages/react/src/shape.tsx
```diff
@@ -0,0 +1,66 @@
...
+    fill?: string;
+    stroke?: string;
+}
+
+export const Square: React.FunctionComponent<Square> = ({
+    radius = 30,
+    cornerRadius = 0,
+    fill = "white",
+    stroke
+}) => <rect x={-radius} y={-radius} rx={cornerRadius} ry={cornerRadius} width={radius * 2} height={radius * 2} fill={fill} stroke={stroke || fill} />;
```
jbrundage: It looks like nothing can position a `Square`?

packages/react/src/shape.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+interface Rectangle {
+    y?: number;
+    width?: number;
+    height?: number;
+    cornerRadius?: number;
+    fill?: string;
+    stroke?: string;
+}
```
jbrundage: `y` can be specified but not `x`?

packages/react/src/shape.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from "@hpcc-js/preact-shim";
+
+interface Circle {
+    cx?: string;
+    cy?: string;
```
jbrundage: Shouldn't `cx` and `cy` be `number`?

packages/common/src/Utility.ts
GordonSmith: No it supports both a single char (copy / paste from the cheat sheet) or the fa-* name.

packages/react/docs/render.md
GordonSmith: Nope

packages/react/src/render.tsx
GordonSmith: Nearly it will spot the first prop string and suggest `types` for the value (i.e. string, number, boolean etc.)

packages/react/src/render.tsx
GordonSmith: I don't understand?

packages/react/src/faChar.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const FAChar: React.FunctionComponent<FAChar> = ({
+    x,
+    y = 0,
+    height = 12,
+    fill,
+    stroke,
+    faChar = ""
+}) => <text x={x} y={y - height * 3.5 / 12} fill={fill} stroke={stroke} font-family="FontAwesome" font-size={`${height}px`} style="text-anchor: middle;" >{Utility.faChar(faChar)}</text>;
```
jbrundage: This `y` might be a bit off for some of the fa icons. Maybe we should add another utility method that precisely measures the pixel height and offset of a characters?

GordonSmith: @jbrundage textSize should probably work and we could pre-calc all their actual sizes @12px and store with the name/char map we already have?

PR #3482, opened 2020-01-17
feat(dashy): Add accessors to csv/tsv parser

jbrundage:
## Checklist:
  - [x] The commit is signed.


GordonSmith:
This is not a good idea, as its potentially coercing the data types per cell and not per column.
...plus...

Dashy already supports specifying the correct types at the Datasource level and will coerce the data correctly.

image

PR #3486, opened 2020-01-20
feat(util): Add Dijkstra algorithm to graph2

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/util/src/graph2.ts
+++ b/packages/util/src/graph2.ts
@@ -437,4 +437,103 @@ export class Graph2<V = any, E = any, S = any> {
         }
         return retVal;
     }
+
+    dijkstra(source: string, target: string): { ids: string[], len: number } {
+        const edges = this.edges();
+        const Q = new Set<string>();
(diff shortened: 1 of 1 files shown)


packages/util/src/graph2.ts
```diff
@@ -437,4 +437,103 @@ export class Graph2<V = any, E = any, S = any> {
...
+            Q.add(v2);
+
+            dist[v1] = Infinity;
+            dist[v2] = Infinity;
+
+            if (adj[v1] === undefined) adj[v1] = {};
+            if (adj[v2] === undefined) adj[v2] = {};
+
+            adj[v1][v2] = len;
+            adj[v2][v1] = len;
```
jbrundage: It might be a good idea to optionally allow `len` to be data-driven. Ex: Two people connected via a crime might be a shorter length than two people connected via a business.

GordonSmith: Agree - but will need to add "weight" support to the entire container first...

PR #3510, opened 2020-02-13
feat(graph2): Add scaleMultiplier

jbrundage:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/SVGZoomWidget.ts
+++ b/packages/common/src/SVGZoomWidget.ts
@@ -179,9 +179,17 @@ export class SVGZoomWidget extends SVGWidget {
         }
     }
 
-    zoomToFit(transitionDuration?) {
+    zoomToFit(transitionDuration?, bboxScale?: number) {
         if (this._renderElement) {
             const bbox = this._renderElement.node().getBBox();
+            if (bboxScale) {
+                const widthDiff = (bbox.width * bboxScale) - bbox.width;
+                const heightDiff = (bbox.height * bboxScale) - bbox.height;
+                bbox.width += widthDiff;
+                bbox.height += heightDiff;
+                bbox.x -= widthDiff / 2;
+                bbox.y -= heightDiff / 2;
+            }
             this.zoomToBBox(bbox, transitionDuration);
         }
     }
--- a/packages/graph/src/graph2/graph.ts
+++ b/packages/graph/src/graph2/graph.ts
@@ -252,7 +252,7 @@ export class Graph2 extends SVGZoomWidget {
             if (this.applyScaleOnLayout()) {
                 //  Wait for any transitions to finish  ---
                 setTimeout(() => {
-                    this.zoomToFit();
+                    this.zoomToFit(0, this.scaleMultiplier());
                 }, this.transitionDuration());
             }
         });
@@ -860,6 +860,8 @@ export interface Graph2 {
     highlightOnMouseOverEdge(_: boolean): this;
     transitionDuration(): number;
     transitionDuration(_: number): this;
+    scaleMultiplier(): number;
+    scaleMultiplier(_: number): this;
     /*
     showEdges(): boolean;
     showEdges(_: boolean): this;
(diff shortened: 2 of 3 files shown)


jbrundage:
Allows for some padding when using `applyScaleOnLayout`
image

packages/common/src/SVGZoomWidget.ts
```diff
@@ -179,9 +179,17 @@ export class SVGZoomWidget extends SVGWidget {
         }
     }
 
-    zoomToFit(transitionDuration?) {
+    zoomToFit(transitionDuration?, bboxScale?: number) {
         if (this._renderElement) {
             const bbox = this._renderElement.node().getBBox();
+            if (bboxScale) {
+                const widthDiff = (bbox.width * bboxScale) - bbox.width;
+                const heightDiff = (bbox.height * bboxScale) - bbox.height;
```
GordonSmith: bboxScale could be undefined

jbrundage: If it's undefined then it will skip that 185 to 192 conditional.

GordonSmith: Indeed!

GordonSmith:
I am finding it hard to visualize what "scaleMultiplier" actually does?

packages/common/src/SVGZoomWidget.ts
jbrundage: This just means that zoomToFit would act normally if bboxScale is undefined.

jbrundage:
@GordonSmith scaleMultiplier just tricks the "zoomToFit" to think that the bbox (that it's trying to fit to) is larger (or smaller). Here's a before (left) and after (right):
image

jbrundage:
It just gives a bit of padding/spacing when zooming to fit.

packages/common/src/SVGZoomWidget.ts
```diff
@@ -179,9 +179,17 @@ export class SVGZoomWidget extends SVGWidget {
         }
     }
 
-    zoomToFit(transitionDuration?) {
+    zoomToFit(transitionDuration?, bboxScale?: number) {
```
GordonSmith: rename to `padding`

packages/common/src/SVGZoomWidget.ts
GordonSmith: Rather than a "scale" would be better to have an absolute padding (i.e. 16px not 16%)

packages/graph/src/graph2/graph.ts
```diff
@@ -909,6 +911,7 @@ Graph2.prototype.publish("dragSingleNeighbors", true, "boolean", "Dragging a Ver
 Graph2.prototype.publish("layout", "Circle", "set", "Default Layout", GraphLayoutTypeSet, { tags: ["Basic"] });
 Graph2.prototype.publish("scale", "100%", "set", "Zoom Level", ["all", "width", "selection", "100%", "90%", "75%", "50%", "25%", "10%"], { tags: ["Basic"] });
 Graph2.prototype.publish("applyScaleOnLayout", false, "boolean", "Shrink to fit on Layout", null, { tags: ["Basic"] });
+Graph2.prototype.publish("scaleMultiplier", 1.1, "number", "Multiplier applied to scaled layout", null, {disable: (w: any) => !w.applyScaleOnLayout() });
```
GordonSmith: rename to `zoomToFitPadding` or similar.
Move into SVGZoomWIdget

packages/common/src/SVGZoomWidget.ts
jbrundage: Scaling it with a multiplier is a quick win, but supporting pixel padding seems like a bigger change... because I think the zoom would have to be considered?

GordonSmith: Would it not be a matter of subtracting 16px from  bbox.width / height?

packages/graph/src/graph2/graph.ts
GordonSmith: @jbrundage bump

packages/common/src/SVGZoomWidget.ts
jbrundage: You're right, I'm not sure why the scaling seemed more reliable, but px can be used here.

PR #3524, opened 2020-02-20
feat(Graph2): Various improvements for ECL Watch integration

GordonSmith:

## Checklist:
  - [x] The commit is signed.


GordonSmith:
@jbrundage Please Review

packages/react/src/text.tsx
```diff
@@ -2,26 +2,74 @@ import { Utility } from "@hpcc-js/common";
...
+                height={height}
+                fontFamily={fontFamily}
+                fill={fill}
+                y={yOffset + i * (height + 2)}
+            />
+        </g>;
+    });
+
+    return <>{TextLines}</>;
+};
```
jbrundage: This seems like `Text` is trying to do too much. The ability to line wrap text isn't always needed. Seems like this functionality could be added by leaving `Text` alone and doing this:
```
export const WrappingText: React.FunctionComponent<IWrappingText> = ({
...
}) => {
    const TextLines = parts.map((p, i) => {
        return <g transform={`translate(0 ${yOffset + i * (height + 2)})`}>
            <Text
                text={p}
                height={height}
                fontFamily={fontFamily}
                fill={fill}
            />
        </g>;
    });
    return <g>{TextLines}</g>;
};
```

GordonSmith:
@jbrundage plrease re-review

packages/react/src/text.tsx
GordonSmith: Is this just a naming thing?  I have "TextLine" + "Text" - you have "Text" + "WrappingText"?

GordonSmith: I have removed "y" from Text and moved that logic into the <g> element as per your suggestion above.

PR #3525, opened 2020-02-21
feat(graph2): Abstract CategoryVertex from Vertex

jbrundage:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/graph/src/graph2/graph.ts
+++ b/packages/graph/src/graph2/graph.ts
@@ -1,5 +1,5 @@
 ﻿import { d3Event, drag as d3Drag, Palette, select as d3Select, Selection, Spacer, SVGGlowFilter, SVGZoomWidget, ToggleButton, Utility, Widget } from "@hpcc-js/common";
-import { FunctionComponent, IconEx, Icons, render, Subgraph, Vertex } from "@hpcc-js/react";
+import { CategoryVertex, FunctionComponent, IconEx, Icons, render, Subgraph, Vertex } from "@hpcc-js/react";
 import { Graph2 as GraphCollection } from "@hpcc-js/util";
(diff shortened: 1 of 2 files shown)


packages/graph/src/graph2/graph.ts
```diff
@@ -550,6 +550,15 @@ export class Graph2 extends SVGZoomWidget {
...
 
+    private _categoryRenderer: FunctionComponent<Vertex> = CategoryVertex;
+    categoryRenderer(): FunctionComponent;
+    categoryRenderer(_: FunctionComponent): this;
+    categoryRenderer(_?: FunctionComponent): this | FunctionComponent {
+        if (!arguments.length) return this._categoryRenderer;
+        this._categoryRenderer = _;
+        return this;
+    }
+
```
GordonSmith: Not needed - Graoh2 should not be changed

GordonSmith:
@jbrundage bump

GordonSmith:
@jbrundage bump

packages/graph/src/graph2/graph.ts
GordonSmith: @jbrundage bump

jbrundage:
@GordonSmith I'm trying to wrap my mind around making this change without modifying graph2. 

The way I imagine this problem being solved is by moving the decision between two rendering modes into graph2. 
`_vertexRenderer` when a node doesn't have a categoryID
`_categoryRenderer` when a node does have a categoryID

This would allow users to specify a renderer for each case. Thoughts?

GordonSmith:
I don't think a single graph instance would mix and match between category vertices and plain vertices?

Also after some usage - the Category Vertex didn't add any performance enhancement and brought general pain (selection colors for example).

jbrundage:
Interesting (@ the fact that the category vertex version didn't add performance enhancements). I'll revise this PR when I finish with the latest Bar enhancements.

PR #3528, opened 2020-02-28
chore(SelectionBag): Add typings to Selection Bag

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/Utility.ts
+++ b/packages/common/src/Utility.ts
@@ -1,6 +1,6 @@
 import { isArray } from "@hpcc-js/util";
 import { ascending as d3Ascending, descending as d3Descending } from "d3-array";
-import { select as d3Select } from "d3-selection";
+import { select as d3Select, Selection as d3SelectionT } from "d3-selection";
 import { timeFormat as d3TimeFormat } from "d3-time-format";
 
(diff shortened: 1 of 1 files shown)


packages/common/src/Utility.ts
```diff
@@ -117,14 +122,14 @@ export class SelectionBag extends SelectionBase {
         return retVal;
     }
 
-    set(itemArray) {
+    set(itemArray: ISelectionItem[]) {
         this.clear();
         itemArray.forEach(function (item) {
             this.append(item);
         }, this);
     }
```
jbrundage: Is this intention with this method that we can just pass in an array of:
```
someWidget._selection.set([
    {
        _id: "some_arbitrary_string_id",
        element: d3SelectionOfAnElementWithADataRowBoundToIt
    }
])
```
...to set the selection for that widget?

GordonSmith: Yes the logic didn't change, just the syntax checking (which makes it easier when adding to new Widgets).

GordonSmith: (I was using it incorrectly in Graph2 and changing this highlighted it straight away)

PR #3533, opened 2020-03-04
feat: Add input.tsx

jbrundage:

## Checklist:
  - [ ] The commit is signed.

### DIFF
--- a/packages/react/src/input.tsx
+++ b/packages/react/src/input.tsx
@@ -1,0 +1,252 @@
+import { IPublishExt, PublishTypes } from "@hpcc-js/common";
+import * as React from "@hpcc-js/preact-shim";
+
+export interface IInput {
+    className?: string;
+    callbacks?: any;
+    property: IPublishedProperty;
+}
+
+export const RangeInput: React.FunctionComponent<IInput> = ({
+    callbacks,
+    className,
+    property
+}) => {
+    return <input
+        {...callbacks}
+        type="range"
+        min={property.ext.min}
+        max={property.ext.max}
+        step={property.ext.step}
+        value={property.value}
+        className={className}
+    />;
+};
+export const NumberInput: React.FunctionComponent<IInput> = ({
+    callbacks,
+    className,
+    property
+}) => {
+    return <input
+        {...callbacks}
+        type="number"
+        min={property.ext.min}
+        max={property.ext.max}
+        step={property.ext.step}
+        value={property.value}
+        className={className}
+    />;
+};
+export const BooleanInput: React.FunctionComponent<IInput> = ({
+    callbacks,
+    className,
+    property
+}) => {
+    return <input
+        {...callbacks}
+        type="checkbox"
+        checked={property.value ? "checked" : false}
+        className={className}
+    />;
+};
+export const StringInput: React.FunctionComponent<IInput> = ({
+    callbacks,
+    className,
+    property
+}) => {
+    return <input
+        {...callbacks}
+        type="text"
+        value={property.value}
+        className={className}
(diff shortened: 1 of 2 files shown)


packages/react/src/input.tsx
```diff
@@ -0,0 +1,96 @@
...
+export const Input: React.FunctionComponent<IInput> = ({
+    type = "text",
+    className,
+    value,
+
+    attributes = {}
+}) => {
+    return <input
+        value={value}
+        {...attributes}
```
GordonSmith: attributes should be specific and not general (will help re-use in the long run)

packages/react/src/input.tsx
```diff
@@ -0,0 +1,96 @@
...
+    labelMinWidth = 30,
+    labelMaxWidth = 120,
+    flexDirection = "row",
+    input
+}) => {
+    const i18n = label => label;
+    const inputs = [];
+    if (input.type === "number" || input.type === "range") {
+        inputs.push(<Input {...input}/>);
+        inputs.push(<Input {...input} type="range"/>);
```
GordonSmith: Why are there two inputs in this scenario?

packages/react/src/input.tsx
```diff
@@ -0,0 +1,96 @@
...
+    inputs.forEach(input => {
+        items.push(
+            <LabelInput
+                {...input}
+            />
+        );
+    });
+    const divStyle = "display:flex;flex-direction:column;";
+    return <div style={divStyle}>
+        {items}
```
GordonSmith: When there are lists of items like this, they should have "keys" to improve performance.

packages/react/src/input.tsx
jbrundage: Okay, I was on the fence about this too... more vs less explicit. What are the right attributes for input here?

At first I had something like this:
```
const Input = {
    // attributes: value, placeholder, className
    //    onkeyup, onclick, onfocus, onblur, oninput
}
const NumericInput = {
    // type="number" | "range"
    // attributes: min, max, step
}
const StringInput = {
    // type="text"
    // no unique attributes come to mind
}
const BooleanInput = {
    // type="checkbox"
    // attributes: checked
}
```

packages/react/src/input.tsx
jbrundage: It's a slider with a number input next to it. It's super handy to be able to use a slider to test some values OR set a specific value with the number input.

jbrundage: ...now that I'm looking at it again it *could* be a range with a range next to it if `type="range"` was passed in.

GordonSmith: For range yes - but for just numeric input, I would have thought a "spinner" would be the norm?

packages/react/src/input.tsx
jbrundage: @GordonSmith does the above look good?

packages/react/src/input.tsx
jbrundage: It's super handy to have an editable number input alongside the slider... best of both worlds... sometimes it can be frustrating using a slider and trying to set a specific value.

packages/react/src/input.tsx
GordonSmith: I don't think you want to have NumericInput with types of number + range.  Rather you would have NumericInput and RangeInput?

packages/react/src/input.tsx
GordonSmith: Agree with usability - but that should be a SliderInput.

Also I would make Label an attribute of Input and not the other way round?

PR #3544, opened 2020-03-26
feat(observable-md): Add new Observable Markdown Widget

GordonSmith:

## Checklist:
  - [x] The commit is signed.


packages/observable-md/index.html
```diff
@@ -0,0 +1,163 @@
+<!doctype html>
```
jbrundage: I like the idea of having an `index.html` in a package directory, but this is not standard for the project. Should we aim to have an `index.html` in each package showcasing some simple usage?

packages/loader/src/meta.ts
```diff
@@ -42,13 +42,19 @@ export const npmPackages = {
     "dagre": "dagre/dist/dagre",
     "@deck.gl/core": "@deck.gl/core/dist",
     "@deck.gl/layers": "@deck.gl/layers/dist",
+    "@observablehq/parser": "@observablehq/parser/dist/parser.min",
+    "@observablehq/runtime": "@observablehq/runtime/dist/runtime.umd",
+    "@observablehq/inspector/dist/inspector.css": "@observablehq/inspector/dist/inspector.css",
+    "acorn": "acorn/dist/acorn",
+    "acorn-walk": "acorn-walk/dist/walk",
```
jbrundage: What will we use `acorn` for?

GordonSmith: We don't but its pulled in by observable...

packages/observable-md/index.html
GordonSmith: I see an index.html as being part of docs, so the more the better...

PR #3553, opened 2020-04-03
fix(TitleBar):  Select had incorrect "selected" value.

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/TitleBar.ts
+++ b/packages/common/src/TitleBar.ts
@@ -141,6 +141,13 @@ export class SelectDropDown extends HTMLWidget {
         super.enter(domNode, element);
         const context = this;
         element.on("change", function (d) {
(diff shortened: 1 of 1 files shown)


GordonSmith:
Your change will fail in IE11

jbrundage:
Shouldn't IE11 use the ES5 build?

GordonSmith:
It does - but I didn't think typescript will transpile that?

GordonSmith:
Property 'entries' does not exist on type 'ObjectConstructor'.

jbrundage:
Ah okay, after a bit of searching it looks like folks are importing polyfills from `core-js` to add `Object.entries` support. It's probably not worth it.

PR #3566, opened 2020-04-14
feat: lazyRender returns this

### LINKED ISSUE #3551: lazyRender should return this to preserve chaining

jbrundage:
Fixes #3551

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/Widget.ts
+++ b/packages/common/src/Widget.ts
@@ -652,9 +652,14 @@ export abstract class Widget extends PropertyExt {
         });
     }
 
-    lazyRender = debounce(function (debouncedCallback?: (w: Widget) => void) {
+    private _lazyRender = debounce(function (debouncedCallback?: (w: Widget) => void) {
         this.render(debouncedCallback);
     }, 100);
+
+    lazyRender(): this {
+        this._lazyRender();
+        return this;
+    }
 
     animationFrameRender(): this {
         if (requestAnimationFrame) {


packages/common/src/Widget.ts
```diff
@@ -652,9 +652,12 @@ export abstract class Widget extends PropertyExt {
         });
     }
 
-    lazyRender = debounce(function (debouncedCallback?: (w: Widget) => void) {
-        this.render(debouncedCallback);
-    }, 100);
+    lazyRender(): this {
+        debounce(function (debouncedCallback?: (w: Widget) => void) {
+            this.render(debouncedCallback);
+        }, 100);
```
GordonSmith, line 658: This looks like if lazyRender is called multiple times within 100ms that render will also get called multiple times?

jbrundage: Hmmmm, good point. How can this be avoided? Cache the callback function and compare against subsequent calls?

jbrundage: ...or maybe return `this` from within `debounce`? But `debounce` returns a `debounced` function. I'll give this issue another look from the perspective of rapidly calling it.

GordonSmith: rename to `private _lazyRender` and then have a new lazyRender call it and return this.

GordonSmith: @jbrundage bump

packages/common/src/Widget.ts
```diff
@@ -652,10 +652,15 @@ export abstract class Widget extends PropertyExt {
...
 
-    lazyRender = debounce(function (debouncedCallback?: (w: Widget) => void) {
+    private _lazyRender = debounce(function (debouncedCallback?: (w: Widget) => void) {
         this.render(debouncedCallback);
     }, 100);
 
+    lazyRender(): this {
+        this._lazyRender();
+        return this;
+    }
```
jbrundage, line 662: @GordonSmith is this what you had in mind?

GordonSmith: Yup.

PR #3589, opened 2020-05-07
feat: Add readOnly mode to Editor.ts

jbrundage:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/demos/gallery/src-umd/config.js
+++ b/demos/gallery/src-umd/config.js
@@ -725,6 +725,16 @@ var config = {
             }
           },
           {
+            "path": "./samples/codemirror/ReadOnly.js",
+            "name": "ReadOnly.js",
+            "type": "file",
+            "imports": {
+              "@hpcc-js/codemirror": [
+                "JSEditor"
+              ]
+            }
+          },
+          {
             "path": "./samples/codemirror/XMLEditor.js",
             "name": "XMLEditor.js",
             "type": "file",
--- a/packages/codemirror/src/Editor.ts
+++ b/packages/codemirror/src/Editor.ts
@@ -15,6 +15,7 @@ export class Editor extends HTMLWidget {
 
     options(): any {
         return {
+            readOnly: this.readOnly(),
             lineNumbers: true,
             tabSize: 2
         };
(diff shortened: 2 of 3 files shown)


demos/gallery/src-umd/config.js
```diff
@@ -724,6 +724,16 @@ var config = {
...
+          {
+            "path": "./samples/codemirror/ReadOnly.js",
+            "name": "ReadOnly.js",
+            "type": "file",
+            "imports": {
+              "@hpcc-js/codemirror": [
+                "JSEditor"
+              ]
+            }
+          },
```
GordonSmith, line 736: This file is missing?

jbrundage: I've added it.

packages/codemirror/src/Editor.ts
```diff
@@ -9,12 +9,13 @@ export interface IPosition {
 }
 
 export class Editor extends HTMLWidget {
-    private _codemirror: CodeMirror.EditorFromTextArea;
+    protected _codemirror: CodeMirror.EditorFromTextArea;
```
GordonSmith, line 12: This change (which is not needed) is causing the build failure.

jbrundage: I noticed that too, had it changed to "protected" to make debugging easier. I've rolled it back and amended.

packages/codemirror/src/Editor.ts
```diff
@@ -15,6 +15,7 @@ export class Editor extends HTMLWidget {
 
     options(): any {
         return {
+            readOnly: this.readOnly(),
```
GordonSmith, line 18: This should not be needed?

jbrundage: That allows the option to be set at the point of instantiation.

GordonSmith: Your code in the update method also gets called on instantiation?

jbrundage: The codemirror instantiation takes place in the enter method: `CodeMirror.fromTextArea`

Then the update method sets the option on the already created instance of codemirror. That seems to work well enough with this readOnly option. It just seems like a good idea to allow codemirror to set the option before *their* instantiation.

PR #3614, opened 2020-05-21
feat(DataGraph): Add dynamic mapping

GordonSmith:
Add better support for graph wasm loading

## Checklist:
  - [x] The commit is signed.


GordonSmith:
@jbrundage please review

packages/graph/src/graph2/dataGraph.ts
```diff
@@ -87,6 +88,13 @@ export class DataGraph extends Graph2 {
 
     constructor() {
         super();
+        this
+            .vertexRenderer(Vertex3)
+            .centroidRenderer(CentroidVertex3)
+            .edgeColor("#287EC4" as any)
```
jbrundage, line 94: Why `as any`?

GordonSmith: For a while it had been PP as a number
I will fix.

packages/util/src/platform.ts
```diff
@@ -6,3 +6,18 @@ export const root: any = new Function("return this;")(); //  Prevent bundlers fr
...
+    for (let i = scripts.length - 1; i >= 0; --i) {
+        const script: any = scripts[i];
+        const attrVal: string = script.getAttribute.length !== undefined ? script[attr] : script.getAttribute(attr) || "";
+        const idx = attrVal.indexOf(partial);
+        if (idx >= 0) {
+            return attrVal.substring(0, idx);
+        }
+    }
+    return "";
+}
```
jbrundage, line 23: This function could be:
```js
export function getScriptSrc(partial: string) {
    for(const script of document.scripts) {
        if(!script.src)continue;
        const idx = script.src.indexOf(partial);
        if(idx !== -1)return script.src.substring(0, idx);
    }
    return "";
}
```

GordonSmith: Yes (need to iterate backwards though)

PR #3639, opened 2020-06-17
fix: Column/Bar events moved to parent dataCell

### LINKED ISSUE #3631: Bar Chart: Make Inner Text and Value be part of the on click event of the chartPanel
Hi,

We would like to request for the onclick event of the chartPanel to include everything within the bar (innerText and value).

Thank you!

jbrundage:
Fixes #3631

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/chart/src/Column.ts
+++ b/packages/chart/src/Column.ts
@@ -173,21 +173,22 @@ export class Column extends XYAxis {
                             context.tooltip.show.apply(context, arguments);
                         }
                     })
+                    .call(host._selection.enter.bind(host._selection))
+                    .on("click", function (d: any) {
+                        context.click(host.rowToObj(d.origRow), d.column, host._selection.selected(this));
+                    })
+                    .on("dblclick", function (d: any) {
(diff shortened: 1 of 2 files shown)


packages/chart/src/Column.ts
```diff
@@ -306,7 +307,7 @@ export class Column extends XYAxis {
                     const dataTextEnter = dataText.enter().append("g")
                         .attr("class", "dataText")
                         .each(function (this: SVGElement, d) {
-                            context.textLocal.set(this, new Text().target(this));
+                            context.textLocal.set(this, new Text().target(this).colorStroke_default("transparent"));
```
GordonSmith, line 310: Why is "transparent" needed by default?

jbrundage: When the parent element has the "over" class (added on hover) this style is applied to the inner text:
https://github.com/hpcc-systems/Visualization/blob/master/packages/api/src/ITooltip.css#L77

PR #3671, opened 2020-07-15
feat: Vertex3 placement adjusts for missing data

jbrundage:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/react/src/vertex3.tsx
+++ b/packages/react/src/vertex3.tsx
@@ -59,7 +59,10 @@ export const Vertex3: React.FunctionComponent<IVertex3> = ({
     const annoOffsetY = 0;
 
     const labelWidth = Utility.textSize(text, textFontFamily, textHeight, false).width;
-    const labelShapeWidth = labelWidth + (textPadding * 2) + (textboxStrokeWidth * 2);
+    let labelShapeWidth = 0;
+    if(text !== ""){
+        labelShapeWidth = labelWidth + (textPadding * 2) + (textboxStrokeWidth * 2)
+    }
     fullAnnotationWidth += labelShapeWidth + annotationGutter;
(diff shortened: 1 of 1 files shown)


packages/react/src/vertex3.tsx
```diff
@@ -59,7 +59,10 @@ export const Vertex3: React.FunctionComponent<IVertex3> = ({
     const annoOffsetY = 0;
 
     const labelWidth = Utility.textSize(text, textFontFamily, textHeight, false).width;
-    const labelShapeWidth = labelWidth + (textPadding * 2) + (textboxStrokeWidth * 2);
+    let labelShapeWidth = 0;
+    if(text !== ""){
```
GordonSmith, line 63: Can text be `undefined`?

packages/react/src/vertex3.tsx
```diff
@@ -88,7 +91,7 @@ export const Vertex3: React.FunctionComponent<IVertex3> = ({
         fullAnnotationWidth += annotationGutter * (annotations.length - 1);
     }
     const textElement = <g transform={`translate(${textOffsetX} ${annoOffsetY})`}>
-        <TextBox
+        { text === "" ? null : <TextBox
```
GordonSmith, line 94: The "norm" seems to be to use "undefined" here instead of null (but suspect both work as expected)

packages/react/src/vertex3.tsx
jbrundage: I don't think it can because the default value is set to empty string. If someone explicitly set text to `undefined` it should just default to empty string, but I'll test it.

jbrundage: Yes, if someone sets text to `undefined` it just defaults to that empty string.

PR #3718, opened 2020-10-16
feat(dataflow): Add support for "sensors"

GordonSmith:
BREAKING CHANGE: chain renamed to pipe
BREAKING CHANGE: ScalarActivities refactored into "observers" and "sensors"

## Checklist:
  - [x] The commit is signed.


packages/dataflow/src/__tests__/readme.ts
```diff
@@ -1,25 +1,78 @@
...
+            filter(row => row.index % 2 === 0), //  Filter even row indecies 
+            sensor(c3),                         //  Keep running count of final rows
+            sensor(m1),                         //  Track largest value
+        );
+
+        for (const row of p1(generate(Math.random, 1000000))) {
+            if (row.index % 100000 === 0) {
+                console.log(`${row.index}: ${c1.peek()}, ${c2.peek()}, ${c3.peek()}, ${m1.peek()}`);
+            }
+        }
```
jbrundage, line 70: This _is_ interesting. So each row yielded by the generator is piped through p1 one at a time? I wonder if the sensors prevent garbage collection?

jbrundage (approved):

packages/dataflow/src/__tests__/readme.ts
GordonSmith: In this example there is no memory overhead as its one row at a time "flowing" through the pipe, so no need for garbage collection at all...

GordonSmith: Unfortunatly the same isn't true for all activities (I am looking at you `sort`).

PR #3818, opened 2021-06-04
feat(comms): Update WsDFU API

GordonSmith:

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/services/wsDFU.ts
+++ b/packages/comms/src/services/wsDFU.ts
@@ -3,8 +3,8 @@ import { Service } from "../espConnection";
 
 /*
     Response structures generated via:
-    * http://localhost:8010/WsDfu/?ver_=1.5&reqjson_
-    * http://localhost:8010/WsDfu/?ver_=1.5&respjson_
+    * http://localhost:8010/WsDfu/?reqjson_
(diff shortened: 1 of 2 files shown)


packages/comms/src/services/wsDFU.ts
```diff
@@ -493,6 +508,8 @@ export namespace WsDfu {
         Persistent: boolean;
         IsProtected: boolean;
         KeyType: string;
+        NumOfSubfiles: number;
+        Accessed: string;
```
jeclrsg, line 512: @GordonSmith Not really relevant to this pr, but I was wondering if TS had something like a date type? I noticed there were a handful of members in a few interfaces in here that, like this one, looked like they were most likely dates. Or maybe these date-like strings don't necessarily always conform to the same format when they come in? Might not be useful or worth changing.

GordonSmith: They are somewhat consistent, but do not match the JS Date serialization format...

Typically:
* Date:  2021-05-21
* Time:  13:33:22.123
* Date Time:  2021-05-21T13:33:22.123Z  (ISO Date Time I think)

PR #3886, opened 2022-01-06
feat(utils): Wsdl to TS typings

jeclrsg:
## Checklist:
  - [ ] The commit is signed.
- [] The change has been fully tested:


GordonSmith:
A potential (and correct) output for "WUQueryActivationMode" could be:
```
export enum WUQueryActivationMode {
    Do_not_activate_query = 0,
    Activate_query = 1,
    Activate_query__suspend_previous = 2,
    Activate_query__delete_previous = 3
}
```

GordonSmith:
Next you could generate the calling stub for each operation:
```
<operation name="WUAction">
<soap:operation soapAction="WsWorkunits/WUAction?ver_=1.85" style="document"/>
<input>
<soap:body use="literal"/>
</input>
<output>
<soap:body use="literal"/>
</output>
<fault name="excfault">
<soap:fault name="excfault" use="literal"/>
</fault>
</operation>
```

You will need to use the associated `SoapIn` and `SoapOut` `messages`:
```
<message name="WUActionSoapIn">
<part name="parameters" element="tns:WUAction"/>
</message>
<message name="WUActionSoapOut">
<part name="parameters" element="tns:WUActionResponse"/>
</message>
```

GordonSmith:
FYI - I tidied up the util folder a bit in this draft PR:  https://github.com/jeclrsg/Visualization/pull/1

packages/comms/utils/index.ts
```diff
@@ -0,0 +1,253 @@
...
+                }
+                parsedTypes[propName] = defn;
+            } else {
+                typeDefn[propName] = prop;
+            }
+
+        } else {
+            if (ignoredWords.indexOf(prop) < 0) {
+                const primitiveType = operation[prop].replace(/xsd:/gi, "");
+                if (prop.indexOf("[]") > 0) {
```
GordonSmith, line 123: `prop.endsWith("[]")` might be better here?

packages/comms/utils/index.ts
```diff
@@ -0,0 +1,253 @@
...
+                    typeDefn[prop.slice(0, -2)] = primitiveType + "[]";
+                } else if (operation[prop].match(/.*\|.*\|.*/)) {
```
GordonSmith, line 125: I have no idea what this looks like, but does this catch enums with only 2 options (TRUE|FALSE)?

packages/comms/utils/index.ts
```diff
@@ -0,0 +1,253 @@
...
+        })
+    };
+}
+
+function parseTypeDefinition(operation: JsonObj, opName: string, types) {
+
+    const typeDefn: JsonObj = {};
+    printDbg(`processing ${opName}`, operation);
+    for (const prop in operation) {
+        const propName = (prop.indexOf("[]") < 0) ? prop : prop.slice(0, -2);
```
GordonSmith, line 104: prop.endsWith("[]") might be better here?

packages/comms/utils/index.ts
```diff
@@ -0,0 +1,253 @@
...
+            lines.push(`type ${primitive} = ${primitiveMap[primitive]};`);
+        });
+        lines.push("\n\n");
+
+        for (const name in parsedEnums) {
+            lines.push(`export enum ${name} {\n\n`);
+            lines.push(parsedEnums[name].join(",\n"));
+            lines.push("\n\n}");
+            lines.push("\n\n");
+        }
```
GordonSmith, line 189: Minor, but the following might be easier to read (and the same for your other "templates"?:
```
lines.push(`\
export enum ${name} {
  ${parsedEnums[name].join(",\n  ")}
}

`);
```

packages/comms/utils/index.ts
```diff
@@ -0,0 +1,253 @@
...
+
+        lines.push(`export class ${namespace.replace("Ws", "")}Service extends Service {\n`);
+
+        const methods = [];
+
+        for (const service in bindings) {
+            const binding = bindings[service];
+            for (const method in binding.methods) {
+                const soapAction = binding.methods[method].soapAction;
+                const url = `http://localhost:8010/${soapAction}`;
```
GordonSmith, line 212: `http://localhost:8010` should be extracted from the URL passed in as an argument?

GordonSmith (changes requested):

packages/comms/utils/index.ts
jeclrsg: Added a comment to this. Looking at it again, realized that regex wasn't matching value(s) of an enum, but the string that node soap generates to represent an enum, which is always EnumName|type|values

packages/comms/utils/index.ts
jeclrsg: Added a comment and changed to "example.org" instead

jeclrsg:
I think I covered the most recent requests for changes - the path where the .ts files are written is now of the format .../_ServiceName_/v#/_ServiceName_.ts, and the body of each operation is calling the connection wrapper around the service call. 

I changed the usage comments slightly as well, noting that you can alternatively pass a local path to a WSDL file as well as a URL. Rodrigo was nice enough to provide me with a copy of the ws_logaccess WSDL yesterday afternoon.

GordonSmith:
The request interface need to optional either by:
```
WUQuery(request: Partial<WsWorkunits.WUQuery>): Promise<WsWorkunits.WUQueryResponse> {
	return this._connection.send("WUQuery", request);
}
```
or
```
export interface WUQuery {
    Wuid?: string;
    Type?: string;
    Cluster?: string;
    RoxieCluster?: string;
    Owner?: string;
    State?: string;
    StartDate?: string;
    EndDate?: string;
    ECL?: string;
    Jobname?: string;
    LogicalFile?: string;
    LogicalFileSearchType?: string;
...
}
```

I suspect the later is needed for nested structures

GordonSmith (changes requested):
See comment re optional fields in the reqeusts

GordonSmith (approved):

PR #3946, opened 2022-03-29
feat(comms): add WsDFU service generated from wsdl

jeclrsg:
Signed-off-by: Jeremy Clements <jeremy.clements@lexisnexisrisk.com>

## Checklist:
  - [x] The commit is signed.


packages/comms/src/ecl/logicalFile.ts
```diff
@@ -72,14 +74,15 @@ export class LogicalFile extends StateObject<FileDetailEx, FileDetailEx> impleme
     get IsCompressed(): boolean { return this.get("IsCompressed"); }
     get BrowseData(): boolean { return this.get("BrowseData"); }
     get jsonInfo(): string { return this.get("jsonInfo"); }
-    get binInfo(): string { return this.get("binInfo"); }
+    get binInfo(): base64Binary { return this.get("binInfo"); }
```
jeclrsg, line 77: This was one of those synonymous number types, but I'm not sure if this is right?

packages/comms/src/services/wsdl/WsDfu/v1.62/WsDfu.ts
```diff
@@ -0,0 +1,1455 @@
+import { IConnection, IOptions } from "../../../../connection";
+import { Service } from "../../../../espConnection";
+
+type base64Binary = number[];
```
GordonSmith, line 4: Did you double check that this is a `number[]` and not a `string` when its received?

GordonSmith: @jeclrsg - did you get anywhere with this?

GordonSmith (changes requested):
You should keep the "base64Binary" type, but type it to be a string (as it does really help with the self documentation).

PR #3974, opened 2022-05-09
feat(wsdl): use typescript-formatter to reformat .ts files according to rules

jeclrsg:
Signed-off-by: Jeremy Clements <jeremy.clements@lexisnexisrisk.com>

## Checklist:
  - [ ] The commit is signed.

### DIFF
--- a/packages/comms/package.json
+++ b/packages/comms/package.json
@@ -78,6 +78,7 @@
         "d3-time-format": "^2",
         "soap": "0.43.0",
         "tslib": "2.3.1",
+        "typescript-formatter": "^7.2.2",
         "whatwg-fetch": "3.6.2"
     },
     "repository": {
--- a/packages/comms/utils/index.ts
+++ b/packages/comms/utils/index.ts
@@ -4,6 +4,7 @@ import { mkdirp, writeFile } from "fs-extra";
 import * as path from "path";
 import * as soap from "soap";
 import minimist from "minimist";
+import * as tsfmt from "typescript-formatter";
 
 import { Case, changeCase } from "./util";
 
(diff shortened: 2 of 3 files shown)


packages/comms/package.json
```diff
@@ -78,6 +78,7 @@
         "d3-time-format": "^2",
         "soap": "0.43.0",
         "tslib": "2.3.1",
+        "typescript-formatter": "^7.2.2",
```
GordonSmith, line 81: You will need to include the updated package-lock.json file with this PR

packages/comms/utils/index.ts
```diff
@@ -38,6 +39,11 @@ const outDir = args?.outDir ? args?.outDir : "./temp/wsdl";
 
 const ignoredWords = ["targetNSAlias", "targetNamespace"];
 
+const tsFmtOpts = {
+    replace: true, verify: false, tsconfig: true, tsconfigFile: null, tslint: true, tslintFile: null,
+    editorconfig: true, vscode: true, vscodeFile: null, tsfmt: false, tsfmtFile: null
+};
+
```
GordonSmith, line 46: Do these options match the VSCode formatting?  IOW if I open a generated file and re format it in VSCode will it change?

GordonSmith: What does `replace: true` do?

packages/comms/package.json
```diff
@@ -69,7 +69,7 @@
         "tmp": "0.2.1"
     },
     "devDependencies": {
-        "@hpcc-js/bundle": "^2.11.1",
+        "@hpcc-js/bundle": "^2.10.23",
```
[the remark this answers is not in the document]
jeclrsg: Yes, I forgot to change that back. I was getting an error for that when installing dependencies, probably because I had done so from within the comms package. But I had looked at https://www.npmjs.com/package/@hpcc-js/bundle and didn't see that version listed. I didn't think to check the version in the actual package.

packages/comms/utils/index.ts
jeclrsg: They do match the current VSCode formatting, I think perhaps only coincidentally because there's not anything defined that overrides the defaults styles.

The replace option decides whether the script should overwrite the file in place.

GordonSmith (approved):

PR #4055, opened 2022-12-07
feat: bump WsDFU to latest to include support for ExtendedIndexInfo

jeclrsg:
Signed-off-by: Jeremy Clements <Jeremy.Clements@lexisnexisrisk.com>

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/package.json
+++ b/packages/comms/package.json
@@ -69,14 +69,14 @@
         "tmp": "0.2.1"
     },
     "devDependencies": {
-        "@hpcc-js/bundle": "^2.11.3",
+        "@hpcc-js/bundle": "^2.10.23",
         "@types/d3-request": "1.0.6",
         "@types/d3-time-format": "2.3.1",
         "@types/node": "^10",
         "@types/xmldom": "0.1.31",
         "d3-array": "^1",
         "d3-time-format": "^2",
-        "soap": "0.45.0",
+        "soap": "0.43.0",
         "tslib": "2.4.1",
         "typescript-formatter": "^7.2.2",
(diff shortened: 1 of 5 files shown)


packages/comms/package.json
```diff
@@ -69,10 +69,10 @@
         "tmp": "0.2.1"
     },
     "devDependencies": {
-        "@hpcc-js/bundle": "^2.11.3",
+        "@hpcc-js/bundle": "^2.10.23",
```
jeclrsg, line 72: When I was building this to test, this version number showed as incorrect; there doesn't seem to be a 2.11.3 for bundle

packages/comms/package.json
```diff
@@ -69,10 +69,10 @@
...
         "@types/d3-request": "1.0.6",
         "@types/d3-time-format": "2.3.1",
-        "@types/node": "^10",
+        "@types/node": "^14",
```
jeclrsg, line 75: I don't believe this caused any problems, as I was able to build without warnings or errors. It was a required for a change in the "fs-extra" dependency, which was causing any of the "wsdl-..." npm scripts to fail

GordonSmith: "@types/node": "^16" would match what we are actually using?

GordonSmith: Looks like the CI failed because your package-lock file is out of sync (do a npm install and commit).

GordonSmith (changes requested):

GordonSmith (approved):
(Waiting on tests to rerun)

PR #4073, opened 2023-03-13
GH-4072 Graph2 add support for image icons

### LINKED ISSUE #4072: Adding image support to the Graph2
We are using Graph2 of Visualization project of HPCC-Systems for our project. We want to use images more specifically country images instead of icons in the Graph2 Chart. Please see the below link for the project we are referring.

https://raw.githack.com/hpcc-systems/Visualization/trunk/apps/docs/index.html

We already have a folder which we store the country flags in png format but ideally it should be like  tags in HTML which allows us to use it with file in the project, link or source. 

Please see the below image. 

images for hpcc graph

Basically icons should be replaced by an image either from a link or by an .png .jpeg etc. (basically any kind of image) file

GordonSmith:
DataGraph:
 * vertexImageUrl
 * vertexImageUrlColumn

Icon:
 * imageUrl

Fixes #4072

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/react/src/image.tsx
+++ b/packages/react/src/image.tsx
@@ -1,0 +1,25 @@
+import * as React from "@hpcc-js/preact-shim";
+
+interface Image {
+    href: string;
+    x?: number;
+    y?: number;
+    height?: number;
+    yOffset?: number;
+}
+
+export const Image: React.FunctionComponent<Image> = ({
+    href,
+    x,
+    y = 0,
+    height = 12
+}) => {
+
+    return <image
+        href={href}
+        x={x - height / 2}
+        y={y - height / 2}
+        width={height}
+        height={height}
+    ></image>;
+};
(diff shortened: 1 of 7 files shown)


packages/react/src/image.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const Image: React.FunctionComponent<Image> = ({
+    href,
+    x,
+    y = 0,
+    height = 12
+}) => {
+
+    return <image
```
jeclrsg, line 18: Should this just be an `` or `<picture>`? Or is there a library somewhere defining that <image> element?

jeclrsg (commented):
Wasn't sure about that <image> tag. If that's the native one, I don't think it's standard anymore.

packages/react/src/image.tsx
GordonSmith: Its one of these:  https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image

jeclrsg (approved):

PR #4084, opened 2023-05-24
fix(comms): incorrect default for line count returned by GetLogs

jeclrsg:
eg, when TotalLogLinesAvailable is 0, this returns 10000

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/services/wsLogaccess.ts
+++ b/packages/comms/src/services/wsLogaccess.ts
@@ -199,7 +199,7 @@ export class LogaccessService extends LogaccessServiceBase {
                 }
                 return {
                     lines: lines,
-                    total: response.TotalLogLinesAvailable || 10000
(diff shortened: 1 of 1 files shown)


packages/comms/src/services/wsLogaccess.ts
```diff
@@ -199,7 +199,7 @@ export class LogaccessService extends LogaccessServiceBase {
                 }
                 return {
                     lines: lines,
-                    total: response.TotalLogLinesAvailable || 10000
+                    total: response.TotalLogLinesAvailable ?? 0
```
GordonSmith, line 202: If the total number of log lines is unknown (undefined) it does not mean, there are none...  I think we should keep the 10000 to prevent any compatibility issues.

GordonSmith (approved):

GordonSmith (approved):

PR #4100, opened 2023-06-19
feat(comms): update, abort, delete methods for dfuWorkunit

jeclrsg:
for use in ECL Watch DFU Workunit page, methods implemented include update, abort, delete, fetchXML

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/ecl/dfuWorkunit.ts
+++ b/packages/comms/src/ecl/dfuWorkunit.ts
@@ -2,7 +2,7 @@ import { Cache, IEvent, scopedLogger, StateCallback, StateEvents, StateObject, S
 import { IConnection, IOptions } from "../connection";
 import { ESPExceptions } from "../espConnection";
 import { WsSMC } from "../services/wsSMC";
-import { FileSpray, SprayFixedEx, FileSprayService, SprayVariableEx } from "../services/fileSpray";
+import { FileSpray, FileSprayService, SprayFixedEx, SprayVariableEx, UpdateDFUWorkunitEx } from "../services/fileSpray";
 import * as WsTopology from "../services/wsTopology";
 
 const logger = scopedLogger("@hpcc-js/comms/dfuWorkunit.ts");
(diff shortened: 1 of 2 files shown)


packages/comms/src/ecl/dfuWorkunit.ts
```diff
@@ -154,6 +154,20 @@ export class DFUWorkunit extends StateObject<UDFUWorkunitState, IDFUWorkunitStat
...
+        return this.connection.UpdateDFUWorkunitEx({
+            wu: {
+                ...request,
+                ID: this.ID,
+                State: this.State
+            },
+            ClusterOrig: this.ClusterName,
+            JobNameOrig: this.JobName,
+            isProtectedOrig: this.isProtected,
+            StateOrig: this.State
```
GordonSmith, line 167: Why is update needed?  For protected?  Any fields that should not be changed, you should be able to omit.

packages/comms/src/ecl/dfuWorkunit.ts
```diff
@@ -212,6 +238,36 @@ export class DFUWorkunit extends StateObject<UDFUWorkunitState, IDFUWorkunitStat
...
+                return response;
+            });
+        });
+    }
+
+    protected AbortDFUWorkunit(): Promise<FileSpray.AbortDFUWorkunitResponse> {
+        return this.connection.AbortDFUWorkunit({
+            wuid: this.ID
+        });
+    }
```
GordonSmith, line 269: This is probably not need, could be inside the "abort" method above?

packages/comms/src/services/fileSpray.ts
```diff
@@ -100,6 +100,134 @@ export interface DesprayEx {
...
+    failIfNoSourceFile?: boolean;
+    recordStructurePresent?: boolean;
+    quotedTerminator?: boolean;
+    preserveCompression?: boolean;
+    expireDays?: number;
+    PreserveFileParts?: boolean;
+    FileAccessCost?: number;
+    KbPerSecAve?: number;
+    KbPerSec?: number;
+}
```
GordonSmith, line 229: These are duplicated from the generated code?  You can use `Partial<FileSpray.UpdateDFUWorkunit>` to achieve the same thing?

GordonSmith (changes requested):

packages/comms/src/ecl/dfuWorkunit.ts
jeclrsg: The ECL Watch UI only allows JobName and Protected to be updated currently.

packages/comms/src/services/fileSpray.ts
jeclrsg: Yeah, I actually started there. But I couldn't work out why type the ECL Watch side was clashing because of that wu: {...} property that contains all the actual properties. I'm not sure what I was missing, but it would either end up submitting something like `...wu.wu.Jobname: "xyz"...` or `...wu: {...}, JobName: "xyz"...` which is invalid and was ignored...

GordonSmith: Do you think this is an issue with the wsdl generation or with the WSDL definition?

jeclrsg: I think it's the generation. Seems like it's not taking into account the "minOccurs" attribute to indicate whether the properties in the created interfaces should be optional.

GordonSmith (approved):

PR #4103, opened 2023-06-28
fix:  Query fails fetching schemas

GordonSmith:
## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/ecl/query.ts
+++ b/packages/comms/src/ecl/query.ts
@@ -1,7 +1,10 @@
-import { Cache, StateObject } from "@hpcc-js/util";
+import { Cache, StateObject, scopedLogger } from "@hpcc-js/util";
 import { IConnection, IOptions } from "../connection";
 import { EclService, IWsEclRequest, IWsEclResponse, IWsEclResult } from "../services/wsEcl";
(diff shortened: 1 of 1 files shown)


packages/comms/src/ecl/query.ts
```diff
@@ -76,41 +74,81 @@ export class Query extends StateObject<QueryEx, QueryEx> implements QueryEx {
         return retVal;
     }
 
+
+    private _eclService: Promise<EclService>;
+    protected async wsEclService(): Promise<EclService | undefined> {
+        if (!this._eclService) {
+            this._eclService = this.topology.fetchServices({}).then(services => {
+                for (const espServer of services?.TpEspServers?.TpEspServer) {
```
jeclrsg, line 82: These couple for...of loops were the only thing I saw as possible problems, and GH actions already made annotations. Optionally chained forEach()s instead maybe? Though is it possible for the ESP responses to even have those as undefined, rather than empty arrays? I don't think I've ever run into that at least.

GordonSmith:
@jeclrsg please re-review.

jeclrsg (approved):

PR #4108, opened 2023-08-08
fix(comms): compiled WUs not always considered complete

jeclrsg:
## Checklist:
  - [x] The commit is signed.


GordonSmith (changes requested):
```
this.State === "compiled" 
``` 
 and
```
  switch (this.StateID) {
            case WUStateID.Compiled:
```
Is the same thing?  IOW it will return TRUE when it shouldn't?

jeclrsg:
Looking into changing how polling WUs in ECL Watch works instead of this change.

PR #4131, opened 2023-10-19
fix(comms): extend FileSpray.DFUWUFile to change reponse type to text

jeclrsg:
consistent with WsWorkunits.WUFile

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/services/fileSpray.ts
+++ b/packages/comms/src/services/fileSpray.ts
@@ -13,6 +13,10 @@ export interface UpdateDFUWorkunitEx extends UpdateDFUWorkunitMinusWU {
 
 export class FileSprayService extends FileSprayServiceBase {
 
+    DFUWUFileEx(request: FileSpray.DFUWUFileRequest): Promise<string> {
+        return this._connection.send("DFUWUFile", request, "text");
+    }
+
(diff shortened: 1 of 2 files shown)


packages/comms/src/services/fileSpray.ts
```diff
@@ -13,6 +13,10 @@ export interface UpdateDFUWorkunitEx extends UpdateDFUWorkunitMinusWU {
 
 export class FileSprayService extends FileSprayServiceBase {
 
+    DFUWUFileEx(request: FileSpray.DFUWUFileRequest): Promise<string> {
```
GordonSmith, line 16: Should keep the `DFUWUFile` name (especially since the generated version is invalid).

GordonSmith (changes requested):

packages/comms/src/services/fileSpray.ts
jeclrsg: I haven't been able to find a solution to this that doesn't seem like working around the rules. I did come across this answer on SO, which seemed like a good breakdown https://stackoverflow.com/a/62515586. Specifically, the third part where they suggested making the extended type be generic with a default of the preferred type. I thought that might be a good enhancement to the wsdl generator?

GordonSmith (approved):

PR #4140, opened 2023-11-02
fix(html/common): correct BreakdownTable tooltip height calculation

jeclrsg:
## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/common/src/Utility.ts
+++ b/packages/common/src/Utility.ts
@@ -736,11 +736,17 @@ export function textSize(_text: string | string[], fontName: string = "Verdana",
     const hash = `${bold}::${fontSize}::${fontName}::${text.join("::")}`;
     let retVal = g_fontSizeContextCache[hash];
     if (!retVal) {
+        retVal = { width: 0, height: 0 };
         g_fontSizeContext.font = `${bold ? "bold " : ""}${fontSize}px ${fontName}`;
-        g_fontSizeContextCache[hash] = retVal = {
-            width: Math.max(...text.map(t => g_fontSizeContext.measureText("" + t).width)),
-            height: fontSize * text.length
-        };
+        text.forEach(line => {
+            const textMeasurement = g_fontSizeContext.measureText("" + line);
+            const width = textMeasurement.width;
+            const height = (textMeasurement.fontBoundingBoxDescent + textMeasurement.fontBoundingBoxAscent) * text.length;
+            g_fontSizeContextCache[hash] = retVal = {
+                width: width > retVal.width ? width : retVal.width,
+                height: height > retVal.height ? height : retVal.height
+            };
+        });
     }
     return retVal;
 }
--- a/packages/html/src/BreakdownTable.ts
+++ b/packages/html/src/BreakdownTable.ts
@@ -53,8 +53,8 @@ export class BreakdownTable extends StyledTable {
         this._tooltip
             .tooltipHTML(data => {
                 const rowCount = this.useCalculatedRowCount() ? this.calculateRowCount() : this.rowCount();
-                const rowHeight = this.fontSize();
-                const widestLabel = Math.max(...data.map(row => this.textSize(row[0], "Verdana", this.fontSize()).width));
+                const rowHeight = Math.max(...data.map(row => this.textSize(row[0], this.fontFamily(), this.fontSize()).height)) ?? this.fontSize();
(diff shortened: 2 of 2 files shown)


packages/html/src/BreakdownTable.ts
```diff
@@ -53,8 +53,8 @@ export class BreakdownTable extends StyledTable {
         this._tooltip
             .tooltipHTML(data => {
                 const rowCount = this.useCalculatedRowCount() ? this.calculateRowCount() : this.rowCount();
-                const rowHeight = this.fontSize();
-                const widestLabel = Math.max(...data.map(row => this.textSize(row[0], "Verdana", this.fontSize()).width));
+                const rowHeight = Math.max(...data.map(row => this.textSize(row[0], this.fontFamily(), this.fontSize()).height)) ?? this.fontSize();
```
jeclrsg, line 56: I realized that fontFamily was already a published property in the StyledTable extended class, it just wasn't being used for these font size calculations.

packages/html/src/BreakdownTable.ts
```diff
@@ -65,13 +65,13 @@ export class BreakdownTable extends StyledTable {
                 return `<div style="
                     width: 100%;
                     height: 100%;
-                    font-size: ${this.fontSize()}px
-                ">${
-                    otherData.map(row => `<div style="
+                    font-size: ${this.fontSize()}px;
+                ">${otherData.map(row =>
```
jeclrsg, line 69: I believe all this was just automatically changed by VSCode when I saved this file.

packages/common/src/Utility.ts
```diff
@@ -739,7 +739,10 @@ export function textSize(_text: string | string[], fontName: string = "Verdana",
         g_fontSizeContext.font = `${bold ? "bold " : ""}${fontSize}px ${fontName}`;
         g_fontSizeContextCache[hash] = retVal = {
             width: Math.max(...text.map(t => g_fontSizeContext.measureText("" + t).width)),
-            height: fontSize * text.length
+            height: Math.max(...text.map(t => {
+                const textMeasurement = g_fontSizeContext.measureText("" + t);
+                return textMeasurement.fontBoundingBoxDescent + textMeasurement.fontBoundingBoxAscent;
```
jeclrsg, line 744: Based upon these docs https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics#examples

GordonSmith: 1.  The result still needs to be multiplied by `text.length`
2. (optional) The calculation for width and height could be moved into a single `text.reduce` function.

GordonSmith (changes requested):

packages/common/src/Utility.ts
jeclrsg: I couldn't think of a way to make `text.reduce` work, because the type of the accumulator must that of the individual items in the array, rather than whatever you want, eg an object containing the dimensions.

packages/common/src/Utility.ts
```diff
@@ -736,11 +736,17 @@ export function textSize(_text: string | string[], fontName: string = "Verdana",
     const hash = `${bold}::${fontSize}::${fontName}::${text.join("::")}`;
     let retVal = g_fontSizeContextCache[hash];
     if (!retVal) {
+        let dimensions: TextSize = { width: 0, height: 0 };
```
GordonSmith, line 739: Change to `retVal = { width: 0, height: 0 };`
And then remove dimension below...

packages/common/src/Utility.ts
```diff
@@ -736,11 +736,17 @@ export function textSize(_text: string | string[], fontName: string = "Verdana",
...
-        };
+        text.forEach(line => {
+            const textMeasurement = g_fontSizeContext.measureText("" + line);
+            const width = textMeasurement.width;
+            const height = (textMeasurement.fontBoundingBoxDescent + textMeasurement.fontBoundingBoxAscent) * text.length;
+            g_fontSizeContextCache[hash] = retVal = dimensions = {
+                width: width > dimensions.width ? width : dimensions.width,
+                height: height > dimensions.height ? height : dimensions.height
+            };
+        });
```
GordonSmith, line 749: Optional:  Replace with `text.reduce`

GordonSmith (changes requested):

GordonSmith (approved):

PR #4156, opened 2024-01-19
feat:  Parse the DefinitionList at source

GordonSmith:
Ensure there are always "some" values for TimeLocalExecute

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/ecl/workunit.ts
+++ b/packages/comms/src/ecl/workunit.ts
@@ -22,6 +22,118 @@ function formatNum(num: number | string): string {
     }
     return num as string;
 }
+const DEFINITION_LIST = "DefinitionList";
+const definitionRegex = /([a-zA-Z]:)?(.*[\\\/])(.*)(\((\d+),(\d+)\))/;
+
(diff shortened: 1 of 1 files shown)


packages/comms/src/ecl/workunit.ts
```diff
@@ -22,6 +22,8 @@ function formatNum(num: number | string): string {
     }
     return num as string;
 }
+const DEFINITION_LIST = "DefinitionList";
+const definitionRegex = /([a-zA-Z]:)?(.*[\\/])(.*)(\((\d+),(\d+)\))/;
```
jeclrsg, line 26: I think forward slashes should also be escaped? `[\\\/]`

jeclrsg (changes requested):
@GordonSmith maybe a small problem in that regex? Not sure, tbh. I think it works either way.

GordonSmith:
@jeclrsg converting to draft as there are some more changes that should go in here...

jeclrsg (approved):

PR #4173, opened 2024-02-26
feat(comms): add publishEx function to workunit

jeclrsg:
extended version of publish(...) using the request definition in WsWorkunits v1.98

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/ecl/workunit.ts
+++ b/packages/comms/src/ecl/workunit.ts
@@ -1096,6 +1096,20 @@ export class Workunit extends StateObject<UWorkunitState, IWorkunitState> implem
         });
     }
 
+    publishEx(request: Partial<WsWorkunits.WsWorkunits.WUPublishWorkunit>) {
+        const service = new WsWorkunits.WorkunitsServiceEx({ baseUrl: "" });
+        const publishRequest = {
+            Wuid: this.Wuid,
+            Cluster: this.Cluster,
+            JobName: this.Jobname,
+            AllowForeignFiles: true,
+            Activate: 1,
+            Wait: 5000,
+            ...request
+        };
+        return service.WUPublishWorkunitEx(publishRequest);
+    }
+
     protected WUCDebug(command: string, opts: any = {}): Promise<XMLNode | null> {
         let optsStr = "";
         for (const key in opts) {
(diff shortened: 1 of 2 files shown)


packages/comms/src/services/wsWorkunits.ts
```diff
@@ -10,6 +10,49 @@ import { ESPConnection, Service } from "../espConnection";
...
+        IncludeFileErrors?: boolean;
+        DfuCopyFiles?: boolean;
+        DfuQueue?: string;
+        DfuWait?: number;
+        DfuOverwrite?: boolean;
+        OnlyCopyFiles?: boolean;
+        StopIfFilesCopied?: boolean;
+        DfuPublisherWuid?: string;
+        RemoteStorage?: string;
+    }
```
GordonSmith, line 52: Where did this come from?

jeclrsg: I'd added that manually for use as the type in the extended publish function. The interfaces generated from the WSDL don't currently respect mandatory/optional parameters. Or rather, it often seems to imply every property is optional, which doesn't seem correct. Unless I'm misunderstanding what minOccurs=0 means. I was going to ask about that in the next ECL Watch meeting. 

image

packages/comms/src/ecl/workunit.ts
```diff
@@ -1096,6 +1096,22 @@ export class Workunit extends StateObject<UWorkunitState, IWorkunitState> implem
...
+            ...{
+                Wuid: this.Wuid,
+                Cluster: this.Cluster,
+                JobName: this.Jobname,
+                AllowForeignFiles: true,
+                Activate: 1,
+                Wait: 5000
+            },
+            ...request
+        };
```
GordonSmith, line 1111: Could be shortened to:
```ts
        const publishRequest = {
            Wuid: this.Wuid,
            Cluster: this.Cluster,
            JobName: this.Jobname,
            AllowForeignFiles: true,
            Activate: 1,
            Wait: 5000,
            ...request
        } as WsWorkunits.WsWorkunits.WUPublishWorkunit;
        return service.WUPublishWorkunit(publishRequest);
```

Then you wouldn't need a `WUPublishWorkunitEx` overload?

GordonSmith (changes requested):

GordonSmith (approved):

PR #4174, opened 2024-03-04
feat: Add showDomainTotal to Column Chart 

GordonSmith:
## Checklist:
  - [x] The commit is signed.


jeclrsg (approved):

GordonSmith:
@jeclrsg can you re-review:
* Enabled Total to work on stacked and regular column charts
* Moved Total text out of the rect rendering loop and into the "row" rendering loop

packages/chart/src/Column.ts
```diff
@@ -209,7 +205,7 @@ export class Column extends XYAxis {
                                 valueText = d3Format(context.showValueAsPercentFormat())(valueText / seriesSum);
                                 break;
                             case "domain":
-                                const domainSum = typeof dm.sum !== "undefined" ? dm.sum : domainSums[d.idx];
+                                const domainSum = typeof dm.sum !== "undefined" ? dm.sum : domainSums[d.idx - 1];
```
jeclrsg, line 208: I missed this before, but could valueText end up as NaN when dm.sum is undefined and d.idx is 0 (domainSums[-1])?

jeclrsg (changes requested):
@GordonSmith Just noticed one array index issue (maybe) that I didn't see the first time

packages/chart/src/Column.ts
GordonSmith: That is taken care of with the filter here:  https://github.com/hpcc-systems/Visualization/pull/4174/files#diff-d152ca42ea481d31dcf64caff056f9748655185f1cc12aa26ae12d8f685c332cR157

jeclrsg (approved):

PR #4203, opened 2024-06-14
fix(comms): wsLogaccess GetLogs results not sorted by default

jeclrsg:
logs returned by the log engine don't seem to use a default sort, so some default should be supplied in the request

## Checklist:
  - [x] The commit is signed.


jeclrsg:
Though the Logs grid in ECL Watch doesn't currently allow user-defined sorting, I was noticing that the rows returned (at least by Elastic) don't seem to be strictly guaranteed to be sorted by some default.

image

GordonSmith (approved):

PR #4215, opened 2024-07-24
fix: Tidy up StatChart to be more consistent

GordonSmith:
## Checklist:
  - [x] The commit is signed.


packages/chart/src/__tests__/heat.ts
```diff
@@ -28,7 +28,8 @@ export class Test extends Heat {
 
 //  Calculate the standard deviation of the weight column  ---
 const extractWeight = df.map(row => row[2]);
-const distribution = df.scalar(df.distribution())(extractWeight(data));
+export const distribution = df.scalar(df.distribution())(extractWeight(data));
+export const quartiles = df.scalar(df.quartile())(extractWeight(data));
```
jeclrsg, line 32: This "quartiles" const doesn't appear to be used later on in the test.

jeclrsg (approved):
@GordonSmith There was one unused const in one of the tests. Looked good otherwise.

packages/chart/src/__tests__/heat.ts
GordonSmith: Its exported and used in the StatChart test

jeclrsg (approved):

PR #4223, opened 2024-08-15
feat:  Add esbuild-plugins package

GordonSmith:
## Checklist:
  - [x] The commit is signed.


packages/esbuild-plugins/src/build.ts
```diff
@@ -0,0 +1,143 @@
...
+        describe: "Watch for changes",
+        type: "boolean"
+    })
+    .help("h")
+    .alias("h", "help")
+    .epilog("https://github.com/hpcc-systems/hpcc-js-wasm")
+    ;
+
+export const pkg = JSON.parse(readFileSync(path.join(process.cwd(), "./package.json"), "utf8"));
+export const NODE_MJS = pkg.type === "module" ? "js" : "mjs";
```
jeclrsg, line 34: Are these backwards, if type === module then mjs?

packages/esbuild-plugins/src/sfx-wrapper.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+} `.trim();
+}
+
+export async function wrap(path: string) {
+    const base91 = await Base91.load();
+    const zstd = await Zstd.load();
+
+    const wasm = await readFile(path);
+    path = path.replace(/\.js$/, ".xxx");
```
jeclrsg, line 85: Is this replacement correct? It didn't seem to get converted back anywhere.

jeclrsg (commented):
@GordonSmith just a couple of questions

packages/esbuild-plugins/src/build.ts
GordonSmith: No - if the type is "module" then it expects all "js" files to be modern "import" style.

packages/esbuild-plugins/src/sfx-wrapper.ts
GordonSmith: Its a bit circular, but without this you would end up with an infinite loop...

jeclrsg (approved):

PR #4250, opened 2024-10-08
feat: Add markdown-it-plugin

GordonSmith:
## Checklist:
  - [x] The commit is signed.


GordonSmith:
FYI - This code is what drives the embedding of interactive visualizations within VitePress (and other) generated sites.  See https://gordonsmith.github.io/hpcc-js-docs/ for the POC

packages/markdown-it-plugins/esbuild.js
```diff
@@ -0,0 +1,7 @@
+import { nodeTpl, browserTpl } from "@hpcc-js/esbuild-plugins";
+
+//  config  ---
+await Promise.all([
+    nodeTpl("src/index.ts", "dist/index.node", { externalx: ["@observablehq/runtime", "@observablehq/stdlib", "@observablehq/inspector"] }),
+    browserTpl("src/index.ts", "dist/index", { externalx: ["@observablehq/runtime", "@observablehq/stdlib", "@observablehq/inspector"] })
```
jeclrsg, line 6: externalx?

packages/markdown-it-plugins/src/fence.ts
```diff
@@ -0,0 +1,59 @@
+import type MarkdownIt from "markdown-it";
+import type { Options } from "markdown-it";
+import type Token from "markdown-it/lib/token.mjs";
+import type Renderer from "markdown-it/lib/renderer.mjs";
```
jeclrsg, line 4: Seems weird that these can't all be one include, but that's just how this library has their stuff structured?

packages/markdown-it-plugins/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+        return retVal;
+    };
+}
+
+function hookVitepressRender(md: MarkdownIt) {
+
+    const originalRender = md.render;
+    md.render = (src: string, env?: any): string => {
+        const myEnv = env ?? {};
+        let retVal = originalRender.call(md, src, myEnv);
```
jeclrsg, line 22: why's this different from how hookRender does the env stuff above?

packages/markdown-it-plugins/src/util.ts
```diff
@@ -0,0 +1,78 @@
...
+    return fetchInfo.exec == true && fetchInfo.hide !== true;
+}
+
+let idx = 0;
+function calcPlaceholders(content: string, fenceInfo: FenceInfo): RenderNode[] {
+    const retVal: RenderNode[] = [];
+    ++idx;
+    try {
+        const cellNb = ojs2notebook(content);
+        for (let i = 0; i < cellNb.nodes.length; ++i) {
```
jeclrsg, line 39: this looks like it always steps over the first node, is that correct?

packages/markdown-it-plugins/tests/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const minimalMd = fetch("/tests/Basic usage/Minimal Example.md").then(response => {
+    return response.text();
+}).catch(e => {
+    console.error("Error: ", e.message ?? e);
+});
+const minimalMdText = await minimalMd;
+
+const minimalHtml = md.render("" + await minimalMd);
+// expect(minimalHtml).to.be.a.string;
```
jeclrsg, line 22: some commented out code in here, looked like it was moved into the index.browser.spec.ts

jeclrsg (commented):
@GordonSmith just a few minor comments. only thing that seemed like it might be a bug was that for loop index in util.ts?

packages/markdown-it-plugins/esbuild.js
GordonSmith: Yea should be removed.

packages/markdown-it-plugins/src/fence.ts
GordonSmith: I agree...  But at least they are only types (we don't bundle marked-it in our dist folder)

packages/markdown-it-plugins/src/index.ts
GordonSmith: Up to that point both are the same (and I could re-structure to make cleaner), its how the Observable render gets triggered that is different depending on the host environment.

packages/markdown-it-plugins/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+        if (env[ENV_KEY]?.length) {
+            const content = encodeURI(JSON.stringify(env[ENV_KEY]));
+            retVal += `<NotebookComponent content="${content}" />`;
```
GordonSmith, line 25: This is specific to the VitePress implementation, on reflection I might want to include the "NotebookComponent" in this packages as a resource...

packages/markdown-it-plugins/src/util.ts
GordonSmith: It shouldn't, both `for (let i = 0; i < cellNb.nodes.length; ++i)` and `for (let i = 0; i < cellNb.nodes.length; i++)` are the same thing, FWIW in old c++ compilers the ++i was was more efficient, so its just a habit.

(assuming that is what your picking up on)?

packages/markdown-it-plugins/src/util.ts
```diff
@@ -0,0 +1,78 @@
...
+            const id = `fence-${idx}-${i + 1}`;
+            const content = cellNb.nodes[i].value;
+            retVal.push({
+                ...fenceInfo,
+                id,
+                content,
+                innerHTML: `\
+        <span id="${id}" >
+            ${renderExecutedSrc(fenceInfo) ? content : ""}
+        </span>`
```
GordonSmith, line 49: I should remove the leading spaces here.

packages/markdown-it-plugins/tests/index.ts
GordonSmith: Yea I will remove comment code, it was actually copied from the spec file to here...

jeclrsg (approved):

PR #4253, opened 2024-10-22
feat: Additional VitePress Support

GordonSmith:
## Checklist:
  - [x] The commit is signed.


packages/markdown-it-plugins/src/loader.ts
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+}
+
+const regex = /^(.+\.(csv|tsv|txt|json|arrow|parquet))(?:\..*)?$/;
+interface FileInfo {
+    path: string;
+    data: any;
+}
+export async function findDataFiles(directory: string, recursive: boolean = false): Promise<FileInfo[]> {
```
jeclrsg, line 127: I didn't see anywhere else in this PR that was calling this function, but I couldn't remember what you'd said about the context from which this would be run. So I didn't know if you should be concerned with exploitative path traversal, lots of "../" and such? It is an exported function.

jeclrsg (commented):
@GordonSmith Looks good. I only had one question about that findDataFiles function in loader.ts. But it's probably not possible. Assuming it's a GH action's runner, somehow trying to get secrets, logs, etc?

packages/markdown-it-plugins/src/loader.ts
GordonSmith: It would typically get called in a `*.data.ts` file, which is what VitePress uses to bridge the gap between markdown and JS / Data at build time, in this file: `packages/markdown-it-plugins/docs/obt/summary.data.ts` there is a call to `DataFile.attach`, the findDataFiles would be used to find a bunch of potential candidates for the attach method.
Its not a security issue as its going to be called by the consumer of this package.

GordonSmith: FYI - the docs are here:  https://vitepress.dev/guide/data-loading

PR #4290, opened 2024-11-12
feat: Upgrade tree to v3

GordonSmith:
## Checklist:
  - [x] The commit is signed.


packages/tree/tests/tree.browser.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+                                break;
+                            case Treemap:
+                                render(new Treemap()
+                                    .columns(["label", "size"])
+                                    .data(flare.data)
+                                );
+                                break;
+
+                            default:
+                                it("Has render test", () => {
```
jeclrsg, line 440: I believe it's been the case in the last few PRs I reviewed also, but I only just thought about it now. Should these other cases not also have some kind of `it(...)` expectation in them as well?

jeclrsg (approved):
@GordonSmith just one question about the tests, approved though

packages/tree/tests/tree.browser.spec.ts
GordonSmith: The `it` call is buried inside the "render" function, it does a few checks but the main one is:
1. Set target to a placeholder div
2. Renders the widget
3. Sets the target to null
4. Sets the target to a new placeholder
5. Renders the widget

jeclrsg: I realized that when I was just reviewing another; chased what that `render(...)` import actually was

PR #4303, opened 2024-11-15
chore: Drop leaflet-shim as no longer needed

GordonSmith:
## Checklist:
  - [x] The commit is signed.


packages/map/src/leaflet/plugins/HeatLayer.ts
```diff
@@ -55,7 +53,7 @@ export class HeatLayer extends LeafletLayer2 {
     }
 
     redraw() {
-        if (this._heat && !this._frame && !this._map._animating) {
+        if (this._heat && !this._frame && !this._map["_animating"]) {
```
jeclrsg, line 56: Why all the changes from `.property` to `["property"]` in this?

jeclrsg (approved):

PR #4370, opened 2025-04-16
feat(comms): add Export function to WsDali service

jeclrsg:
## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/services/wsDali.ts
+++ b/packages/comms/src/services/wsDali.ts
@@ -4,5 +4,37 @@ export {
     WsDali
 };
 
+export interface ExportRequest {
+    Path: string;
+    Safe: boolean;
+}
+
+export interface ExportResponse {
+    filename: string;
+    content: Blob;
+}
+
 export class DaliService extends DaliServiceBase {
+
+    async Export(request: ExportRequest): Promise<ExportResponse> {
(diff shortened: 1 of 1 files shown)


packages/comms/src/services/wsDali.ts
```diff
@@ -4,5 +4,36 @@ export {
...
+        const blob = await response.blob();
+
+        const urlObj = window.URL.createObjectURL(blob);
+
+        const link = document.createElement("a");
+        link.href = urlObj;
+        link.download = filename ? filename.split("filename=")[1] : "WsDali_Export";
+        link.click();
+        link.remove();
+        window.URL.revokeObjectURL(urlObj);
```
GordonSmith, line 36: You can't assume that `document` and `window` will be available (NodeJS, ECL Extension etc.).  It should just return the blob.

Alternatively you could just (in ECL Watch):
```
        const link = document.createElement("a");
        link.href = `${exportUrl}?${reqParams.toString()}`;
        link.download = filename ? filename.split("filename=")[1] : "WsDali_Export";
        link.click();
        link.remove();
        window.URL.revokeObjectURL(urlObj);
  ```

GordonSmith (changes requested):

GordonSmith (approved):

PR #4399, opened 2025-06-23
fix: Sourcemaps out of sync

GordonSmith:
## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/dgrid/vite.config.ts
+++ b/packages/dgrid/vite.config.ts
@@ -1,36 +1,15 @@
 import { resolve } from "node:path";
-import { defineConfig } from "vite";
-import cssInjectedByJsPlugin from "vite-plugin-css-injected-by-js";
-import { hpccBundleNames } from "@hpcc-js/esbuild-plugins";
+import { createHpccViteConfig } from "@hpcc-js/esbuild-plugins";
 import pkg from "./package.json" with { type: "json" };
 
-const { alias, external, globals } = hpccBundleNames(pkg);
-
-export default defineConfig({
-    // Setting root to the `Visualizations/packages` folder to simplify debugging of dgrid-shim
-    root: resolve(__dirname, ".."),
-    build: {
-        outDir: resolve(__dirname, "dist"),
-        lib: {
-            entry: resolve(__dirname, "src/index.ts"),
-            name: pkg.name,
-            fileName: "index",
-        },
-        rollupOptions: {
-            external,
-            output: {
-                globals,
-            },
-        },
-        sourcemap: true
-    },
-    resolve: {
-        alias
-    },
-    esbuild: {
-        minifyIdentifiers: false
-    },
-    plugins: [
-        cssInjectedByJsPlugin()
-    ],
+export default createHpccViteConfig(pkg, {
+    configOverrides: {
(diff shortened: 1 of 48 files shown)


packages/observablehq-compiler/vite.config.ts
```diff
@@ -28,6 +28,8 @@ export default defineConfig({
         minifyIdentifiers: false
     },
     plugins: [
-        cssInjectedByJsPlugin()
+        cssInjectedByJsPlugin({
```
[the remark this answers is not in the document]
jeclrsg: A reasonable comment; I think you could take it even further and say (almost?) all of these config files are exactly the same, completely. So, maybe pull the whole thing out into something, maybe in that esbuild-plugins package? Not now, just saying for future reference.

packages/graph/vite.config.ts
```diff
@@ -1,33 +1,4 @@
...
-
-export default defineConfig({
-    build: {
-        lib: {
-            entry: resolve(__dirname, "src/index.ts"),
-            name: pkg.name,
-            fileName: "index",
-        },
-        rollupOptions: {
-            external: [...external, "react", "react-dom"],
```
jeclrsg, line 17: This was appending "react" and "react-dom" to the array "external"

packages/react/vite.config.ts
```diff
@@ -1,35 +1,4 @@
...
-    },
-    resolve: {
-        alias
-    },
-    esbuild: {
-        minifyIdentifiers: false
-    },
-    plugins: [
-        cssInjectedByJsPlugin(),
-        preact()
```
jeclrsg, line 33: This was including a plugin for preact

packages/dgrid/vite.config.ts
```diff
@@ -1,36 +1,4 @@
...
+import { createHpccViteConfig } from "@hpcc-js/esbuild-plugins";
 import pkg from "./package.json" with { type: "json" };
 
-const { alias, external, globals } = hpccBundleNames(pkg);
-
-export default defineConfig({
-    // Setting root to the `Visualizations/packages` folder to simplify debugging of dgrid-shim
-    root: resolve(__dirname, ".."),
-    build: {
-        outDir: resolve(__dirname, "dist"),
```
jeclrsg, line 13: Not sure that it matters, but I think this was the only package explicitly setting this outDir and the root a couple lines up

packages/dgrid2/vite.config.ts
```diff
@@ -1,37 +1,4 @@
...
-    },
-    resolve: {
-        alias
-    },
-    esbuild: {
-        minifyIdentifiers: false
-    },
-    plugins: [
-        cssInjectedByJsPlugin(),
-        react(),
```
jeclrsg, line 34: This one was including a plugin for react

packages/dgrid/vite.config.ts
GordonSmith: It does matter (as it needs to load the shim)

GordonSmith: (But only while in development environment)

jeclrsg (approved):

PR #4454, opened 2025-09-26
feat: bump wsdl versions

GordonSmith:
## Checklist:
  - [ ] The commit is signed.


packages/comms/src/services/wsdl/WsDali/v1.07/WsDali.ts
```diff
@@ -3,11 +3,12 @@ import { Service } from "../../../../espConnection.ts";
 
 export namespace WsDali {
 
+    export type base64Binary = string;
     export type unsignedInt = number;
 
     export interface AddRequest {
-        Path?: string;
-        Value?: string;
+        dstname?: string;
```
jeclrsg: This isn't right. And I think the change to the DeleteRequest props isn't either

packages/comms/src/services/wsdl/WsWorkunits/v2.02/WsWorkunits.ts
```diff
@@ -880,7 +881,7 @@ export namespace WsWorkunits {
         SelectColumnMode: LogSelectColumnMode;
         CustomColumns: CustomColumns;
         ComponentsFilter: ComponentsFilter;
-        Format: LogAccessLogFormat;
+        Format: string;
```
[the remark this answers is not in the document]
jeclrsg: This seems like it might not be right either, but it does reflect what the WSDL shows. In v2.01 this type was LogAccessLogFormat.

packages/comms/src/services/wsdl/ws_machine/v1.18/ws_machine.ts
```diff
@@ -0,0 +1,497 @@
+import { IConnection, IOptions } from "../../../../connection.ts";
+import { Service } from "../../../../espConnection.ts";
+
+export namespace WsMachine {
+
+    export type int = number;
+    export type long = number;
+    export type unsignedInt = number;
+
+    export enum ThresholdType {
```
jeclrsg, line 10: Were these two enums (ThresholdType and TpMachineType) added manually? Neither of them exist in the WSDL. 

I guess I've never noticed that a WSDL can refer to a type that isn't defined in it? ie, the ws_machine WSDL has several properties that have a `type="xsd:ThresholdType"` even though no such type is defined in it.

jeclrsg:
Also made me think I might start looking at using copilot to frame out some tests for any of these we don't already have

packages/comms/src/services/wsdl/ws_machine/v1.18/ws_machine.ts
GordonSmith: They were manually added

GordonSmith:
@jeclrsg I bumped the generation again so it includes the compression type (which is new from gavan).

PR #4465, opened 2025-10-16
fix(comms): adds back some interfaces to wsDFUXRef

jeclrsg:
there are apparently several interfaces that were previously defined manually in the wsDFUXRef service file, but are missing from the WSDL. this adds back any such interfaces & extends two functions where the WSDL defines the return type as a string instead of an object.

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/comms/src/services/wsDFUXRef.ts
+++ b/packages/comms/src/services/wsDFUXRef.ts
@@ -1,7 +1,121 @@
 import { DFUXRefServiceBase, type WsDFUXRef } from "./wsdl/WsDFUXRef/v1.04/WsDFUXRef.ts";
 
-export { type WsDFUXRef };
+export {
+    type WsDFUXRef
+};
+
+export namespace WsDFUXRefEx {
+
+    export interface Request {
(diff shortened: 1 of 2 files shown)


jeclrsg:
@GordonSmith I don't think there's been situation like this one, where so many types were missing from the WSDL. The only thing similar I saw was ws_machine, where it has a WsMachineEx namespace; but everything in there is extending existing interfaces. But do you think these should be wrapped up in a namespace also?

jeclrsg:
@GordonSmith regarding merging namespaces - with "isolatedModules" enabled (https://github.com/hpcc-systems/Visualization/blob/trunk/packages/comms/tsconfig.json#L14) merging across files evidently isn't possible. At one point copilot had recommended creating an "augmentation file (eg packages/comms/src/services/wsDFUXRef.augment.d.ts)" with something like:

```ts
declare module "./wsdl/WsDFUXRef/v1.04/WsDFUXRef.ts" {
  export namespace WsDFUXRef {
    //interfaces...
  }
}
```

I just opted for using a "WsDFUXRefEx" naming convention as had been done for ws_machine.

PR #4484, opened 2025-11-13
feat: update pipe.ts to use unlimited vardic types

GordonSmith:
## Checklist:
  - [x] The commit is signed.


packages/dataflow/src/utils/pipe.ts
```diff
@@ -1,74 +1,191 @@
...
+
+// =============================================================================
+// Type Resolution for Activity Chains (Without Source)
+// =============================================================================
+
+/**
+ * Recursively resolves the output type of a chain of activities starting from an iterable activity.
+ * Validates that each activity's input type matches the previous activity's output type.
+ */
+type ResolveIterableTail<CurrentOut, InitialIn, Tail extends readonly AnyActivity[]> =
```
jeclrsg, line 53: nitpick: these type defns are pretty hard to parse, but not sure what could improve them; maybe just indentation on the ternaries?

jeclrsg (approved):
@GordonSmith approve. just had a couple nitpicks, one about some type definitions, and then not specific to this per se, but just thinking it seems like it would be good for new tests to err on the side of having longer / more descriptive names

packages/dataflow/src/utils/pipe.ts
GordonSmith: Broke up the types into individual checks for readability (still a headache to read...)

PR #4489, opened 2025-11-20
fix: requestAmimatonFrame only needed when observed

GordonSmith:
Ensure only 1 animationFrame is active at a time.

## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/util/tests/observer.spec.ts
+++ b/packages/util/tests/observer.spec.ts
@@ -65,7 +65,101 @@ class ConflatingMessage extends Message {
     }
 }
 
-describe("observer2", function () {
+describe("dispatch", function () {
+    it("hasObserver - initially false", () => {
+        const ms = new Dispatch<SimpleMessage>();
+        expect(ms.hasObserver()).to.be.false;
+    });
+
+    it("hasObserver - true after attach", () => {
+        const ms = new Dispatch<SimpleMessage>();
+        const handle = ms.attach(() => { });
+        expect(ms.hasObserver()).to.be.true;
+        handle.release();
+    });
+
+    it("hasObserver - false after release", () => {
+        const ms = new Dispatch<SimpleMessage>();
+        const handle = ms.attach(() => { });
+        expect(ms.hasObserver()).to.be.true;
+        handle.release();
+        expect(ms.hasObserver()).to.be.false;
+    });
+
+    it("send - immediate synchronous dispatch", () => {
+        const ms = new Dispatch<SimpleMessage>();
+        let callCount = 0;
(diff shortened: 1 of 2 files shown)


packages/util/tests/observer.spec.ts
```diff
@@ -65,7 +65,113 @@ class ConflatingMessage extends Message {
...
+        });
+    });
+
+    it("RAF handle 0 is treated as scheduled", () => {
+        return new Promise<void>((done) => {
+            const ms = new Dispatch<SimpleMessage>();
+            let callCount = 0;
+            const handle = ms.attach((messages) => {
+                callCount++;
+                expect(messages.length).to.be.greaterThan(0);
```
jeclrsg: Could this not be `.to.equal(3)`, if the previous test could expect the message length to be 2?

packages/util/tests/observer.spec.ts
```diff
@@ -65,7 +65,113 @@ class ConflatingMessage extends Message {
...
+        return new Promise<void>((done) => {
+            const ms = new Dispatch<SimpleMessage>();
+            const handle = ms.attach(() => {
+                throw new Error("Should not be called");
+            });
+            ms.post(new SimpleMessage("test", "value"));
+            ms.flush();
+            handle.release();
+            // Wait for RAF to ensure no callback
+            setTimeout(() => done(), 50);
```
jeclrsg, line 111: Instead of implying success because of the lack of a thrown Error, couldn't you do something more like the others and have a count (initially 0), incremented in the handle callback, and then an expectation in the setTimeout that the count still be 0 because of the flush?

packages/util/tests/observer.spec.ts
```diff
@@ -65,7 +65,113 @@ class ConflatingMessage extends Message {
...
+        const ms = new Dispatch<SimpleMessage>();
+        // Should not throw
+        ms.send(new SimpleMessage("test", "value"));
+    });
+
+    it("post - no dispatch without observers", () => {
+        const ms = new Dispatch<SimpleMessage>();
+        // Should not throw
+        ms.post(new SimpleMessage("test", "value"));
+    });
```
jeclrsg: These two tests seem irrelevant? It doesn't look like there's anything in Dispatch that would have thrown an error, with or without the `if (!this.hasObservers()) return;`

packages/util/tests/observer.spec.ts
GordonSmith: This is checking that flush clears the message bugger prior to the message being dispatched.

PR #4573, opened 2026-04-08
chore(comms): bump the FileSpray service to v1.29

jeclrsg:
## Checklist:
  - [ ] The commit is signed.

### DIFF
--- a/packages/comms/src/services/fileSpray.ts
+++ b/packages/comms/src/services/fileSpray.ts
@@ -1,4 +1,4 @@
-import { FileSprayServiceBase, FileSpray } from "./wsdl/FileSpray/v1.27/FileSpray.ts";
+import { FileSprayServiceBase, FileSpray } from "./wsdl/FileSpray/v1.29/FileSpray.ts";
 
 export {
     FileSpray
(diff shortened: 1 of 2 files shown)


packages/comms/src/services/fileSpray.ts
```diff
@@ -20,31 +20,6 @@ export enum FileSprayStates {
...
-     * Savemap = "savemap",
-     * Addgroup = "addgroup",
-     * Server = "server",
-     */
-    Monitor = "monitor",
-    Copymerge = "copymerge",
-    Supercopy = "supercopy",
-    Publish = "publish",
-}
-
```
GordonSmith, line 47: I think we might want to keep these?

GordonSmith (changes requested):

packages/comms/src/services/fileSpray.ts
jeclrsg: They'll come out of the WSDL now, the `DFUCommand` on line 21 of the generated service. This one's kind of weird because it depends on the ESP changes in https://github.com/hpcc-systems/HPCC-Platform/pull/21133 but also is a dependency of the UI changes in that PR

PR #4575, opened 2026-04-09
chore: update vitest and related dependencies; refactor browser provi…

GordonSmith:
…der to use playwright

- Updated vitest and related packages to version 4.1.4 in package.json.
- Refactored browser provider in vite-utils.ts to use the new playwright integration.
- Changed test descriptions in react.browser.spec.ts from it() to describe() for better structure and readability.

## Checklist:
  - [x] The commit is signed.


packages/react/tests/react.browser.spec.ts
```diff
@@ -40,39 +40,39 @@ describe("@hpcc-js/react", () => {
     }
 
     describe("React Components - SVG Adapter", async () => {
-        it("edge.tsx", () => {
+        describe("edge.tsx", () => {
```
jeclrsg, line 43: changing these to "describe" doesn't seem right. should they be either "test" or "skip"?

jeclrsg (changes requested):
@GordonSmith seems like those tests won't run?

packages/react/tests/react.browser.spec.ts
GordonSmith: The newer vitest does not allow a "describe" inside an "it", the functions being called their own "describe" + "it" within them, so this is correct,

GordonSmith:
npm ci failing...

packages/react/tests/react.browser.spec.ts
jeclrsg: Ah, ok. It wasn't obvious that all those `renderMedium` functions contained `it(...)`

PR #4579, opened 2026-04-13
Marshaller port 3

GordonSmith:
## Checklist:
  - [x] The commit is signed.


GordonSmith:
@jeclrsg - this is in two commits:
1. The orig 2.x.x sources
2. The changes needed to port to 3.x.x

Its only the second commit that needs review

packages/marshaller/src/publish.ts
jeclrsg, line 1: should this file be empty?

packages/phosphor/src/SplitPanel.ts
```diff
@@ -66,5 +66,21 @@ export class SplitPanel extends HTMLWidget {
...
+                const widget = (msg as Msg.WAActivateRequest).wa.widget;
+                if (this._prevActive !== widget) {
+                    this._prevActive = widget;
+                    this.childActivation(widget);
+                }
+                break;
+        }
+    }
+
+    childActivation(w: Widget) {
```
jeclrsg, line 84: no function body?

package-lock.json
jeclrsg, line 1: should there be a change to package.json as well?

packages/marshaller/src/__package__.ts
jeclrsg, line 1: why is this one different from all of the other v3 `__package__.ts` files?

packages/phosphor/src/SplitPanel.ts
GordonSmith: Its the default event handler (so will be either overridden or hooked with `on(`childActivation`, ...)`

packages/marshaller/src/publish.ts
GordonSmith: Should not be included!

package-lock.json
GordonSmith: I don't think so as the marshaller gets pulled in via the workspaces glob, but the lock gets updated with its deps...

PR #4601, opened 2026-05-20
feat(dgrid): add ColumnSetTable and integrate with ColumnSetGrid

GordonSmith:
## Checklist:
  - [x] The commit is signed.

### DIFF
--- a/packages/dgrid/src/ColumnSetTable.ts
+++ b/packages/dgrid/src/ColumnSetTable.ts
@@ -1,0 +1,97 @@
+import { ColumnSetGrid } from "./dgrid-shim.ts";
+import { type ColumnType } from "./RowFormatter.ts";
+import { Table } from "./Table.ts";
+
+export class ColumnSetTable extends Table {
+
+    constructor() {
+        super();
+    }
+
+    protected _buildColumnSets(): ColumnType[][][] {
+        const sets = this.columnSets();
+        const colNames = this.columns();
+        const used = new Set<string>();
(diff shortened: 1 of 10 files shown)


packages/dgrid/src/ColumnSetTable.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    protected _createGrid(opts: any, node: HTMLElement): any {
+        return new ColumnSetGrid({ ...opts }, node);
+    }
+
+    protected _applyColumnsToGrid() {
+        const sets = this._buildColumnSets();
+        this._dgrid.set("columnSets", sets);
+        for (let i = 0; i < sets.length - 1; i++) {
+            const setWidth = sets[i][0].reduce((total: number, col: ColumnType) => total + (col.width ?? 0), 0);
```
jeclrsg, line 73: Thinking it would be good to more prominently call out this > 0 width requirement somewhere in here too. Either a comment, or even something like a `console.warn` if someone tried to use it and none of the columns were visible?

jeclrsg (changes requested):
@GordonSmith just one comment about how obvious it is that the columns require a user-defined width

packages/dgrid/src/ColumnSetTable.ts
GordonSmith: This is the internal dgrid col width and not the user supplied / calculated width.  IOW its adding up the sets columns widths and then setting the column set width to be that total.

PR #4632, opened 2026-08-13
feat(comms): Add opt-in arraybuffer response type to fetchDetails

sodon-lnr:
Added the `options` argument to `fetchDetails()` signatures and the response type of `arraybuffer` to allow callers of fetchDetails/fetchDetailsRaw to handle their own parsing. 

## Checklist:
  - [ ] The commit is signed.

## Testing:

- Unit tests failure rate was the same before and after the changes. I believe the failures were due to not having an esp running locally
- Built the package locally with `npm pack` and used it in the following snippet to test against an environment and ensure outputs are correct and types match:

```typescript
export async function commsBufferTest(serverOpts: IOptions) {
  const wu = Workunit.attach(serverOpts, 'W20260226-200207');

  // Existing parsed behavior
  const scopes = await wu.fetchDetails();
  console.log('Parsed scopes:', scopes.length);

  // New raw behavior
  const raw = await wu.fetchDetails(
    { ScopeFilter: { Scopes: ['workunit'] } },
    { bypassResponseParsing: true },
  );
  console.log('Raw bytes:', raw.byteLength);

  const text = new TextDecoder().decode(new Uint8Array(raw));
  const payload = JSON.parse(text);
  console.log('Top level keys:', Object.keys(payload));
}
```

### DIFF
--- a/packages/comms/src/connection.ts
+++ b/packages/comms/src/connection.ts
@@ -163,7 +163,7 @@ function doFetch(opts: IOptions, action: string, requestInit: RequestInit, heade
         }
     }
 
-    function handleResponse(response: Response): Promise<any> {
+    function handleResponse(response: Response): Promise<any | ArrayBuffer | string> {
         if (response.ok) {
             switch (responseType) {
                 case "json":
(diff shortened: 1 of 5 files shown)


GordonSmith:
@sodon-lnr - I:
1. Fixed the tests on main (and added a specific WUDetails test)
2. Rebased your PR onto this new main branch
3. Tweaked your PR to be a typing descriminator rather than an additional method (as a seperate commit on your PR)
4. Optimised the response fixup to avoid doubling the memory usage/copy.

If your happy with this I will merge and release.

GordonSmith (approved):

sodon-lnr:
@GordonSmith I tested with the new commit locally just to be sure. Everything looks good to me. Thanks for your help!

